PR #6, opened 2025-10-28
Add `identifierutil` package

snelusha:
This pull request introduces a new package, `identifierutil`, that provides comprehensive utilities for encoding and decoding identifiers, with robust handling for special characters, Unicode codepoints, and JVM-reserved characters. It also adds an extensive suite of unit tests to ensure correctness and reliability of these utilities.

### Identifier encoding/decoding utilities

* Added `identifierutil/util.go` with functions for encoding and decoding identifiers, including handling of special characters, JVM-reserved characters, Unicode escapes, and generated method names. This includes functions like `EncodeFunctionIdentifier`, `EncodeNonFunctionIdentifier`, `DecodeIdentifier`, `UnescapeJava`, and related helpers.

### Test coverage

* Added `identifierutil/util_test.go` with thorough unit tests for all major utility functions, covering edge cases and various scenarios for escaping, unescaping, encoding, and decoding identifiers.

### DIFF
--- a/identifierutil/util.go
+++ b/identifierutil/util.go
@@ -1,3 +1,21 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (http://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


identifierutil/util.go
```diff
@@ -0,0 +1,241 @@
+package identifierutil
```
warunalakshitha, line 19: Should we include licence header?

PR #13, opened 2025-11-03
Add constants and log utils

### LINKED ISSUE #11: Migrate `ballerina-runtime`

snelusha:
This pull request introduces new error handling and logging utilities, along with a set of constants for symbol flags. The main changes include adding stack trace support to errors, structured logging for critical failures, and a new constants package for commonly used values. Comprehensive unit tests have also been added to verify the new error handling features.

**Error handling and logging enhancements:**

* Added `ErrorWithStackTrace` type and the `DecorateWithStackTrace` function in `errors.go` to support attaching stack traces to errors when the `BAL_BACKTRACE` environment variable is enabled.
* Implemented `LogBadSad` and `PrintCrashLog` functions in `errors.go` for structured logging of critical, unexpected errors, including timestamp and severity level.

**Constants and symbol flags:**

* Introduced a new `constants.go` file with commonly used string constants and a set of symbol flag bitmasks, plus a utility function `IsFlagOn` to check flag values.

**Testing improvements:**

* Added `errors_test.go` with unit tests for stack trace decoration, error formatting, and edge cases for the new error handling utilities.

**Refactoring and imports:**

* Updated imports in `errors.go` to support new error handling and logging features.

Closes #11

### DIFF
--- a/common/constants/constants.go
+++ b/common/constants/constants.go
@@ -21,28 +21,28 @@ const (
 	UserHome   = "user.home"
 )
 
-// SymbolFlags contains flag constants for symbols.
+type SymbolFlag int64
+
 const (
-	PUBLIC        int64 = 1
-	NATIVE        int64 = 2
-	FINAL         int64 = 4
-	ATTACHED      int64 = 8
-	READONLY      int64 = 32
-	REQUIRED      int64 = 256
-	PRIVATE       int64 = 1024
-	OPTIONAL      int64 = 4096
-	REMOTE        int64 = 32768
-	CLIENT        int64 = 65536
-	RESOURCE      int64 = 131072
-	SERVICE       int64 = 262144
-	TRANSACTIONAL int64 = 33554432
-	CLASS         int64 = 268435456
-	ISOLATED      int64 = 536870912
-	ENUM          int64 = 8589934592
-	ANY_FUNCTION  int64 = 549755813888
+	PUBLIC SymbolFlag = 1 << iota
+	NATIVE
+	FINAL
+	ATTACHED
+	READONLY
+	REQUIRED
+	PRIVATE
+	OPTIONAL
+	REMOTE
+	CLIENT
+	RESOURCE
+	SERVICE
+	TRANSACTIONAL
+	CLASS
+	ISOLATED
+	ENUM
+	ANY_FUNCTION
 )
 
-// IsFlagOn checks if a specific flag is set in the bitmask.
-func IsFlagOn(bitmask, flag int64) bool {
-	return (bitmask & flag) == flag
+func (sf SymbolFlag) IsOn(flag SymbolFlag) bool {
+	return (sf & flag) == flag
 }


common/constants/constants.go
```diff
@@ -0,0 +1,48 @@
...
+	OPTIONAL      int64 = 4096
+	REMOTE        int64 = 32768
+	CLIENT        int64 = 65536
+	RESOURCE      int64 = 131072
+	SERVICE       int64 = 262144
+	TRANSACTIONAL int64 = 33554432
+	CLASS         int64 = 268435456
+	ISOLATED      int64 = 536870912
+	ENUM          int64 = 8589934592
+	ANY_FUNCTION  int64 = 549755813888
```
heshanpadmasiri: Usually bit mask flags are defined like this https://pkg.go.dev/go/build#ImportMode. I think you still need uint64, but might be worth it to make a type alias (you can attach methods to alias directly)

common/errors/errors.go
```diff
@@ -59,3 +66,96 @@ func NewIllegalArgumentError(argument any) *IllegalArgumentError {
...
+}
+
+func DecorateWithStackTrace(err error) error {
+	if os.Getenv("BAL_BACKTRACE") != "true" {
+		return err
+	}
+
+	if err == nil {
+		return nil
+	}
```
heshanpadmasiri:
```suggestion
	if err == nill || os.Getenv("BAL_BACKTRACE") != "true" {
		return err
	}
```

common/errors/errors.go
```diff
@@ -59,3 +66,96 @@ func NewIllegalArgumentError(argument any) *IllegalArgumentError {
...
+
+	const (
+		maxDepth = 32
+		skip     = 2 // skip DecorateWithStackTrace and its caller
+	)
```
heshanpadmasiri: These are single use right? I think it is easier to just pass the value at the call site

common/errors/errors_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Error message should contain original error
+	if !strings.Contains(decorated.Error(), "test error with stack") {
+		t.Errorf("Error message should contain original error text")
+	}
+
+	// Error message should contain stack trace
+	errorMsg := decorated.Error()
+	if !strings.Contains(errorMsg, "\n\tat ") {
```
heshanpadmasiri, line 64: I think it is cleaner for you to check for the actual error message instead of matching part by part. It clear state what you are expecting

common/errors/errors.go
```diff
@@ -59,3 +66,96 @@ func NewIllegalArgumentError(argument any) *IllegalArgumentError {
...
+
+	stack := make([]uintptr, maxDepth)
+	length := runtime.Callers(skip, stack[:])
+	return &errorWithStackTrace{
+		err:   err,
+		stack: stack[:length],
+	}
+}
+
+func (e *errorWithStackTrace) StackTrace() []byte {
```
heshanpadmasiri: Do we need this to be public?

common/errors/errors_test.go
snelusha: Exact match would be flaky since the stack trace contains absolute paths and architecture

PR #43, opened 2026-01-26
Support Ballerina Runtime Interpreter for Language Subset 1

### LINKED ISSUE #38: Support Ballerina Runtime Interpreter for Language Subset 1
**Summary**

Implement runtime interpreter support in the Ballerina interpreter for the Subset 1 language programs defined in the corpus/bal/subset1 directory of the ballerina-lang-go repository (GitHub). This issue tracks building the interpreter so it can execute the programs in that subset directory successfully.

The “Subset 1 corpus” contains representative .bal programs that exercise the basic language features expected in the first subset of Ballerina semantics.

**Goals**

- The interpreter should be able to:
- Load a Ballerina program from the subset1 corpus
- Parse and execute its compiled BIR or the original .bal semantics (depending on your architecture)
- Produce correct program outputs for all test programs in that directory

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/38


corpus/integration_test.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		subsetNum := formatSubsetNumber(subset)
+		fmt.Printf("Subset %s\n", subsetNum)
+		fmt.Println("==========================")
+
+		balFiles := findBalFiles(corpusBalDir)
+
+		for _, balFile := range balFiles {
+			if isFileSkipped(balFile) {
```
heshanpadmasiri: Shall we turn these in to subtests that run in parallel? That's what existing corpus tests do already https://github.com/ballerina-platform/ballerina-lang-go/blob/b2fcfdd5c8c2186c2f59fcbfde22ef1f7aa7c276/ast/corpus_ast_test.go#L83

runtime/internal/exec/binary_instructions.go
```diff
@@ -0,0 +1,486 @@
...
+import (
+	"ballerina-lang-go/bir"
+	"fmt"
+	"math"
+)
+
+func execBinaryOpAdd(binaryOp *bir.BinaryOp, frame *Frame) {
+	rhsOp1, rhsOp2, lhsOp := extractBinaryOpIndices(binaryOp)
+	op1 := frame.locals[rhsOp1]
+	op2 := frame.locals[rhsOp2]
```
heshanpadmasiri: I wonder if it would be worth it at this point to abstract away the logic for getting and setting operands. Technically subset 1 already has globals (constants) currently I have hack to make them behave as local constants after constant propagation, since we haven't decided on how to represent and initialize globals. I am thinking of using negative indices for them so package can have it's own array for globals, which we initilize with the _init_ function. Either way when we make that decision then it would be an easy fix

runtime/internal/exec/executor.go
```diff
@@ -0,0 +1,286 @@
...
+		bb = execTerminator(term, bb, frame)
+	}
+	// Return the value of the return variable
+	// Return variable is always at index 0
+	return frame.locals[0]
+}
+
+func execInstruction(inst bir.BIRNonTerminator, frame *Frame) {
+	switch inst.GetKind() {
+	case bir.INSTRUCTION_KIND_CONST_LOAD:
```
heshanpadmasiri: Would it be better to do a outer switch on type and then if needed (I assume this is needed for some like binary and unary) check the kind. Cast anyway has do the type check, so if that is enough I think it is better to avoid a virtual function call. (Also after the type is narrowed in the switch it a static dispatch)

runtime/internal/exec/non_terminator_instructions.go
```diff
@@ -0,0 +1,57 @@
...
+	// For now, this is a placeholder - the actual TypeCast type doesn't exist yet
+	// When implemented, it should extract RhsOp and LhsOp and perform type conversion
+	panic("TypeCast instruction not yet implemented")
+}
+
+func execNewArray(newArray *bir.NewArray, frame *Frame) {
+	// Get result variable and array size from frame
+	if newArray.SizeOp == nil {
+		// Create empty array if no size specified
+		frame.locals[newArray.LhsOp.Index] = make([]any, 0)
```
heshanpadmasiri: I believe the convention is to treat nil the same as an empty array (which would be the 0 value in this case; I assume). At least all the standard library functions like size, and append treat it that way

runtime/internal/exec/terminators_instructions.go
```diff
@@ -0,0 +1,77 @@
...
+
+package exec
+
+import (
+	"ballerina-lang-go/bir"
+	"ballerina-lang-go/runtime/internal/modules"
+	"fmt"
+)
+
+func execBranch(branchTerm *bir.Branch, frame *Frame) bir.BIRBasicBlock {
```
heshanpadmasiri: I wonder is it better to return the BB by pointer instead because we can have BBs with lot's of instructions which would be expensive to create copies of?

main.go
```diff
@@ -24,8 +24,11 @@ import (
 	debugcommon "ballerina-lang-go/common"
 	"ballerina-lang-go/context"
 	"ballerina-lang-go/parser"
+	"ballerina-lang-go/runtime"
+	_ "ballerina-lang-go/stdlibs/io" // Initialize io module with extern functions
```
heshanpadmasiri: Wouldn't it be better to have this in runtime.go file? That would ensure initialization has happened before running Interpret is called irrespective of how it is called

corpus/integration_test.go
```diff
@@ -0,0 +1,269 @@
...
+		if err == nil && !info.IsDir() && filepath.Ext(path) == ".bal" {
+			files = append(files, path)
+		}
+		return nil
+	})
+	return files
+}
+
+func isFileSkipped(filePath string) bool {
+	content, err := os.ReadFile(filePath)
```
heshanpadmasiri: Wouldn't it be better to skip if the filePath ends with -e (this is what we already do https://github.com/ballerina-platform/ballerina-lang-go/blob/b2fcfdd5c8c2186c2f59fcbfde22ef1f7aa7c276/ast/corpus_ast_test.go#L46) and perhaps have an hardcoded list if we need skip something else temporarily. By having annotations I the file it self I fear that we might accidentally leave a test as skipped because we forgot to update the comment.

heshanpadmasiri (commented):
Generally looks okay to me except few minor nits

corpus/integration_test.go
warunalakshitha: Fixed. Test are running parallel now.

runtime/internal/exec/executor.go
warunalakshitha: Fixed.

runtime/internal/exec/non_terminator_instructions.go
warunalakshitha: Fixed

corpus/integration_test.go
warunalakshitha: removed @disabled and skip -e files

main.go
warunalakshitha: Moved to libs.go

runtime/internal/exec/binary_instructions.go
warunalakshitha: Moved to setting and getting operands to frame so we can handle global operands future.

corpus/integration_test.go
```diff
@@ -0,0 +1,453 @@
...
+	panicRegex       = regexp.MustCompile(`//\s*@panic\s+(.+)`)
+	supportedSubsets = []string{"subset1"}
+
+	// skipTestsMap contains file paths (relative to corpus/bal/subset1) that should be skipped
+	skipTestsMap = makeSkipTestsMap([]string{
+		"01-function/assign8-v.bal",
+		"01-function/assign9-v.bal",
+		"01-function/call08-v.bal",
+		"01-nil/rel-v.bal",
+	})
```
warunalakshitha: Following tests are disabled. Need to investigate more on abnormal errors when running those

warunalakshitha: Fixed

heshanpadmasiri (approved):
LGTM

PR #56, opened 2026-01-29
Port Semantic Analysis

### LINKED ISSUE #45: Port Semantic Analysis
Fallow on from #41. After type resolution all the AST nodes that declare a Type should have resolved (we may still need to propagate type to other nodes).  Therefore we should be able to simplify some of the checks using semtypes.

heshanpadmasiri:
## Purpose
$subject.
Resolves #45

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/test_util/test_util.go
+++ b/test_util/test_util.go
@@ -69,25 +69,16 @@ func GetTests(t *testing.T, kind TestKind, filterFunc func(string) bool) []TestC
 		outputBaseDir = "ast"
 		outputExt = ".txt"
 		subsets = []string{"subset1"}
-
 	case Parser:
 		outputBaseDir = "parser"
 		outputExt = ".json"
-		// Parser processes entire corpus tree, no subset restriction
-
 	case BIR:
 		outputBaseDir = "bir"
 		outputExt = ".txt"
 		subsets = []string{"subset1"}
 	}
-
-	// Resolve input and output directories
 	resolvedInputDir, resolvedOutputDir := resolveDir(t, inputBaseDirAlt, outputBaseDir)
-
-	// Discover files
 	files := discoverFiles(t, resolvedInputDir, subsets, filterFunc)
-
-	// Create test pairs with expected paths
 	testPairs := make([]TestCase, 0, len(files))
 	for _, inputPath := range files {
 		expectedPath := computeExpectedPath(inputPath, resolvedInputDir, resolvedOutputDir, outputExt)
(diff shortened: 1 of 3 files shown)


test_util/update.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// ReadExpectedFile reads the expected output file and returns its contents
+func ReadExpectedFile(t *testing.T, expectedPath string) string {
+	t.Helper()
+
```
warunalakshitha: Do we need the extra new lines in the code?

test_util/test_util.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, inputPath := range files {
+		expectedPath := computeExpectedPath(inputPath, resolvedInputDir, resolvedOutputDir, outputExt)
+		relPath, _ := filepath.Rel(resolvedInputDir, inputPath)
+		testPairs = append(testPairs, TestCase{
+			InputPath:    inputPath,
+			ExpectedPath: expectedPath,
+			Name:         relPath,
+		})
+	}
+
```
warunalakshitha, line 92: I think we can remove these extra new lines in some places

semtypes/function_ops.go
```diff
@@ -129,3 +129,74 @@ func (this *FunctionOps) functionTheta(cx Context, t0 SemType, t1 SemType, pos *
 		return ((IsSubtype(cx, t0, s0) || this.functionTheta(cx, Diff(s0, t0), s1, pos.Next)) && (IsSubtype(cx, t1, Complement(s1)) || this.functionTheta(cx, s0, Intersect(s1, t1), pos.Next)))
 	}
 }
+
+// Corresponds to dom^? in AMK tutorial.
+func FunctionParamListType(cx Context, fnTy SemType) SemType {
+	if !IsSubtypeSimple(fnTy, FUNCTION) {
+		return nil
+	}
+	switch ty := fnTy.(type) {
```
warunalakshitha, line 138: If we use numeric typetag will that increase perfromance?

heshanpadmasiri: type tag part of the type descriptor not the semtype. With #76 caller can determine if they want to optimize away this check based on information in symbol for example when you know function type is atomic. Though I am not entirely sure if that would be worth it given it is executed only in semantic analysis which anyway runs concurrently for all packages once.

semantics/type_resolver.go
```diff
@@ -0,0 +1,342 @@
...
+	"ballerina-lang-go/tools/diagnostics"
+	"fmt"
+	"math/big"
+	"strconv"
+)
+
+// UniformRef is a reference to a symbol that is same whether we are referring to it from different package or same
+// so we can efficiently implement lookup logic.
+// TODO: string here is just a placeholder
+type UniformRef string
```
heshanpadmasiri, line 33: #76 will properly implement type lookup of symbols

PR #76, opened 2026-02-02
Add symbol resolutions

### LINKED ISSUE #67: Add symbol resolutions
I think before #60 we need to properly handle symbol resolutions. Currently we have some hacks to make it work in semenatic analysis and type resolution. Instead of adding on to that we should handle this properly. We can deal with imports later. My current thoughts on how to implement this as fallows

- We need to do symbol resolution before type resolution
  - Type can refer to other type. We need a way to get the AST node for that type and then "resolve" that type when resolving this given type
    - Recursion requires a Type definition. So we need a resolve type definition function. Type definition node will have the semtype attached to it and we can use that get the semtype if already defined

    ```ballerina
    // We start with depth 0
    function resolveTypeDefn(ModuleSymbols mod, s:TypeDefn defn, int depth) returns t:SemType|ResolveTypeError {
        t:SemType? t = defn.semType;
        if t == () {
            if depth == defn.cycleDepth {
                return err:semantic(`invalid cycle detected for ${defn.name}`, s:defnLocation(defn));
            }
            defn.cycleDepth = depth;
            t:SemType s = check resolveTypeDesc(mod, defn, depth, defn.td);
            t = defn.semType;
            if t == () {
                defn.semType = s;
                defn.cycleDepth = -1;
                return s;
            }
            else {
                // This can happen with recursion
                // We use the first definition we produced
                // and throw away the others
                return t;
            }
        }
        else {
            return t;
        }
    }
    ```
- We need a way to look up symbols (and then import exports) from anywhere (assuming you have some sort of a symbol table)
- **Remember for values you can declare them withing scopes like functions**
- We can treat scopes as linked list and inside each scope we have a map of bindings
  - We need to keep track of what sort of bindings we are dealing with for handling things like type narrowing
  - When we enter some sort of statement that can introduce a new scope (if block, function defn, etc) we'll add a new link to the scope chain
    - From visitors perspective this means we need a new visitor.
      - Otherwise we don't know when we have visited all the child nodes of that given node
    - **Not necessarily require to be a new kind of visitor just a new instance**
- Root scope will at the end be converted to the `PackageSymbolTable` at the end
- All ast nodes that represent a reference will have a **reference** to a symbol
  - Variable reference expression
    - Need to think how to handle function references. May be for the time being holding a `Symbol` is good enough.
  - Variable declaration statements
    - Given symbol (accessed via the variable reference) we should be able to find the definition for LSP
  - Function invocation
  - Function definition
- At the definition we'll initialize the symbol and add them to the current scope. Then at reference we'll try to look it up in the scope and get the symbol reference.
  - This means we need to initialize all the top level symbols before visiting since they can refer each other.
    - These would be just name -> zero object actual `resolution` should still happen in the visitor
      - This way we end up in any "deadlocks"
  - All other symbols needs to be defined before use

### Symbol table

- For each package we need to track all the symbols it exports (see [#Import / exports])
- This symbol table will be a part of the compiler context
  - Store this behind a read write lock
- At the beginning we should initialize this with implicit imports
  - For time being add a hack to see if `ballerina/io` is imported is so add it's symbols to the table
- For efficient lookup we'll have different kinds of symbols
  - Type
  - Function
  - Variable

  ```go
  type Symbol interface {
    Type() semtype
  }

  type symbolBase struct {
    ty semtype
  }

  type TypeSymbol struct {
    symbolBase
  }

  // TODO: think about how to support closures correctly
  // -- We may need to track captured symbols as well
  // -- But closures don't capture symbols from other functions?
  type FunctionSymbol struct {
    symbolBase
    Signature FunctionSignature
  }

  type FunctionSignature struct {
    ParamTypes []semtype
    ReturnType semtype
    // RestParamType is nil if there is no rest param
    RestParamType semtype
  }

  type ValueSymbol interface {
    Symbol
  }

  type ConstantSymbol struct {
    symbolBase
    // TODO: We can set the value during symbol resolution
    // -- This should allow us to do constant propagation concurrently
    Value any
  }

  // TODO: add other symbols for module level declarations that can be referred

  ```

- **This is the same as import / export**
  - We can model this as package symbol table where each package has it's own symbol table.
    - **Each package's symbol resolution return the symbol table for that package**

    ```go
    type PackageSymbolTable struct {
      // These could be either constants or variables
      // IMPORTANT: these needs to be a single unique value within the whole system so pointers instead of values
      Values map[string]*ValueSymbol
      Functions map[string]*FunctionSymbol
      Types map[string]*TypeSymbol
    }

    // TODO: will need to refactor this given it is already defined
    type SymbolTable struct {
      // here package ID is the org, package name and version triple
      PackageSymbols map[PackageID] PackageSymbolTable
    }
    ```

  - Compiler context stores symbol table for each package.

### Import / exports

- Each package exports a set of definitions
  - Function signature
  - Type descriptors (I think we need more that just the semtype here since we need to know how to create the value)
  - Constants
    - For each constant we need to know the type
- We should be able to get all of these before semantic analysis
  - We'll do these at type resolution
  - After this point all the code generation steps can happen concurrently for each package

heshanpadmasiri:
## Purpose
$subject
Resolves #67
Depends on #56

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/semantics/corpus_symbol_resolver_test.go
+++ b/semantics/corpus_symbol_resolver_test.go
@@ -30,7 +30,6 @@ import (
 
 func TestSymbolResolver(t *testing.T) {
 	flag.Parse()
-
 	testPairs := test_util.GetValidTests(t, test_util.AST)
 
 	for _, testPair := range testPairs {
@@ -66,10 +65,8 @@ func testSymbolResolution(t *testing.T, testCase test_util.TestCase) {
 	env := semtypes.GetIsolatedTypeEnv()
 	importedSymbols := ResolveImports(cx, env, pkg)
 	ResolveSymbols(cx, pkg, importedSymbols)
-
 	validator := &symbolResolutionValidator{t: t, testPath: testCase.InputPath}
 	ast.Walk(validator, pkg)
-
 	// If we reach here, symbol resolution completed without panicking
 	t.Logf("Symbol resolution completed successfully for %s", testCase.InputPath)
 }
@@ -83,7 +80,6 @@ func (v *symbolResolutionValidator) Visit(node ast.BLangNode) ast.Visitor {
 	if node == nil {
 		return nil
 	}
-
 	// Check if this node should have a symbol resolved
 	if nodeWithSymbol, ok := node.(ast.BNodeWithSymbol); ok {
 		if nodeWithSymbol.Symbol() == nil {
@@ -91,7 +87,6 @@ func (v *symbolResolutionValidator) Visit(node ast.BLangNode) ast.Visitor {
 				node, node.GetPosition(), v.testPath)
 		}
 	}
-
 	return v
 }
 
@@ -99,7 +94,6 @@ func (v *symbolResolutionValidator) VisitTypeData(typeData *model.TypeData) ast.
 	if typeData.TypeDescriptor == nil {
 		return nil
 	}
-
 	// Check if this type descriptor should have a symbol resolved
 	if typeWithSymbol, ok := typeData.TypeDescriptor.(ast.BNodeWithSymbol); ok {
 		if typeWithSymbol.Symbol() == nil {
@@ -107,6 +101,5 @@ func (v *symbolResolutionValidator) VisitTypeData(typeData *model.TypeData) ast.
 				typeData.TypeDescriptor, typeData.TypeDescriptor.GetPosition(), v.testPath)
 		}
 	}
-
 	return v
 }
--- a/semantics/corpus_type_resolver_test.go
+++ b/semantics/corpus_type_resolver_test.go
@@ -108,10 +108,8 @@ func (v *typeResolutionValidator) Visit(node ast.BLangNode) ast.Visitor {
 }
 
 func isExpr(node ast.BLangNode) bool {
-	if _, ok := node.(model.ExpressionNode); ok {
-		return true
-	}
-	return false
+	_, ok := node.(model.ExpressionNode)
+	return ok
 }
 
 func (v *typeResolutionValidator) VisitTypeData(typeData *model.TypeData) ast.Visitor {
--- a/semantics/type_resolver.go
+++ b/semantics/type_resolver.go
@@ -288,10 +288,8 @@ func (t *TypeResolver) resolveHexFloatingPointLiteral(n *ast.BLangNumericLiteral
 func updateSymbolType(ctx *context.CompilerContext, node ast.BLangNode, ty semtypes.SemType) {
 	if nodeWithSymbol, ok := node.(ast.BNodeWithSymbol); ok {
 		symbol := nodeWithSymbol.Symbol()
-		if symbol != nil {
-			// symbol resolver should initialize the symbol
-			ctx.SetSymbolType(symbol, ty)
-		}
+		// symbol resolver should initialize the symbol
+		ctx.SetSymbolType(symbol, ty)
 	}
 }
 


lib/io.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
```
warunalakshitha, line 1: Can we move the runtime code here to. basically,

lib/io/compiler
lib/io/runtime

I hope compiler code will be removed once project API imports supported.

heshanpadmasiri: Yes.

semantics/type_resolver.go
```diff
@@ -274,6 +283,18 @@ func (t *TypeResolver) resolveHexFloatingPointLiteral(n *ast.BLangNumericLiteral
 	return nil
 }
 
+// updateSymbolType updates the symbol's type if the node has an associated symbol.
+// This synchronizes the symbol's type with the node's resolved type.
+func updateSymbolType(ctx *context.CompilerContext, node ast.BLangNode, ty semtypes.SemType) {
+	if nodeWithSymbol, ok := node.(ast.BNodeWithSymbol); ok {
+		symbol := nodeWithSymbol.Symbol()
+		if symbol != nil {
```
warunalakshitha: Can symbol be nil? Do we need this check?

semantics/corpus_type_resolver_test.go
```diff
@@ -62,26 +63,57 @@ func testTypeResolution(t *testing.T, testCase test_util.TestCase) {
...
 	return v
 }
 
+func isExpr(node ast.BLangNode) bool {
+	if _, ok := node.(model.ExpressionNode); ok {
+		return true
+	}
+	return false
+}
+
```
warunalakshitha, line 114:
```suggestion
func isExpr(node ast.BLangNode) bool {
	_, ok := node.(model.ExpressionNode)
	return ok
}

```

semantics/corpus_symbol_resolver_test.go
```diff
@@ -0,0 +1,112 @@
...
+			v.t.Errorf("Symbol not resolved for %T at %s in %s",
+				node, node.GetPosition(), v.testPath)
+		}
+	}
+
+	return v
+}
+
+func (v *symbolResolutionValidator) VisitTypeData(typeData *model.TypeData) ast.Visitor {
+	if typeData.TypeDescriptor == nil {
```
warunalakshitha, line 94: can this be nil?

semantics/corpus_symbol_resolver_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"ballerina-lang-go/parser"
+	"ballerina-lang-go/semtypes"
+	"ballerina-lang-go/test_util"
+	"flag"
+	"testing"
+)
+
+func TestSymbolResolver(t *testing.T) {
+	flag.Parse()
+
```
warunalakshitha: no need new lines here

semantics/corpus_semantic_analysis_test.go
```diff
@@ -65,24 +65,30 @@ func testSemanticAnalysis(t *testing.T, testCase test_util.TestCase) {
 	}
 	pkg := ast.ToPackage(compilationUnit)
 
-	// Step 1: Type Resolution
+	// Step 1: Symbol Resolution
+	env := semtypes.GetIsolatedTypeEnv()
```
warunalakshitha, line 69: do we need seperate semantic analyser test. should that covered with integration tests

lib/io.go
warunalakshitha: I'll move with subset 2 support PR

semantics/corpus_semantic_analysis_test.go
heshanpadmasiri: My view is each compiler stage should validate its output/invariants so that when something changes we know from which stage things started to break. Given we run each corpus test in parallel I don't think it will have much effect on performance.

semantics/corpus_symbol_resolver_test.go
heshanpadmasiri: It is a hold overfrom jBallerina where each node had a BType. We can gurantee that each node in the AST has a type (functions, classes etc are NEVER) but not a type descriptor (you can't create values from most AST nodes). Ideally we should refactor the ast such that we explicitly add TypeDescriptor nodes, but for timebeing I thought this is good enough

PR #109, opened 2026-02-10
Add proper support for list mutation

### LINKED ISSUE #95: Add proper support for list mutation
I think we already have frontend support for this. May need to properly validate in semantic analysis if we properly catch mutations to readonly lists at compile time. In runtime we should be able to use the same logic as the compile time to validate the element type (this may be less efficient but should be correct; I think it is worth it to revisit this when this becomes a perf issue)

heshanpadmasiri:
## Purpose
$subject
Resolves #95

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Documentation**
  * Added production/debug build instructions and detailed profiling guide with examples.

* **New Features**
  * Debug-mode profiling CLI flags and start/stop hooks (server address configurable).
  * Improved list construction and runtime list behavior visible to programs.
  * Cycle-safe, improved printing/formatting for complex or nested values.

* **Chores**
  * Expanded example/test corpus demonstrating list behaviors and edge cases.

### DIFF
--- a/bir/bir_gen.go
+++ b/bir/bir_gen.go
@@ -554,10 +554,6 @@ func listConstructorExpression(ctx *stmtContext, bb *BIRBasicBlock, expr *ast.BL
 	for i := len(expr.Exprs); i < lat.Members.FixedLength; i++ {
 		ty := lat.MemberAt(i)
 		fillerVal := values.DefaultValueForType(ty)
-		if fillerVal == values.NeverValue {
-			// this should never happen; semantic analyzer should have cought these
-			panic("never value in filler position")
-		}
 		fillerOperand := ctx.addTempVar(ty)
 		fillerLoad := &ConstantLoad{}
 		fillerLoad.Value = fillerVal
--- a/cli/cmd/prof_debug.go
+++ b/cli/cmd/prof_debug.go
@@ -1,5 +1,3 @@
-//go:build debug
-
 // Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
 //
 // WSO2 LLC. licenses this file to you under the Apache License,
@@ -15,6 +13,8 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
+
+//go:build debug
 
 package main
 
--- a/cli/cmd/prof_release.go
+++ b/cli/cmd/prof_release.go
@@ -1,5 +1,3 @@
-//go:build !debug
-
 // Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
 //
 // WSO2 LLC. licenses this file to you under the Apache License,
@@ -15,6 +13,8 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
+
+//go:build !debug
 
 package main
 


bir/bir_gen.go
```diff
@@ -541,18 +542,43 @@ func typeConversionExpression(ctx *stmtContext, curBB *BIRBasicBlock, expr *ast.
...
-	constantLoad.Value = int64(-1)
+	constantLoad.Value = int64(len(expr.Exprs))
 	constantLoad.LhsOp = sizeOperand
 	bb.Instructions = append(bb.Instructions, constantLoad)
 
+	lat := expr.AtomicType
+	for i := len(expr.Exprs); i < lat.Members.FixedLength; i++ {
+		ty := lat.MemberAt(i)
+		fillerVal := values.DefaultValueForType(ty)
+		if fillerVal == values.NeverValue {
```
warunalakshitha: IMO we do not need to check this. It will break at runtime if it happens (should never)

cli/cmd/prof_debug.go
```diff
@@ -0,0 +1,82 @@
+//go:build debug
```
warunalakshitha: Should we add this before header or after header?

runtime/internal/exec/non_terminators.go
```diff
@@ -101,15 +100,18 @@ func castValue(value any, targetType semtypes.SemType) any {
...
-func resizeArrayIfNeeded(arrPtr *[]any, arr []any, idx int) []any {
-	if idx >= len(arr) {
-		newArr := make([]any, idx+1)
-		copy(newArr, arr)
-		*arrPtr = newArr
-		return newArr
-	}
-	return arr
-}
+// func valueType(value any) semtypes.SemType {
```
warunalakshitha: We can remove this?

values/list.go
```diff
@@ -0,0 +1,63 @@
...
+// under the License.
+
+package values
+
+import (
+	"ballerina-lang-go/semtypes"
+	"math"
+)
+
+type List struct {
```
warunalakshitha, line 24: Should we use Array or List as struct name?

bir/bir_gen.go
heshanpadmasiri: Check removed

cli/cmd/prof_debug.go
heshanpadmasiri: I don't have a strong opinion about either approach I think the restriction is there can't be code before it. Moved to after license in both files

values/list.go
heshanpadmasiri: I prefer List because that matches the terminology of the spec and internally we use terms Array and Tuple in some places for subsets of List

corpus/integration_test.go
```diff
@@ -56,7 +58,9 @@ var (
 	errorRegex  = regexp.MustCompile(`//\s*@error`)
 
 	// Skip tests that cause unrecoverable Go runtime errors
-	skipTestsMap = makeSkipTestsMap([]string{})
+	skipTestsMap = makeSkipTestsMap([]string{
+		"subset3/03-list/24-v.bal",
```
heshanpadmasiri, line 62: @warunalakshitha this is the test case where I mentioned we have an issue with how we resize the underlying array (copying instead of appending)

warunalakshitha: Will check on this.

warunalakshitha (approved):
LGTM

Issue #85, opened 2026-02-04
Add conditional type narrowing


heshanpadmasiri:
Needs #84 . We need to do this before semantic analysis after we have created the CFG

heshanpadmasiri:
# Conditional type narrowing

- [Relevant section of the spec](https://ballerina.io/spec/lang/master/#conditional_variable_type_narrowing)
- At expression level we need to know the type of the expression, thus needs to be done after type resolution
  - But this needs to be done before semantic analysis since it do the type checking of the code.
  - We will also need to mutate symbols in the AST while doing this so semantic analysis will work as is
- Having control flow information with us will be useful so probably we should do this after CFG
  - Otherwise this will need to "create" a CFG of it's own

## Design

- I think the way to do this is with a binding chain

  ```go
  type binding struct {
    // this is the reference to the underlying symbol
    ref SymbolRef
    bindingSymbol *model.Symbol
    prev *binding
  }
  ```

- Analyze expression will take a binding chain and return a expression effect

  ```go
  type expressionEffect struct {
    ifTrue *binding
    // For expression that are not conditional both will be same
    ifFalse *binding
  }
  ```

  - When we have a var ref we'll go over the binding chain until we find the nearest binding to the symbol

- Same is true for statements. This needs to handle cases like return, it will bind the type to never

  ```go
  type statementEffect struct {
    // this is nil for statements like return/ panic which will not effect the bindings?
    // -- I think the CFG graph already handle them properly but just in case
    binding *binding
  }
  ```

- We can implement type merging on top of CFG, we analyze all the parent (only those with forward edges), take their statement effects and merge them
  - First we need to group all the binding symbols by the underlying symbol (`ref` in binding). Then we take the type of each binding symbols type and intersect it to form the type of the new symbol.
    - One optimization would be to keep track if in any branch the bindingSymbol is a narrowed symbol. If that is not the case we can simply bind it back to original symbol without the need for narrowing.
    - For each binding chain we need to take the nearest binding

- We'll introduce a new Symbol kind for narrowed symbols
  - **Do we need this to be public in model? I don't think reset of the code needs to know what kind of symbol this is**
  - All operations except type will be deligated to the underlying symbol

    ```go
    type NarrowedSymbol struct {
      ty Semtype

      // underlying is the reference to the "original" un-narrowed symbol. Repeated narrowings should point to the same
      // underlying symbol and this would allow us to add un-narrowing symbol easily
      underlying SymbolRef
[...]

PR #141, opened 2026-02-16
Refactor AST representation of symbol

heshanpadmasiri:
## Purpose
Make it explicit in AST that it only hold SymbolRef 
Cleanup some compiler diagnostics
Related #85

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Refactor**
  * Core symbol representation converted to use symbol references throughout the compiler; public APIs updated accordingly—integrations must adapt to the new symbol-ref-based interfaces.
  * Replaced several conditional chains with switch statements and added explicit constant type annotations for clearer, safer code.

* **New Features**
  * Added a helper to check whether a symbol is set on AST nodes.
  * Introduced interfaces to improve scope/symbol-space handling.

* **Bug Fixes**
  * Removed a runtime validation that could cause unexpected panics when assigning else-statements.


semantics/type_resolver.go
```diff
@@ -738,16 +733,20 @@ func (t *TypeResolver) resolveIndexBasedAccess(expr *ast.BLangIndexBasedAccess)
...
 		t.ctx.InternalError("invocation has no symbol", expr.GetPosition())
 		return nil
 	}
 	if deferredMethodSymbol, ok := symbol.(*deferredMethodSymbol); ok {
 		return t.resolveMethodCall(expr, deferredMethodSymbol)
-	} else {
-		return t.resolveFunctionCall(expr, symbol)
 	}
+	symbolRef, ok := symbol.(*model.SymbolRef)
+	if !ok {
```
warunalakshitha, line 745: Can this be happen?

heshanpadmasiri: That shouldn't happen but type can't gurantee that it is not the case so we can accidently set it as an symbol (unlike other nodes where we can enforce this with types we have to allow deferred symbol until type resolution is complete)

PR #190, opened 2026-02-23
Add support for hex literals

### LINKED ISSUE #97: Add support for hex int literals
I think this is just a matter of type resolution support
https://github.com/ballerina-platform/ballerina-lang-go/blob/560685bcc74899820928d75bde2c12b807b16655/semantics/type_resolver.go#L282

heshanpadmasiri:
## Purpose
Resolves #97 

## Summary by CodeRabbit

* **New Features**
  * Added toHexString() for integer → hexadecimal string conversion.
  * Improved literal handling: better distinction and automatic widening between ints and floats.
  * Added multiple example programs demonstrating integer/float/hex behaviors and conversions.

* **Bug Fixes**
  * Numeric parsing now falls back to floating parsing for values that don't parse as integers, reducing parsing errors.

### DIFF
--- a/doc/subset4.md
+++ b/doc/subset4.md
@@ -30,7 +30,7 @@
 ## Expressions
 
 - [Literal](https://ballerina.io/spec/lang/master/#literal)
-  - Currently support `nil-literal`, `boolean-literal`, `numeric-literal` (see [restrictions](#numeric-literal)), and `string-literal` only
+  - Currently support `nil-literal`, `boolean-literal`, `numeric-literal` and `string-literal` only
 - [lvexpr](https://ballerina.io/spec/lang/master/#section_7.14.1)
   - Currently only support [variable-reference-lvexpr](https://ballerina.io/spec/lang/master/#variable-reference-lvexpr)
 - [`Call`](https://ballerina.io/spec/lang/master/#call-expr)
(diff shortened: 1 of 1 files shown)


lib/int/runtime/int.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func initIntModule(rt *runtime.Runtime) {
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "toHexString", func(args []values.BalValue) (values.BalValue, error) {
+		n, ok := args[0].(int64)
+		if !ok {
+			return nil, fmt.Errorf("first argument must be an int")
+		}
+		var s string
+		if n == 0 {
```
warunalakshitha: Shall we use switch case which is more clear.

warunalakshitha:
```go
 switch {
    case n == 0:
        return "0"
    case n == math.MinInt64:
        return "-8000000000000000"
    case n < 0:
        return "-" + strconv.FormatInt(-n, 16)
    default:
        return strconv.FormatInt(n, 16)
    }```

File: semantics/semantic_analyzer.go [hunk 2]
```diff
@@ -513,7 +513,7 @@ func analyzeExpression[A analyzer](a A, expr ast.BLangExpression, expectedType s
 	// Literals - just validate
 	case *ast.BLangLiteral:
 		widenNumericLiteral(a, expr, expectedType)
-		validateResolvedType(a, expr, expectedType)
+		analyzeLiteral(a, expr, expectedType)
```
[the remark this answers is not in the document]
heshanpadmasiri: widenNumericLiteral only widen if needed

lib/int/runtime/int.go
heshanpadmasiri: Fixed

PR #194, opened 2026-02-24
Add backend support for type test expressions

### LINKED ISSUE #146: Add backend support for type test expressions
Will need #132 in backend and #145

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/146

## Summary by CodeRabbit

* **New Features**
  * Improved runtime type-awareness: package type context is propagated into execution and a value→type mapping is exposed for richer type checks.
* **Tests**
  * Enabled additional integration tests and added sample cases exercising expanded type-check behavior.
* **Refactor**
  * Streamlined execution and call-dispatch paths for more consistent instruction and type-test handling.

### DIFF
--- a/bir/bir_gen.go
+++ b/bir/bir_gen.go
@@ -128,6 +128,7 @@ func GenBir(ctx *context.CompilerContext, ast *ast.BLangPackage) *BIRPackage {
 			birPkg.MainFunction = birFunc
 		}
 	}
+	birPkg.TypeCtx = semtypes.ContextFrom(ctx.GetTypeEnv())
 	return birPkg
 }
 
--- a/bir/bir_gen.go
+++ b/bir/bir_gen.go
@@ -128,7 +128,7 @@ func GenBir(ctx *context.CompilerContext, ast *ast.BLangPackage) *BIRPackage {
 			birPkg.MainFunction = birFunc
 		}
 	}
-	birPkg.TypeCtx = semtypes.ContextFrom(ctx.GetTypeEnv())
+	birPkg.TypeEnv = ctx.GetTypeEnv()
 	return birPkg
 }
 
--- a/bir/model.go
+++ b/bir/model.go
@@ -66,7 +66,7 @@ type (
 		Functions     []BIRFunction
 		Constants     []BIRConstant
 		MainFunction  *BIRFunction
-		TypeCtx       semtypes.Context
+		TypeEnv       semtypes.Env
 	}
 
 	BIRImportModule struct {
--- a/bir/model.go
+++ b/bir/model.go
@@ -66,6 +66,7 @@ type (
 		Functions     []BIRFunction
 		Constants     []BIRConstant
 		MainFunction  *BIRFunction
+		TypeCtx       semtypes.Context
 	}
 
 	BIRImportModule struct {
--- a/runtime/internal/exec/non_terminators.go
+++ b/runtime/internal/exec/non_terminators.go
@@ -78,7 +78,8 @@ func execTypeCast(typeCast *bir.TypeCast, frame *Frame) {
 func execTypeTest(typeTest *bir.TypeTest, frame *Frame, reg *modules.Registry) {
 	sourceValue := frame.GetOperand(typeTest.RhsOp.Index)
 	valueType := values.SemTypeForValue(sourceValue)
-	typeCtx := reg.GetTypeCtx()
+	typeEnv := reg.GetTypeEnv()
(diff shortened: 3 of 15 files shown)


runtime/internal/exec/non_terminators.go
```diff
@@ -70,11 +71,18 @@ func execArrayLoad(access *bir.FieldAccess, frame *Frame) {
...
 	frame.SetOperand(typeCast.LhsOp.Index, result)
 }
 
+func execTypeTest(typeTest *bir.TypeTest, frame *Frame, reg *modules.Registry) {
+	sourceValue := frame.GetOperand(typeTest.RhsOp.Index)
+	valueType := values.SemTypeForValue(sourceValue)
+	typeCtx := reg.GetTypeCtx()
+	matches := semtypes.IsSubtype(typeCtx, valueType, typeTest.Type) != typeTest.IsNegation
+	frame.SetOperand(typeTest.LhsOp.Index, matches)
+}
```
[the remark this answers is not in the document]
warunalakshitha: It will never be nil

values/values.go
```diff
@@ -50,3 +50,24 @@ func DefaultValueForType(t semtypes.SemType) BalValue {
...
+	case int64:
+		return &semtypes.INT
+	case float64:
+		return &semtypes.FLOAT
+	case string:
+		return &semtypes.STRING
+	case *big.Rat:
+		return &semtypes.DECIMAL
+	case *List:
+		return v.Type
```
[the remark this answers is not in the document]
warunalakshitha: That cannot be nil

bir/model.go
```diff
@@ -66,6 +66,7 @@ type (
 		Functions     []BIRFunction
 		Constants     []BIRConstant
 		MainFunction  *BIRFunction
+		TypeCtx       semtypes.Context
```
snelusha: Should we use `*semtypes.Context` instead?

runtime/internal/exec/terminators.go
```diff
@@ -20,56 +20,51 @@ import (
...
 )
 
 func execBranch(branchTerm *bir.Branch, frame *Frame) *bir.BIRBasicBlock {
 	opIndex := branchTerm.Op.Index
-	value := frame.GetOperand(opIndex)
-	cond, ok := value.(bool)
-	if !ok {
-		panic(fmt.Sprintf("invalid branch condition type at index %d: %T (expected bool)", opIndex, value))
-	}
+	v := frame.GetOperand(opIndex)
```
snelusha: Can we inline this?

```
if frame.GetOperand(opIndex).(bool) {
    return branchTerm.TrueBB
}
return branchTerm.FalseBB
```

snelusha (approved):
Just a few minor nits, otherwise LGTM!

bir/model.go
warunalakshitha: Cannot use semtypes.ContextFrom(ctx.GetTypeEnv()) (value of interface type semtypes.Context) as *semtypes.Context value in assignment: semtypes.Context does not implement *semtypes.Context (type *semtypes.Context is pointer to interface, not interface)

bir/bir_gen.go
```diff
@@ -128,6 +128,7 @@ func GenBir(ctx *context.CompilerContext, ast *ast.BLangPackage) *BIRPackage {
 			birPkg.MainFunction = birFunc
 		}
 	}
+	birPkg.TypeCtx = semtypes.ContextFrom(ctx.GetTypeEnv())
```
heshanpadmasiri: I am not sure this is a good idea. Type context is suppose to be for each thread not package. Each thread needs to create that threads type context using the type environment (which should be shared by all in the interpreter)

values/values.go
```diff
@@ -50,3 +50,24 @@ func DefaultValueForType(t semtypes.SemType) BalValue {
 		return NeverValue
 	}
 }
+
+func SemTypeForValue(v BalValue) semtypes.SemType {
```
heshanpadmasiri, line 54: I don't think this is going to work for simple basic types you need the singleton basic type otherwise something like this wont work
```ballerina
int a = 5;
io:println(a is byte)
```

bir/bir_gen.go
warunalakshitha: I have changed the logic pass the type env to the runtime

values/values.go
warunalakshitha: using singleton basic type.

heshanpadmasiri (approved):
LGTM

heshanpadmasiri (approved):

PR #212, opened 2026-02-28
Report syntax errors

### LINKED ISSUE #211: Report diagnostics on syntax errors

snelusha:
## Purpose
Resolves #211 

## Summary by CodeRabbit

* **Bug Fixes**
  * More consistent, user-friendly syntax error messages and improved diagnostic highlights.
  * Replaced crashy behavior with reported diagnostics so syntax issues are surfaced without panics.
  * Minor syntax corrections (missing semicolons) to reduce spurious parse errors.

* **Chores**
  * Compilation now aborts further resolution when diagnostics are present to avoid cascading errors.

### DIFF
--- a/ast/node_builder.go
+++ b/ast/node_builder.go
@@ -597,19 +597,11 @@ func innermostDiagnosticNodes(node tree.Node) []tree.Node {
 }
 
 func diagnosticMessage(node tree.Node) string {
-	if deep := tree.FindDeepestDiagnosticSTNode(node.InternalNode()); deep != nil && len(deep.Diagnostics()) > 0 {
-		key := deep.Diagnostics()[0].DiagnosticCode().MessageKey()
-		if key != "" {
-			return strings.ReplaceAll(strings.TrimPrefix(key, "error."), ".", " ")
-		}
+	deep := tree.FindDeepestDiagnosticSTNode(node.InternalNode())
+	if deep == nil || len(deep.Diagnostics()) == 0 {
+		return "syntax error"
 	}
-	if diags := node.InternalNode().Diagnostics(); len(diags) > 0 {
-		key := diags[0].DiagnosticCode().MessageKey()
-		if key != "" {
-			return strings.ReplaceAll(strings.TrimPrefix(key, "error."), ".", " ")
-		}
-	}
-	return "syntax error"
+	return strings.ReplaceAll(strings.TrimPrefix(deep.Diagnostics()[0].DiagnosticCode().MessageKey(), "error."), ".", " ")
 }
 
 func getPosition(node tree.Node) Location {
--- a/ast/node_builder.go
+++ b/ast/node_builder.go
@@ -577,18 +577,23 @@ func getFileName(node tree.Node) string {
 	return st.FilePath()
 }
 
-func innermostDiagnosticNode(node tree.Node) tree.Node {
+func innermostDiagnosticNodes(node tree.Node) []tree.Node {
 	if !node.HasDiagnostics() {
 		return nil
 	}
+
+	var nodes []tree.Node
 	if nt, ok := node.(tree.NonTerminalNode); ok {
 		for child := range nt.ChildNodes() {
 			if child != nil && child.HasDiagnostics() {
-				return innermostDiagnosticNode(child)
+				nodes = append(nodes, innermostDiagnosticNodes(child)...)
 			}
 		}
 	}
-	return node
+	if len(nodes) > 0 {
+		return nodes
+	}
+	return []tree.Node{node}
 }
 
 func diagnosticMessage(node tree.Node) string {
@@ -1769,7 +1774,7 @@ func (n *NodeBuilder) generateAndAddBLangStatements(statementNodes tree.NodeList
 			continue
 		}
 		if currentStatement.HasDiagnostics() {
-			n.reportSyntaxDiagnostic(currentStatement)
+			n.reportSyntaxDiagnostics(currentStatement)
 			continue
 		}
 		if currentStatement.Kind() == common.FORK_STATEMENT {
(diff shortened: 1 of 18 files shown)


ast/node_builder.go
```diff
@@ -590,6 +590,33 @@ func getFileName(node tree.Node) string {
 	return st.FilePath()
 }
 
+func innermostDiagnosticNode(node tree.Node) tree.Node {
+	if node == nil || !node.HasDiagnostics() {
```
warunalakshitha: Seems node never nil. so lets remove this

ast/node_builder.go
```diff
@@ -590,6 +590,33 @@ func getFileName(node tree.Node) string {
...
+				return innermostDiagnosticNode(child)
+			}
+		}
+	}
+	return node
+}
+
+func diagnosticMessage(node tree.Node) string {
+	key := ""
+	if node != nil {
```
warunalakshitha: can node == nil

ast/node_builder.go
```diff
@@ -3214,3 +3246,14 @@ func (n *NodeBuilder) getBLangVariableNode(bindingPattern tree.BindingPatternNod
 
 	return createSimpleVariableNodeWithLocationTokenLocation(varPos, varName, getPosition(varName))
 }
+
+func (n *NodeBuilder) reportSyntaxDiagnostic(node tree.Node) {
+	if node == nil {
```
warunalakshitha: same as above comment

ast/node_builder.go
```diff
@@ -757,7 +782,8 @@ func (n *NodeBuilder) createTypeNode(typeNode tree.Node) model.TypeDescriptor {
 		return &bLUserDefinedType
 	case common.SIMPLE_NAME_REFERENCE:
 		if typeNode.HasDiagnostics() {
-			panic("unimplemented")
+			n.reportSyntaxDiagnostic(typeNode)
+			return n.types.nilType
```
warunalakshitha: this should be covered by tests

warunalakshitha:
There are some new code lines not covered by tests. Shall we check why

ast/node_builder.go
```diff
@@ -590,6 +590,31 @@ func getFileName(node tree.Node) string {
 	return st.FilePath()
 }
 
+func innermostDiagnosticNode(node tree.Node) tree.Node {
```
heshanpadmasiri: I think this is wrong, a given parent node can have multiple nodes with syntax errors. (you should return an array)
(may be try a function call with mutiple invalid error constructors (I recall error constructors getting diagnostics))

ast/node_builder.go
```diff
@@ -757,7 +782,8 @@ func (n *NodeBuilder) createTypeNode(typeNode tree.Node) model.TypeDescriptor {
 		return &bLUserDefinedType
 	case common.SIMPLE_NAME_REFERENCE:
 		if typeNode.HasDiagnostics() {
```
heshanpadmasiri: I don't think can happen anymore? At statement level if there is a diagnostic we abort so we shouldn't reach here (I so it must be fixed. After this I expect transform* functions to have the invariant paremeters don't have errors). Fix other places as well.

corpus/bal/subset5/05-error/constructor-1-e.bal
```diff
@@ -22,5 +22,5 @@ type Detail record {|
 type ErrorWithDetail error<Detail>;
 
 function bar() returns int|error {
-    return error ErrorWithDetail(message = "message", id = 1) // @error
+    return error ErrorWithDetail(message = "message", id = 1); // @error
```
warunalakshitha, line 25: Why this change?

heshanpadmasiri:
> > > There are some new code lines not covered by tests. Shall we check why
> > 
> > 
> > We currently don’t have any test that produce syntax errors.
> 
> Shall we add a test for syntax errors as well. @heshanpadmasiri can you recommend good test and which subset should be added

1. Add statements directly in the module level 
2. Invalid `,,` cases in things like list constructors and function calls
3. `return` instead of `returns` in function signatures

corpus/cli/output/help/help.txtar
```diff
@@ -5,6 +5,7 @@ Usage:
   bal [command]
 
 Available Commands:
+  completion  Generate the autocompletion script for the specified shell
```
warunalakshitha, line 8: Do we need this?

ast/node_builder.go
```diff
@@ -577,6 +577,41 @@ func getFileName(node tree.Node) string {
...
+			return strings.ReplaceAll(strings.TrimPrefix(key, "error."), ".", " ")
+		}
+	}
+	if diags := node.InternalNode().Diagnostics(); len(diags) > 0 {
+		key := diags[0].DiagnosticCode().MessageKey()
+		if key != "" {
+			return strings.ReplaceAll(strings.TrimPrefix(key, "error."), ".", " ")
+		}
+	}
+	return "syntax error"
```
warunalakshitha: Do we need this. seems whole if condition  not covered

parser/tree/node.go
```diff
@@ -384,28 +384,11 @@ func (n *NonTerminalNodeBase) ChildNodes() iter.Seq[Node] {
...
-				return n.ChildInBucket(listChildIndex)
-			}
-			index += child.BucketCount()
-		} else {
-			if childIndex == index {
-				return n.ChildInBucket(i)
-			}
-			index++
-		}
+	if childIndex < 0 || childIndex >= n.internalNode.BucketCount() {
```
warunalakshitha: Seems this is never will true

ast/node_builder.go
```diff
@@ -577,6 +577,41 @@ func getFileName(node tree.Node) string {
...
+	if len(nodes) > 0 {
+		return nodes
+	}
+	return []tree.Node{node}
+}
+
+func diagnosticMessage(node tree.Node) string {
+	if deep := tree.FindDeepestDiagnosticSTNode(node.InternalNode()); deep != nil && len(deep.Diagnostics()) > 0 {
+		key := deep.Diagnostics()[0].DiagnosticCode().MessageKey()
+		if key != "" {
```
azinneera: Can the key be empty?

ast/node_builder.go
azinneera: Shall we define an enum for the error types?

snelusha: https://github.com/ballerina-platform/ballerina-lang-go/blob/main/parser/common/diagnostic_error_code.go
We do have right?

corpus/cli/output/help/help.txtar
snelusha: We have this in the main branch also.
https://github.com/ballerina-platform/ballerina-lang-go/blob/main/corpus/cli/output/help/help.txtar

warunalakshitha (approved):
LGTM

PR #214, opened 2026-02-28
Migrate corpus integration tests to standard go test

### LINKED ISSUE #213: Migrate corpus integration tests to standard go test

snelusha:
## Purpose
Resolves #213 

## Summary by CodeRabbit

* **Tests**
  * Refined integration test harness for parallelized, per-case execution and more reliable failure isolation
  * Improved stdout/stderr comparison with concise side-by-side diffs for clearer error reporting
  * Added support to regenerate expected outputs to simplify test maintenance
  * Added discovery for a new "integration" test kind and skip logic for selectively skipping tests


corpus/integration_test.go
```diff
@@ -72,122 +60,83 @@ type testResult struct {
...
 
-func TestIntegrationSuite(t *testing.T) {
+func TestIntegration(t *testing.T) {
 	flag.Parse()
 
-	corpusBalDir := corpusBalBaseDir
-	if _, err := os.Stat(corpusBalDir); os.IsNotExist(err) {
-		return
-	}
+	testPairs := test_util.GetValidTests(t, test_util.Integration)
```
heshanpadmasiri: I think you need to validate -p tests are paniking as expected as well

corpus/integration_test.go
```diff
@@ -236,20 +184,17 @@ func runIntegrationCase(balFile string) (stdout, stderr string) {
...
 
 func evaluateTestResult(expectedStdout, expectedStderr, actualStdout, actualStderr string) testResult {
-	expectedStdoutNorm := normalizeOutput(expectedStdout)
-	expectedStderrNorm := normalizeOutput(expectedStderr)
-	actualStdoutNorm := normalizeOutput(actualStdout)
-	actualStderrNorm := normalizeOutput(actualStderr)
-
-	stdoutMatchesExpected := actualStdoutNorm == expectedStdoutNorm
-	stderrMatchesExpected := actualStderrNorm == expectedStderrNorm
+	expectedStdoutNorm := normalizeNewlines(expectedStdout)
```
heshanpadmasiri: Do we need to normalize expected? I would assume --update will write to files in the normalized form

corpus/integration_test.go
```diff
@@ -313,7 +257,7 @@ func capturePrintlnOutput(stdoutBuf *bytes.Buffer) func(args []values.BalValue)
 }
 
 func updateIntegrationTestCase(txtarPath, stdout, stderr string) error {
```
heshanpadmasiri: I think this should be done conditionally only if the file content is different from existing (Otherwise we are touching the file even with same content). Also it should fallow the same convention as existing corpus tests where if it had to update the file it's a test failure

warunalakshitha:
Do we need annotations in bal files now? . @error @output etc. @heshanpadmasiri @snelusha

heshanpadmasiri:
> Do we need annotations in bal files now? . @error @output etc. @heshanpadmasiri @snelusha

I think it is good to have them still at least for errors since at a glance we can see if the error message is where it should be not some where else. I am fine with removing output (not sure we validate this anyway)

PR #220, opened 2026-03-04
Add runtime support for error constructor

### LINKED ISSUE #219: Support error constructor at runtime
Evaluate Ballerina error constructor expressions (error [TypeRef] ( message [, cause] [, name = value, ...] )) at runtime so they produce a proper error value that can be returned, pattern-matched, and printed. 

The front end and BIR already handle error constructors; the interpreter must execute the NewError instruction and build a values.
Error (message, optional cause, optional detail map, optional type name) so these expressions work correctly when the program runs.

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/219

## Summary by CodeRabbit

* **New Features**
  * Added support for typed errors with a TypeName included in error values.
  * Error values now include message, optional cause, and structured detail and have a readable string representation.

* **Tests**
  * Added integration fixtures demonstrating various error outputs.
  * Previously skipped error tests are now enabled to run for fuller coverage.


values/error.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"ballerina-lang-go/semtypes"
+	"strconv"
+	"strings"
+)
+
+// Error represents a Ballerina error value at runtime.
+type Error struct {
+	Message  string
```
heshanpadmasiri, line 28: Wouldn't this needs to hold the type as well (similar to list and map), you can do type checks/casts for subtypes of error (will also need to update SemTypeForValue) as well

heshanpadmasiri (approved):
Generally LGTM except for type information

heshanpadmasiri (dismissed):
Generally LGTM except for type support

values/error.go
warunalakshitha: Fixed

heshanpadmasiri (approved):
LGTM

PR #221, opened 2026-03-04
Fix holding pointers to BDDs in lookup tables

### LINKED ISSUE #217: Fix holding pointers to BDDs in lookup tables
Currently lookup tables such as cx.memotable holds BDD directly which is a pointer to the actual BDD. This mean lookup expects equal BDDs to be canonicalized (which we don't). We need to either not hold pointers or canonicalize BDDs so that table lookup will work as expected.

heshanpadmasiri:
## Purpose
resolves #217 
depends on #182


warunalakshitha (approved):
LGTM

PR #223, opened 2026-03-05
Make type resolution of local nodes concurrent

heshanpadmasiri:
## Purpose
$subject
Depends on #221

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrency safety in symbol handling and diagnostic recording to prevent race conditions.

* **Performance**
  * Parallelized per-function type resolution for faster overall compilation.

* **Refactor**
  * Consolidated import collection to avoid duplicates and simplified type-resolution flow for clearer, more reliable behavior.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -20,10 +20,12 @@ import (
 	"ballerina-lang-go/model"
 	"ballerina-lang-go/semtypes"
 	"ballerina-lang-go/tools/diagnostics"
+	"sync"
 )
 
 type CompilerContext struct {
 	env         *CompilerEnvironment
+	mu          sync.Mutex
 	diagnostics []diagnostics.Diagnostic
 }
 
@@ -107,7 +109,9 @@ func (this *CompilerContext) SemanticError(message string, pos diagnostics.Locat
 
 func (this *CompilerContext) addDiagnostic(code string, severity diagnostics.DiagnosticSeverity, message string, pos diagnostics.Location) {
 	diagnostic := diagnostics.CreateDiagnostic(diagnostics.NewDiagnosticInfo(&code, message, severity), pos)
+	this.mu.Lock()
 	this.diagnostics = append(this.diagnostics, diagnostic)
+	this.mu.Unlock()
 }
 
 func (this *CompilerContext) HasDiagnostics() bool {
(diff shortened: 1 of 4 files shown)


semantics/type_resolver.go
```diff
@@ -64,8 +63,38 @@ func ResolveTopLevelNodes(ctx *context.CompilerContext, pkg *ast.BLangPackage, i
...
+		fn := &pkg.Functions[i]
+		wg.Add(1)
+		go func(idx int) {
+			defer wg.Done()
+			resolvers[idx] = resolveFunctionBody(ctx, pkg, fn, importedSymbols)
+		}(i)
+	}
+	wg.Wait()
+
+	seen := make(map[string]bool)
```
warunalakshitha: seen := make(map[string]bool, len(resolvers))

model/symbol.go
```diff
@@ -176,12 +176,16 @@ func (space *SymbolSpace) AddSymbol(name string, symbol Symbol) {
...
 	}
+	space.mu.Lock()
 	space.lookupTable[name] = len(space.Symbols)
 	space.Symbols = append(space.Symbols, symbol)
+	space.mu.Unlock()
 }
 
 func (space *SymbolSpace) GetSymbol(name string) (SymbolRef, bool) {
+	space.mu.RLock()
 	index, ok := space.lookupTable[name]
```
warunalakshitha, line 187: can lookuptable change? do we need lock since only getting symbol

context/context.go
```diff
@@ -20,12 +20,13 @@ import (
 	"ballerina-lang-go/model"
 	"ballerina-lang-go/semtypes"
 	"ballerina-lang-go/tools/diagnostics"
+	"sync"
 )
 
 type CompilerContext struct {
 	env         *CompilerEnvironment
+	mu          sync.Mutex
 	diagnostics []diagnostics.Diagnostic
```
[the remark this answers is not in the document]
heshanpadmasiri: We don't care about the ordering of diagnostics only that it has all the diagnostics

semantics/type_resolver.go
heshanpadmasiri: Fixed

model/symbol.go
heshanpadmasiri: Currently yes given we add the narrowed symbol to the space where the symbol is declared. so if you narrow a module level symbol it will mutate a shared symbol space; granted this should be rare and we are using a rw lock for this reason, given we are doing type checking concurrently at function level this shouldn't modify any shared symbol spaces. 

I am not fully sold on adding narrowed symbol to same space vs adding it to current space. In most cases it should be the same but to do latter I need to keep track of which symbol space we are in all the time where as former require locks

PR #234, opened 2026-03-05
Add positions to BIR instructions

### LINKED ISSUE #232: Add positions to BIR instructions

snelusha:
## Purpose
Resolves #232 

## Summary by CodeRabbit

* **Refactor**
  * Improved source-position propagation throughout the compiler IR, so diagnostics, error locations, and stack traces are more accurate for conditionals, loops, returns, calls, member/index access, literals, collections, and type conversions—resulting in clearer, more precise locations for warnings and errors.

### DIFF
--- a/bir/bir_gen.go
+++ b/bir/bir_gen.go
@@ -240,15 +240,7 @@ func TransformConstant(ctx *Context, c *ast.BLangConstant) *BIRConstant {
 	valueExpr := c.Expr
 	if literal, ok := valueExpr.(*ast.BLangLiteral); ok {
 		// FIXME: once we have constant propagation these should be propagated and no longer needed
-		return &BIRConstant{
-			BIRDocumentableNodeBase: BIRDocumentableNodeBase{
-				BIRNodeBase: BIRNodeBase{Pos: c.GetPosition()},
-			},
-			Name: model.Name(c.GetName().GetValue()),
-			ConstValue: ConstValue{
-				Value: literal.Value,
-			},
-		}
+		return NewBIRConstant(model.Name(c.GetName().GetValue()), literal.GetValueType(), literal.Value, c.GetPosition())
 	}
 	// TODO: need this think how to actually implement constant value initialization. May be we add these to init function?
 	panic("unexpected constant value type")
@@ -311,7 +303,7 @@ func compoundAssignment(ctx *stmtContext, curBB *BIRBasicBlock, stmt *ast.BLangC
 
 func continueStatement(ctx *stmtContext, curBB *BIRBasicBlock, stmt *ast.BLangContinue) statementEffect {
 	onContinueBB := ctx.loopCtx.onContinueBB
-	curBB.Terminator = NewGoto(stmt.GetPosition(), onContinueBB)
+	curBB.Terminator = NewGoto(onContinueBB, stmt.GetPosition())
 	return statementEffect{
 		// We don't know where to add the next statement so we return nil
 		block: nil,
@@ -320,7 +312,7 @@ func continueStatement(ctx *stmtContext, curBB *BIRBasicBlock, stmt *ast.BLangCo
 
 func breakStatement(ctx *stmtContext, curBB *BIRBasicBlock, stmt *ast.BLangBreak) statementEffect {
 	onBreakBB := ctx.loopCtx.onBreakBB
-	curBB.Terminator = NewGoto(stmt.GetPosition(), onBreakBB)
+	curBB.Terminator = NewGoto(onBreakBB, stmt.GetPosition())
 	return statementEffect{
 		// We don't know where to add the next statement so we return nil
 		block: nil,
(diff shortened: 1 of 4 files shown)


bir/bir_gen.go
```diff
@@ -235,6 +235,9 @@ func TransformConstant(ctx *Context, c *ast.BLangConstant) *BIRConstant {
 	if literal, ok := valueExpr.(*ast.BLangLiteral); ok {
 		// FIXME: once we have constant propagation these should be propagated and no longer needed
 		return &BIRConstant{
+			BIRDocumentableNodeBase: BIRDocumentableNodeBase{
```
warunalakshitha: why we cannot directly have pos in BirConstant itself

bir/bir_gen.go
```diff
@@ -333,7 +343,9 @@ func whileStatement(ctx *stmtContext, bb *BIRBasicBlock, stmt *ast.BLangWhile) s
 	bodyEffect := blockStatement(ctx, loopBody, &stmt.Body)
 	// This could happen if the while block always ends return, break or continue
 	if bodyEffect.block != nil {
-		bodyEffect.block.Terminator = &Goto{BIRTerminatorBase: BIRTerminatorBase{ThenBB: loopHead}}
+		gtBack := &Goto{BIRTerminatorBase: BIRTerminatorBase{ThenBB: loopHead}}
```
warunalakshitha: Shall we pass pos as first arg to terminators and non terminator constructors

bir/bir_gen.go
```diff
@@ -344,53 +356,50 @@ func whileStatement(ctx *stmtContext, bb *BIRBasicBlock, stmt *ast.BLangWhile) s
...
 
-func assignToMemberStatement(ctx *stmtContext, bb *BIRBasicBlock, varRef *ast.BLangIndexBasedAccess, valueEffect expressionEffect) statementEffect {
+func assignToMemberStatement(ctx *stmtContext, bb *BIRBasicBlock, varRef *ast.BLangIndexBasedAccess, valueEffect expressionEffect, assignPos ast.Location) statementEffect {
 	currBB := valueEffect.block
 	containerRefEffect := handleExpression(ctx, currBB, varRef.Expr)
 	currBB = containerRefEffect.block
 	indexEffect := handleExpression(ctx, currBB, varRef.IndexExpr)
 	currBB = indexEffect.block
 	fieldAccess := &FieldAccess{}
+	fieldAccess.Pos = assignPos
```
warunalakshitha: Shall we add pos as first arg

bir/bir_gen.go
```diff
@@ -907,6 +925,7 @@ func simpleVariableReference(ctx *stmtContext, curBB *BIRBasicBlock, expr *ast.B
 	if constant, ok := ctx.birCx.constantMap[symRef]; ok {
 		resultOperand := ctx.addTempVar(constant.Type)
 		constantLoad := &ConstantLoad{}
+		constantLoad.Pos = expr.GetPosition()
 		constantLoad.Value = constant.ConstValue.Value
 		constantLoad.LhsOp = resultOperand
```
warunalakshitha: since there are no continal arg setting we can move to constrcutor

bir/bir_gen.go
snelusha: `Pos` is inherited from `BIRNodeBase` which is common to all BIR nodes.

warunalakshitha: may be move to NewBirConstant

snelusha: Done in c368c6b

bir/bir_gen.go
```diff
@@ -234,26 +236,21 @@ func TransformConstant(ctx *Context, c *ast.BLangConstant) *BIRConstant {
...
-		}
+		return NewBIRConstant(c.GetPosition(), model.Name(c.GetName().GetValue()), literal.GetValueType(), literal.Value)
 	}
 	// TODO: need this think how to actually implement constant value initialization. May be we add these to init function?
 	panic("unexpected constant value type")
 }
 
 func handleBlockFunctionBody(ctx *stmtContext, ast *ast.BLangBlockFunctionBody) {
-	curBB := ctx.addBB()
+	curBB := ctx.addBB(ast.GetPosition())
```
heshanpadmasiri: Why do we have positions in basic blocks? Only instructions can trigger errors not basic blocks. If the intention is to carry information need by the bir_gen (looking at the change below) then that should be implemented properly using the context purpose of which is to carray information necessary to bir_gen.

bir/bir_gen.go
heshanpadmasiri: Personally I prefer position to be the last argument instead of first (that is the convention reset of the front end fallow).

bir/bir_gen.go
```diff
@@ -344,53 +337,50 @@ func whileStatement(ctx *stmtContext, bb *BIRBasicBlock, stmt *ast.BLangWhile) s
...
-	return assignmentStatementInner(ctx, valueEffect.block, stmt.VarRef, valueEffect)
+	return assignmentStatementInner(ctx, valueEffect.block, stmt.VarRef, valueEffect, stmt.GetPosition())
 }
 
-func assignmentStatementInner(ctx *stmtContext, bb *BIRBasicBlock, ref ast.BLangExpression, valueEffect expressionEffect) statementEffect {
+func assignmentStatementInner(ctx *stmtContext, bb *BIRBasicBlock, ref ast.BLangExpression, valueEffect expressionEffect, assignPos ast.Location) statementEffect {
 	switch varRef := ref.(type) {
 	case *ast.BLangIndexBasedAccess:
-		return assignToMemberStatement(ctx, bb, varRef, valueEffect)
+		return assignToMemberStatement(ctx, bb, varRef, valueEffect, assignPos)
```
heshanpadmasiri: NIT: order of the position must be consitent. In some places it is first and in others it's the last. Convention front end (should) fallow is diagnostic insformation is the last arg

bir/bir_gen.go
snelusha: Confirmed with @warunalakshitha, we don't need positions in basic blocks, and removed them!

bir/bir_gen.go
```diff
@@ -344,53 +335,50 @@ func whileStatement(ctx *stmtContext, bb *BIRBasicBlock, stmt *ast.BLangWhile) s
...
-	mov.RhsOp = valueEffect.result
+	mov := NewMove(valueEffect.result, refEffect.result, assignPos)
 	currBB.Instructions = append(currBB.Instructions, mov)
 	return statementEffect{
 		block: currBB,
 	}
 }
 
-func assignToSimpleVariable(ctx *stmtContext, bb *BIRBasicBlock, varRef *ast.BLangSimpleVarRef, valueEffect expressionEffect) statementEffect {
+func assignToSimpleVariable(ctx *stmtContext, bb *BIRBasicBlock, varRef *ast.BLangSimpleVarRef, valueEffect expressionEffect, assignPos ast.Location) statementEffect {
```
warunalakshitha: Shall we use same var name for position ( ex pos ast.Location). We can fix in other places as well

snelusha: Done

warunalakshitha (approved):

PR #258, opened 2026-03-11
Add golangci-lint

f-schnabel:
## Purpose
It seems like [Golangci-lint](https://golangci-lint.run/) is the most used linter of go which bundles many other linters so it might be intersting to also use it for this project.
It has a "--fix" flag that will autofix all lint and formatting errors that support autofixing.
Additionally I added a GitHub CI that runs the linter (should not be required to pass).
I ignored any errors that were not autofixable so they can be fixed in separate commits.

## Automation tests
- Unit tests
  > Code coverage information
- Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Chores**
  * Added automated lint checks via CI to enforce code quality.
  * Large-scale control-flow refactors for maintainability.
  * Optimized formatted output to reduce allocations.
  * Standardized ignoring of non-critical write/close errors to reduce noise.
  * Simplified local code via type inference and reduced boilerplate.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,4 +21,3 @@ linters:
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
-  fix: true


.golangci.yml
```diff
@@ -0,0 +1,33 @@
...
+        # Type assertion to current type
+        - -S1040
+        # redundant break statement
+        - -S1023
+        # ineffective break statement. Did you mean to break out of the outer loop?
+        - -SA4011
+        # empty branch
+        - -SA9003
+
+  disable:
```
warunalakshitha, line 19: IMO we should not disable any linters. can we remove any fix any issues

warunalakshitha: Can we remove this file so it will run on default mode

f-schnabel: Yes, sure. I just thought that you might want to have a successful ci and only incrementally enable more lints once they are fixed.
For example, it also reports unused functions which is not really useful for this early prototype.

warunalakshitha: Agree. Some of the functions are not used right now (e.g., in the parser, since we rewrote it completely), but they will be covered once we implement language features in the future. So we can keep the unused parts and other code for the moment.

f-schnabel: Another lint that was kinda annoying is that go doesnt require a break at the end of a switch branch and in many places these breaks are flagged as unnecessary. Should I fix them?

f-schnabel: On the other hand, switching this
```go
		switch nextToken {
		case NEWLINE, CARRIAGE_RETURN:
			break
		default:
			reader.Advance()
			nextToken = reader.Peek()
		}
```
to this
```go
		switch nextToken {
		case NEWLINE, CARRIAGE_RETURN:
		default:
			reader.Advance()
			nextToken = reader.Peek()
		}
```
Might make people think that NEWLINE and CARRIAGE_RETURN would fall through (as it would in Java) to the default but they don't.
This is pretty much why I did not fix them yet.

warunalakshitha: Lets fix that too

f-schnabel: I've fixed all reported warnings except the unused ones. Each fix is in its own commit, so if you'd rather not include something, I can just drop that commit.

One note: 68935bf is quite large and I used AI to help refactor it, so I can't fully vouch for correctness. That said, all tests pass.

warunalakshitha: Shall we remove https://github.com/ballerina-platform/ballerina-lang-go/commit/68935bf4d5b51001291a0dc8b189379e680cb813 commit since it is too large and will take some time to review. We can separately decide whether we should fix that or not. We can drop that commit.

f-schnabel: I have extracted https://github.com/ballerina-platform/ballerina-lang-go/commit/68935bf4d5b51001291a0dc8b189379e680cb813 into a separate branch and rebased this branch without that commit.

Issue #228, opened 2026-03-05
Add  front-end support for variable capture


heshanpadmasiri:
Fallow on from #226 and #151 . We'll use the 2D indexing from #151 to refer to the parent functions frames as need. Additionally capturing a variable should un-narrow the variable in type resolution.

PR #249, opened 2026-03-10
Implement intersection type support

heshanpadmasiri:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

 

## Summary by CodeRabbit

* **New Features**
  * Full support for intersection types (&) across parsing, printing, traversal, resolution, and diagnostics.
* **Bug Fixes**
  * Removed crashes and improved stability when handling parenthesized and intersection type expressions; added proper semantic error reporting for empty/invalid intersections.
* **Documentation / Tests**
  * Added numerous example programs and test artifacts demonstrating intersection-type behavior and expected diagnostics.

Issue #229, opened 2026-03-05
Add backend support for closures


heshanpadmasiri:
Fallow on from #228 and #151 . Probably all we need to add is treating the parent functions frame correctly in the closures frame and #151 should handle the closure updates

heshanpadmasiri:
Fixed in #264

PR #264, opened 2026-03-12
Add support for closures

heshanpadmasiri:
## Purpose
$subject
Resolves https://github.com/ballerina-platform/ballerina-lang-go/issues/228

Has an implementation for #229 as well
Depends on #249 and #255 

## Summary by CodeRabbit

* **New Features**
  * Scoped variable frames with explicit push/pop around loops and lambdas to ensure correct closure behavior.
  * Closure-aware call handling that tracks cross-function captures and marks closure loads.
  * Safer argument/operand evaluation via automatic snapshotting of temporaries.
  * Improved human-readable output for function loads and scope frames.
  * Many new example/test programs demonstrating closures and function-type behaviors.

* **Bug Fixes**
  * Fixed return routing and evaluation order for expression returns, loops, break/continue, and captured-mutable variables.


runtime/internal/exec/terminators.go
```diff
@@ -74,10 +74,15 @@ func execFpCall(callInfo *bir.Call, frame *Frame, reg *modules.Registry, callSta
 	args := extractArgs(callInfo.Args, frame, reg)
 	fnValue := getOperandValue(callInfo.FpOperand, frame, reg).(*values.Function)
 	lookupKey := fnValue.LookupKey
+	var parentFrame *Frame
+	// FIXME:
+	if fnValue.ParentFrame != nil {
```
warunalakshitha, line 78: seems this is never happens

bir/pretty_print.go
```diff
@@ -201,7 +201,11 @@ func (p *PrettyPrinter) PrintInstruction(instruction BIRInstruction) string {
 }
 
 func (p *PrettyPrinter) PrintFPLoad(fpLoad *FPLoad) string {
-	return fmt.Sprintf("%s = fp %s", p.PrintOperand(*fpLoad.LhsOp), fpLoad.FunctionLookupKey)
+	kind := "fp"
+	if fpLoad.IsClosure {
```
warunalakshitha, line 209: Shall we add a test with closure fp

heshanpadmasiri: Fixed

runtime/internal/exec/terminators.go
heshanpadmasiri: Fixed

bir/bir_gen.go
```diff
@@ -646,18 +714,53 @@ func handleExprFunctionBody(ctx *stmtContext, body *ast.BLangExprFunctionBody) {
...
+// calls that may mutate the variable through a closure.
+func snapshotIfNeeded(ctx *stmtContext, effect expressionEffect, pos ast.Location) expressionEffect {
+	op := effect.result
+	if _, isLocal := op.VariableDcl.(*BIRLocalVariableDcl); isLocal && hasNoStorageIdentity(op.VariableDcl.GetType()) {
+		tempOp := ctx.addTempVar(op.VariableDcl.GetType())
+		effect.block.Instructions = append(effect.block.Instructions, NewMove(op, tempOp, pos))
+		effect.result = tempOp
+	}
+	return effect
+}
```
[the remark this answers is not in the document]
heshanpadmasiri: Capturing variable don't create copies they share the same storage identity. Simple basic types don't have storage identity (fact they have a slot in the frame is an implementation detail)

corpus/bal/subset7/07-closure/bench-v.bal
```diff
@@ -0,0 +1,53 @@
...
+        F f = funcs[i];
+        sum += f(i);
+    }
+    return sum;
+}
+
+public function main() {
+    F[] funcs = genFs(10);
+    io:println(fSum(funcs)); // @output 75000045000000
+}
```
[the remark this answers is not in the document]
heshanpadmasiri: Skipped in integration tests

bir/bir_gen.go
```diff
@@ -364,14 +434,26 @@ func whileStatement(ctx *stmtContext, bb *BIRBasicBlock, stmt *ast.BLangWhile) s
 	// conditionally jump to loop body
 	condEffect.block.Terminator = NewBranch(condEffect.result, loopBody, loopEnd, stmt.GetPosition())
 
+	// Push scope frame for loop body — each iteration gets its own frame
+	pushScope := &PushScopeFrame{}
```
warunalakshitha, line 438: Can we resue frame object and have one frame for loop.

heshanpadmasiri: I don't think that's possible, consider case like https://github.com/heshanpadmasiri/ballerina-lang-go/blob/41f2cf8c538f11ecda2f06385a27f15c2beedbf0/corpus/bal/subset7/07-closure/bench-v.bal#L27. Each closure needs to capture a different frame

Issue #267, opened 2026-03-12
[New Feature]: Implement registration of extern functions with runtime


heshanpadmasiri:
### Description

Fallow on from #266 . We would need to build some sort of function table and statically link corresponding go functions with the interpreter. 

### Related area

Compilation

PR #268, opened 2026-03-12
Add frontend support for extern functions

### LINKED ISSUE #266: [New Feature]: Add frontend support for extern functions
### Description

$subject.

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Purpose
$subject
Resolves #266 

## Summary by CodeRabbit

* **New Features**
  * External/native functions are now recognized by the toolchain, treated as no-op bodies during analysis/walking/desugaring, and excluded from emitted runtime function sets.
* **Tests**
  * Added tests for extern function integration, including runtime integration and diagnostics for argument/return type mismatches.

### DIFF
--- a/extern-test/extern_test.go
+++ b/extern-test/extern_test.go
@@ -74,12 +74,12 @@ func TestExternValid(t *testing.T) {
 	})
 
 	// Register foo() returns "$foo"
-	runtime.RegisterExternFunction(rt, "$anon", ".", "foo", func(args []values.BalValue) (values.BalValue, error) {
+	runtime.RegisterExternFunction(rt, "$anon", "1-v", "foo", func(args []values.BalValue) (values.BalValue, error) {
 		return "$foo", nil
 	})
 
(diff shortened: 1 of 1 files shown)


heshanpadmasiri:
We need to get rid of `extern-test` and move things to corpus (probably as a project with https://github.com/ballerina-platform/ballerina-lang-go/pull/239) when we properly support #267

heshanpadmasiri:
> Shall we move extern_test inside corpus. testdata can be moved to directly with name extern.

IMO they should end up just like a normal project test (`corpus/project`). Driver should handle what ever needed to deal with extern functions automatically without special handling in the test suite. I have temporarily kept it separate because we need some hacks in test runner (I personally don't like having such hacks in the main test suites) to register the extern function until we have #267, at which point we should move it corpus. WDYT?

PR #272, opened 2026-03-17
Add corpus checks to verify node positions

snelusha:
## Purpose
$subject

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects and reports nodes with missing source positions; fills missing positions to improve diagnostic accuracy and avoid silent nil-location issues.
* **New Features**
  * IR and generated control-flow operations now include explicit source locations so errors point to original code.
  * Synthesized imports and references keep source-position info for better traceability.
* **Tests**
  * Added validations and visitors that verify positional data across compilation stages.

### DIFF
--- a/bir/bir_gen.go
+++ b/bir/bir_gen.go
@@ -1156,9 +1156,7 @@ func newExpression(ctx *stmtContext, curBB *BIRBasicBlock, expr *ast.BLangNewExp
 	}
 
 	object := ctx.addTempVar(expr.GetDeterminedType())
-	newObj := &NewObject{}
-	newObj.ClassDef = classDef
-	newObj.LhsOp = object
+	newObj := NewObjectConstructor(classDef, object, expr.GetPosition())
 	curBB.Instructions = append(curBB.Instructions, newObj)
 
 	var args []BIROperand
@@ -1171,13 +1169,8 @@ func newExpression(ctx *stmtContext, curBB *BIRBasicBlock, expr *ast.BLangNewExp
 
 	thenBB := ctx.addBB()
 	initFunc := classDef.VTable["init"]
-	call := &Call{}
-	call.Kind = INSTRUCTION_KIND_CALL
-	call.Args = args
-	call.Name = initFunc.Name
-	call.ThenBB = thenBB
 	result := ctx.addTempVar(expr.DeterminedType)
-	call.LhsOp = result
+	call := NewCall(INSTRUCTION_KIND_CALL, args, initFunc.Name, thenBB, result, expr.GetPosition())
 	call.IsVirtual = true
 	call.CachedBIRFunc = initFunc
 	curBB.Terminator = call
--- a/bir/non_terminator.go
+++ b/bir/non_terminator.go
@@ -332,6 +332,18 @@ func (n *NewObject) GetLhsOperand() *BIROperand {
 	return n.LhsOp
 }
 
+func NewObjectConstructor(classDef *BIRClassDef, lhsOp *BIROperand, pos diagnostics.Location) *NewObject {
+	return &NewObject{
+		BIRInstructionBase: BIRInstructionBase{
+			BIRNodeBase: BIRNodeBase{
+				Pos: pos,
+			},
+			LhsOp: lhsOp,
+		},
+		ClassDef: classDef,
+	}
+}
+
 func (m *MappingConstructorKeyValueEntry) IsKeyValuePair() bool {
 	return true
 }


ast/corpus_ast_test.go
```diff
@@ -91,8 +91,9 @@ func getDiff(expectedAST, actualAST string) string {
 
 // walkTestVisitor tracks node types visited during Walk traversal
 type walkTestVisitor struct {
-	visitedTypes map[string]int
-	nodeCount    int
+	visitedTypes      map[string]int
+	nodeCount         int
+	unpositionedNodes []ast.BLangNode
```
heshanpadmasiri: Why do you need to accumilate these wouldn't just T.Error be enough (mark error and continue)

ast/node_builder.go
```diff
@@ -100,7 +100,10 @@ func NewNodeBuilder(cx *context.CompilerContext) *NodeBuilder {
 }
 
 func getBuiltinPos() diagnostics.Location {
-	return nil
+	return diagnostics.NewBLangDiagnosticLocation(
```
heshanpadmasiri: I prefer this being a single object instead of allocating this for each node

bir/corpus_bir_test.go
```diff
@@ -109,3 +111,30 @@ func testBIRGeneration(t *testing.T, testPair test_util.TestCase) {
...
+	if pkg.MainFunction != nil {
+		verifyFunctionPositions(t, pkg.MainFunction, inputPath)
+	}
+}
+
+func verifyFunctionPositions(t *testing.T, fn *bir.BIRFunction, inputPath string) {
+	for _, bb := range fn.BasicBlocks {
+		for _, inst := range bb.Instructions {
+			if inst.GetPos() == nil {
+				t.Errorf("instruction %T in %s (BB: %s) has no position info for %s", inst, fn.Name, bb.Id, inputPath)
```
heshanpadmasiri, line 133: Use Error instead of Errorf we need to know all the errors and verification is isolated

desugar/corpus_desugar_test.go
```diff
@@ -44,6 +45,26 @@ func TestDesugar(t *testing.T) {
 	}
 }
 
+type walkTestVisitor struct {
+	unpositionedNodes []ast.BLangNode
```
heshanpadmasiri: same as corpus ast test

desugar/statement.go
```diff
@@ -276,24 +278,29 @@ func walkReturn(cx *FunctionContext, stmt *ast.BLangReturn) desugaredNode[model.
 }
 
 func createIncrementStmt(loopVar ast.BLangExpression) *ast.BLangAssignment {
+	basePos := loopVar.GetPosition()
```
heshanpadmasiri, line 281: I think in most of these places what you need to be able to do is set the position of a whole sub tree to a single position. If so it much more straight forward to first build the whole tree and then walk it with a visitor that update the position of any BLangNode it comes across

desugar/expression.go
```diff
@@ -308,42 +312,50 @@ func desugarCheckedExpr(cx *FunctionContext, expr *ast.BLangCheckedExpr, isPanic
...
 	tempVar.SetSymbol(tempSymbol)
 	tempVarDef := &ast.BLangSimpleVariableDef{Var: tempVar}
+	setPositionIfMissing(tempVarDef, basePos)
 	initStmts = append(initStmts, tempVarDef)
 
 	// Type test: $desugar$N is error
 	tempVarRefForTest := &ast.BLangSimpleVarRef{VariableName: tempVarName}
 	tempVarRefForTest.SetSymbol(tempSymbol)
 	tempVarRefForTest.SetDeterminedType(innerTy)
+	setPositionIfMissing(tempVarRefForTest, basePos)
```
heshanpadmasiri: I thought simply doing this once for the new ifStatement will update all of these?

desugar/statement.go
```diff
@@ -349,11 +356,13 @@ func desugarForEachOnList(cx *FunctionContext, collection ast.BLangExpression, l
 	collVar.SetInitialExpression(collExpr)
 	collVar.SetSymbol(collVarSymbol)
 	collVarDef := &ast.BLangSimpleVariableDef{Var: collVar}
+	setPositionIfMissing(collVarDef, basePos)
 	initStmts = append(initStmts, collVarDef)
 
 	collVarRef := &ast.BLangSimpleVarRef{VariableName: collVarName}
 	collVarRef.SetSymbol(collVarSymbol)
 	collVarRef.SetDeterminedType(collType)
+	setPositionIfMissing(collVarRef, basePos)
```
heshanpadmasiri: ditto. YOu should only need to do this for statements and it should update the expression in it

heshanpadmasiri (changes requested):

desugar/pos_update.go
```diff
@@ -0,0 +1,45 @@
...
+	"ballerina-lang-go/ast"
+	"ballerina-lang-go/model"
+	"ballerina-lang-go/tools/diagnostics"
+)
+
+type posUpdateVisitor struct {
+	pos diagnostics.Location
+}
+
+func (v *posUpdateVisitor) Visit(node ast.BLangNode) ast.Visitor {
```
warunalakshitha, line 29: Rather that updating pos later, should we fix in the place we create nodes.

bir/corpus_bir_test.go
warunalakshitha: is this fixed? I still see Errorf

desugar/pos_update.go
snelusha: Can’t set pos from desugar (unexported on `bLangNodeBase` in ast).

Issue #117, opened 2026-02-11
Implement serialisation and deserialisation of semtypes in BIR


heshanpadmasiri:
We need a way to serialise the semtypes and deserialize semtypes in BIR. nBallerina already have an implementation that mostly implement what we need. As far as I recall it however don't merge type environments of different BIRs

PR #275, opened 2026-03-18
Add ability to view front end time stats

heshanpadmasiri:
## Purpose
$subject
Need for #117 (we need to be able to compare the time it takes for fresh build vs from BIR)

## Summary by CodeRabbit

* **New Features**
  * Per-module, per-stage compilation timing with formatted and one-line reports; CLI flags to emit these (--stats, --stats-oneline) and profiler flags to write CPU/memory profiles (--cpuprofile, --memprofile).
  * Exposed package-level methods to retrieve formatted stats reports.
  * Added a small example program demonstrating record/union handling and main entry.

* **Chores**
  * Build option plumbing added to enable/propagate stats.

* **Tests**
  * Tests updated and new tests added to cover stats formatting and the updated environment initialization.

### DIFF
--- a/projects/compilation_stats.go
+++ b/projects/compilation_stats.go
@@ -40,13 +40,7 @@ var analysisStages = []context.CompilationStage{
 	context.StageDesugaring,
 }
 
-func formatStatsReport(moduleList []*moduleContext) string {
-	var allStats []*context.ModuleStats
-	for _, m := range moduleList {
-		if s := m.compilerCtx.GetModuleStats(); s != nil {
-			allStats = append(allStats, s)
-		}
-	}
+func formatStatsReport(allStats []*context.ModuleStats) string {
 	if len(allStats) == 0 {
 		return ""
 	}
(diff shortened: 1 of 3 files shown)


context/context.go
```diff
@@ -145,3 +180,31 @@ func (this *CompilerContext) GetTypeEnv() semtypes.Env {
 func (this *CompilerContext) GetNextAnonymousTypeKey(packageID *model.PackageID) string {
 	return this.env.GetNextAnonymousTypeKey(packageID)
 }
+
+func (this *CompilerContext) InitModuleStats(moduleName string) {
+	if !this.env.statsEnabled {
```
warunalakshitha, line 193: AFAIU this will make sure counting stats will not affect the normal compilation without any flags

heshanpadmasiri: Mostly no effect given but we will be checking this branch always. Ideally I think this should be behind a build tag so it is not evaluated in release version, but I think effect should be minuscule

PR #313, opened 2026-03-25
Fix corpus test stderr comparison

### LINKED ISSUE #293: [Bug]: Fix how we handle corpus errors
### Description

We do most of the analysis that produce error messages concurrent and error messages are shown in the order they were detected. So we run in to false positives like this
Image

I think the correct solution is to have the corpus test acutally validate it has the same errors as the expected errors instead of a simple string comparision.

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

snelusha:
## Purpose
Resolves #293

## Summary by CodeRabbit

* **Tests**
  * Enhanced integration test stderr comparison with normalization logic for more reliable test result validation
  * Re-enabled one previously skipped test case

### DIFF
--- a/corpus/integration/project/visibility-e.txtar
+++ b/corpus/integration/project/visibility-e.txtar
@@ -11,4 +11,3 @@ error[SEMANTIC_ERROR]: Unknown symbol: greet
    |
 23 |     io:println(utils:greet("Ballerina")); // @error
    |                ^^^^^^^^^^^^^^^^^^^^^^^^
-


corpus/integration_test.go
```diff
@@ -171,6 +173,32 @@ func formatExpectedGot(expected, got string) string {
...
+	for part := range strings.SplitSeq(stderr, "\n\n") {
+		diagnostic := strings.TrimSpace(part)
+		if diagnostic != "" {
+			diagnostics = append(diagnostics, diagnostic)
+		}
+	}
+	return diagnostics
+}
+
+func normalizeIntegrationStderr(stderr string) string {
```
warunalakshitha, line 194: Shall we order these functions based on call order.

corpus/integration_test.go
```diff
@@ -457,3 +346,147 @@ func runProjectInterpretPhase(birPkgs []*bir.BIRPackage, stdoutBuf, stderrBuf *b
...
+	stderr = strings.TrimSpace(stderr)
+	if stderr == "" {
+		return ""
+	}
+
+	diagnostics := splitStderrDiagnostics(stderr)
+	if len(diagnostics) == 0 {
```
warunalakshitha: can this be happen if std != ""

PR #319, opened 2026-03-25
Skip corpus tests for unimplemented parser features

f-schnabel:
## Summary
- Convert panics with "parser not implemented" suffix to test skips instead of fatal failures
- Preserves existing fatal behavior for all other panics

## Test plan
- [ ] Run `go test ./parser/...` and verify unimplemented parser tests are skipped, not failed
- [ ] Verify other panics still cause test failures

## Summary by CodeRabbit

* **Tests**
  * Enhanced test behavior for unimplemented parser cases, allowing tests to skip gracefully rather than fail when parser functionality is not yet available.

### DIFF
--- a/parser/corpus_parser_test.go
+++ b/parser/corpus_parser_test.go
@@ -330,7 +330,11 @@ func parseFile(t *testing.T, testCase test_util.TestCase) {
 	// Catch any panics and convert them to errors
 	defer func() {
 		if r := recover(); r != nil {
-			t.Fatalf("panic: %v", r)
+			if msg, ok := r.(string); ok && strings.HasSuffix(msg, "parser not implemented") {
+				t.Skipf("Skipping file due to unimplemented parser feature: %s", testCase.InputPath)
+			} else {
+				t.Fatalf("panic: %v", r)
+			}
 		}
 	}()
 


parser/corpus_parser_test.go
```diff
@@ -330,7 +330,11 @@ func parseFile(t *testing.T, testCase test_util.TestCase) {
 	// Catch any panics and convert them to errors
 	defer func() {
 		if r := recover(); r != nil {
-			t.Fatalf("panic: %v", r)
+			if msg, ok := r.(string); ok && strings.HasSuffix(msg, "parser not implemented") {
+				t.Skipf("Skipping file due to unimplemented parser feature: %s", testCase.InputPath)
+			} else {
+				t.Fatalf("panic: %v", r)
```
snelusha, line 336: Ideally we should not skip tests silently when catching a panic, could we use `t.Errorf` here instead so it marks the failure and continues?

f-schnabel: But should a test that is not implemented yet fail the whole testrun? I just thought that you then need to go through each failing test and see if it is a regression or if its not implemented yet. It was just an idea and doesnt need to be merged if its not wanted.

snelusha: `t.Errorf` won't fail the whole test run; it marks the test as failed but continues running the rest, so you can see all failing tests in one go instead of stopping at the first panic.

https://pkg.go.dev/testing#T.Errorf

Issue #139, opened 2026-02-15
Make type cast fully support the current subset


heshanpadmasiri:
#132 will be necessary to implement this. We need to support fallowing

1. Numeric conversion should support types that are subsets for `int`, `float`, `decimal`
2. Casting list
3. Casting map

For both 2 and 3 (may be for 1 as well) we shoud be able to implement a generic version that uses the type of the value and just check if it a subtype of the target type

heshanpadmasiri:
This should also fix things like 
```
import ballerina/io;

public function main() {
    any[] days = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"];
    map<any> dayNumber = {};
    foreach int i in 0 ..< days.length() {
        dayNumber[<string>days[i]] = i + 1;
    }
    io:println(dayNumber["Sunday"]); // @output 7
}
```

heshanpadmasiri:
Fixed in main

PR #321, opened 2026-03-25
[Runtime] Add runtime support for trap and check expressions

### LINKED ISSUE #260: [Bug]: Type case don't work when target type is not a basic type
### Description

```ballerina
import ballerina/io;

public function main() {
    byte f = foo(1, 2);
    io:println(f); // @output 3
}

function foo(int x, int y) returns byte {
    return <byte>(x + y);
}
```
 Here target type don't necessarily needs to be a basic type https://github.com/ballerina-platform/ballerina-lang-go/blob/263c54fd37ed476f1c63cc968a5a02b39101af0d/runtime/internal/exec/non_terminators.go#L132

Instead first we need to check if the value type is subtype if so just return the value type. Otherwise we need to check if we need to do numeric conversion and do it. Relevant section of the [spec](https://ballerina.io/spec/lang/master/#section_6.20)

> A type-cast-expr is evaluated by first evaluating expression resulting in a value v. Let T be the type described by type-descriptor. If v belongs T, then the result of the type-cast-expr is v. Otherwise, if T includes shapes from exactly one basic numeric type N and v belongs to another basic numeric type, then let n be NumericConvert(N, v); if n is not an error and n belongs to T, then the result of the type-cast-expr is n. Otherwise, the evaluation of the type-cast-expr completes abruptly with a panic.

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Runtime

_No response_

_No response_

_No response_

### LINKED ISSUE #320: Add back end support for check and trap expressions
### Description

$Subject

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/320
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/260

## Summary by CodeRabbit

* **New Features**
  * Improved trap/error handling so nested trap expressions reliably convert and resume with error values; error-table reporting now references block indices for clearer diagnostics.

* **Bug Fixes**
  * More specific numeric cast diagnostics (out-of-range vs non-finite).
  * Prevented panics when reporting empty stack traces.
  * Fixed nil handling in equality comparisons.

* **Tests**
  * Added numerous integration fixtures and re-enabled previously skipped error tests.


corpus/integration/subset5/05-error/trap2-v.txtar
```diff
@@ -0,0 +1,3 @@
+-- stdout --
+error("divide by zero")
+-- stderr --
```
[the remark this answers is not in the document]
warunalakshitha: integration output assertion will not add source. it will assert only actual output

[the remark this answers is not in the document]
warunalakshitha: can we resolve @coderabbitai

corpus/integration/subset5/05-error/check6-e.txtar
```diff
@@ -0,0 +1,20 @@
...
+  |
+2 |     foo(check error("error")); // @error
+  |         ^^^^^^^^^^^^^^^^^^^^
+
+error[SEMANTIC_ERROR]: error type of check expression is not a subtype of the enclosing function's return type
+ --> check6-e.bal:2:9
+  |
+2 |     foo(check error("error")); // @error
+  |         ^^^^^^^^^^^^^^^^^^^^
+
```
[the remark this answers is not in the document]
warunalakshitha: @coderabbitai can u create seperate issue for this?

corpus/integration/subset5/05-error/check7-e.txtar
```diff
@@ -0,0 +1,14 @@
...
+ --> check7-e.bal:3:15
+  |
+3 | const int X = check 1; // @error
+  |               ^^^^^^^
+
+error[SEMANTIC_ERROR]: expression is not a constant expression
+ --> check7-e.bal:3:7
+  |
+3 | const int X = check 1; // @error
+  |       ^^^
```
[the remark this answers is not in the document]
warunalakshitha: @coderabbitai can u create seperate issue for this?

bir/bir_gen.go
```diff
@@ -276,6 +276,26 @@ func transformFunction(ctx *Context, astFunc *ast.BLangFunction, selfSymbolRef *
...
+			if entry.Start != nil {
+				entry.Start = bbMap[entry.Start.Id.Value()]
+			}
+			if entry.End != nil {
+				entry.End = bbMap[entry.End.Id.Value()]
+			}
+			if entry.Target != nil {
+				entry.Target = bbMap[entry.Target.Id.Value()]
+			}
+		}
```
heshanpadmasiri: It is not clear to me why this change is needed. Yes they are pointing to different BBs (since we copy them) but runtime uses Number that should be the same (generally BBs should be safe to be copied; though in-efficient). Actually in retrospect given runtime uses the numbers may be it is better to store the number instead of BB (we should do that at point of creation)

bir/model.go
```diff
@@ -125,6 +125,7 @@ type (
 		Parameters     []BIRFunctionParameter
 		BasicBlocks    []BIRBasicBlock
 		ErrorTable     []BIRErrorEntry
+		HasErrorTable  bool
```
heshanpadmasiri: I am entierly convienced having this field makes much of a difference compared to checking len == 0. (Length of slices are inilned in the struct and both are effectively branch if zero).

runtime/internal/exec/non_terminators.go
```diff
@@ -150,10 +151,9 @@ func castValue(value values.BalValue, targetType semtypes.SemType) values.BalVal
 	case bitsetValue&semtypes.DECIMAL.All() != 0:
 		return toDecimal(value)
 	case bitsetValue&semtypes.BOOLEAN.All() != 0:
-		return value.(bool)
+		return toBoolean(value)
 	}
```
heshanpadmasiri: I think you can do semtypes.IsSubtypeSimple instead of direct bit manipulations here

runtime/internal/exec/non_terminators.go
```diff
@@ -150,10 +151,9 @@ func castValue(value values.BalValue, targetType semtypes.SemType) values.BalVal
 	case bitsetValue&semtypes.DECIMAL.All() != 0:
 		return toDecimal(value)
 	case bitsetValue&semtypes.BOOLEAN.All() != 0:
-		return value.(bool)
+		return toBoolean(value)
```
heshanpadmasiri: I am not sure this branch is correct (or reachable). I assume this part is partially doing NumericConvert, which is not applicable for booleans

heshanpadmasiri (commented):
Generally LGTM other than few nit picks. I think we can address the type cast related ones seperately when we do #139

bir/model.go
warunalakshitha: Agree. I added thinking it will have impact but since inlined not much perf impact.  Removed the HasErrorTable

runtime/internal/exec/non_terminators.go
warunalakshitha: Removed. not needed.

bir/bir_gen.go
warunalakshitha: Refactored to store number.

runtime/internal/exec/non_terminators.go
warunalakshitha: Fixed

warunalakshitha: We can revisit logic when fixing https://github.com/ballerina-platform/ballerina-lang-go/issues/139

runtime/internal/exec/non_terminators.go
```diff
@@ -221,3 +216,10 @@ func toDecimal(value any) *big.Rat {
 		panic(values.NewErrorWithMessage(fmt.Sprintf("bad type cast: cannot cast %v to decimal", value)))
 	}
 }
+
+func toBoolean(value any) bool {
+	if v, ok := value.(bool); ok {
+		return v
+	}
+	panic(values.NewErrorWithMessage(fmt.Sprintf("bad type cast: cannot cast %v to boolean", value)))
+}
```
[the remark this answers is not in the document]
warunalakshitha: Removed

heshanpadmasiri (approved):
LGTM

PR #355, opened 2026-04-07
 Add support for named argumets

### LINKED ISSUE #335: Add support for named argumets
### Description

I think the most straightforward way to implement this is to desugar named arguments to positional arguments.  However at the semantic analysis we need to detect if we have a named arg and positional arg both for same argument. 

Also note type don't contain parameter names so you can't use projection to determine the type of a named argument. This would introduce complications with allowing named arguments for function values (value don't have type descriptor only type) So we will defer implementing support for function values (I don't think this is common enough to spend cycles immediately) 

Given we desugar these away this shouldn't have any backend implications

heshanpadmasiri:
## Purpose
$subject
Depends on #340 

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive named-argument support for function calls, including cross-module scenarios.

* **Bug Fixes**
  * More precise invocation diagnostics (missing/undefined parameter, duplicate named args, named-then-positional error).
  * Corrected default-parameter evaluation order.

* **Tests**
  * Added multiple example programs and integration fixtures covering valid and error cases for named arguments.


lib/map/compile/map.go
```diff
@@ -62,7 +62,7 @@ func GetMapSymbols(ctx *context.CompilerContext) model.ExportedSymbolSpace {
 	ctx.SetSymbolType(keysRef, libcommon.FunctionSignatureToSemType(env, &keysSignature))
 
 	// remove: generic (mapType, STRING) -> memberType
-	removeSymbol := model.NewGenericFunctionSymbol("remove", space, createRemoveMonomorphizer(ctx))
+	removeSymbol := model.NewGenericFunctionSymbol("remove", space, []string{"m", "k"}, createRemoveMonomorphizer(ctx))
```
warunalakshitha, line 65: Why this change?

heshanpadmasiri: All function symbols now expect you to provide param names, so that you can use named arguments

warunalakshitha (approved):
LGTM

PR #374, opened 2026-04-09
Implement benchmark tool

### LINKED ISSUE #312: Benchmark tool
### Description

It's good to have a way to quickly compare the performance differences between two versions of the compiler. I think we can add a new compiler tool that takes two git refs (commits / tag I assume we can implement both in the same way) build the compiler and run all the `.bal` files in a given folder in both versions and give breakdown on time it took for each versions. 
For starters it can give the total time (frontend + backend) and with #275 may be breakdown for front end stat differences as well. I think we need to have flags to control the nature of the output.  For single files stdout is enough but when handling multiple files I think some sort of HTML report is better
For the target we can support local (for quick dev comparisions) and cloud like aws where it expect user to provide keys + configs and tool will provision VM and run the benchmarks. Latter is needed for reproducible benchmarks

Later we can create github workflows that uses this tool to run either periodical (my preference) or on release benchmark comparisions 

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Other Area

_No response_

_No response_

_No response_

snelusha:
## Purpose
Resolves #312

Sample HTML Report

image

## Summary by CodeRabbit

* **New Features**
  * CLI benchmark tool to compare performance between two Git refs with warmup/iteration options, required positional refs/target, and optional HTML export.
  * Generates an HTML report with per-case metrics, statistical deltas, and winner indication.
  * Supports single file, package, or multi-file directory targets and includes runtime input validation.

* **Tests**
  * Added an integration test that runs the benchmark and verifies HTML export output.

* **Chores**
  * Added module metadata for the benchmark tool.

### DIFF
--- a/compiler-tools/benchmark/config.go
+++ b/compiler-tools/benchmark/config.go
@@ -28,7 +28,6 @@ type config struct {
 	target     string
 	warmup     int
 	runs       int
-	export     bool
 	exportPath string
 }
 
@@ -61,9 +60,6 @@ func parseConfig() (*config, error) {
 		runs:       *runs,
 		exportPath: *exportPath,
 	}
-	if cfg.exportPath != "" {
-		cfg.export = true
-	}
 	return cfg, nil
 }
 
@@ -73,9 +69,6 @@ func (c *config) validate() error {
 	}
 	if c.headRef == "" {
 		return fmt.Errorf("headRef is required")
-	}
-	if c.baseRef == c.headRef {
-		return fmt.Errorf("baseRef and headRef cannot be the same")
 	}
 	if c.target == "" {
 		return fmt.Errorf("target is required")
@@ -89,7 +82,7 @@ func (c *config) validate() error {
 	if c.runs <= 0 {
 		return fmt.Errorf("runs must be greater than zero")
 	}
-	if c.export && c.exportPath == "" {
+	if c.exportPath == "" {
 		return fmt.Errorf("export path is required")
 	}
 	return nil
--- a/compiler-tools/benchmark/config.go
+++ b/compiler-tools/benchmark/config.go
@@ -1,0 +1,85 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+)
+
+type config struct {
+	baseRef string
+	headRef string
+	target  string
+	warmup  int
+	runs    int
+}
+
+func parseConfig() (*config, error) {
+	fs := flag.NewFlagSet(os.Args[0], flag.ContinueOnError)
+	fs.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage: %s [options] <base-ref> <head-ref> <target>\n", os.Args[0])
+		fs.PrintDefaults()
+	}
+
+	warmup := fs.Int("warmup", 0, "Number of warmup iterations")
+	runs := fs.Int("runs", 10, "Number of benchmark runs")
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		return nil, err
+	}
+
+	args := fs.Args()
+	if len(args) < 3 {
+		fs.Usage()
+		return nil, fmt.Errorf("missing required arguments")
+	}
+
+	return &config{
+		baseRef: args[0],
+		headRef: args[1],
+		target:  args[2],
+		warmup:  *warmup,
+		runs:    *runs,
+	}, nil
+}
+
+func (c *config) validate() error {
+	if c.baseRef == "" {
+		return fmt.Errorf("baseRef is required")
+	}
+	if c.headRef == "" {
+		return fmt.Errorf("headRef is required")
+	}
+	if c.baseRef == c.headRef {
+		return fmt.Errorf("baseRef and headRef cannot be the same")
+	}
+	if c.target == "" {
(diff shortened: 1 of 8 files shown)


compiler-tools/benchmark/runner.go
```diff
@@ -0,0 +1,218 @@
...
+		BaseRef:   b.baseRef,
+		HeadRef:   b.headRef,
+		Generated: time.Now(),
+		results:   results,
+	}
+	if err := report.export(b.config.exportPath); err != nil {
+		return fmt.Errorf("failed to export report: %w", err)
+	}
+	fmt.Printf("Benchmark report exported to %s\n", b.config.exportPath)
+	return nil
```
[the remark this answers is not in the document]
snelusha: This is enough for now!

warunalakshitha:
Shall we write some tests for the benchmark tool. wonder why does not have any patch coverage for the tool

compiler-tools/benchmark/go.mod
```diff
@@ -0,0 +1,3 @@
+module balbench
```
warunalakshitha, line 1: Why seperate go mod

warunalakshitha: i think all tools should be part of root module

compiler-tools/benchmark/config.go
```diff
@@ -0,0 +1,96 @@
...
+	"os"
+)
+
+type config struct {
+	baseRef    string
+	headRef    string
+	target     string
+	warmup     int
+	runs       int
+	export     bool
```
heshanpadmasiri: Why do you need seperate expor flag here?

compiler-tools/benchmark/config.go
```diff
@@ -0,0 +1,96 @@
...
+func (c *config) validate() error {
+	if c.baseRef == "" {
+		return fmt.Errorf("baseRef is required")
+	}
+	if c.headRef == "" {
+		return fmt.Errorf("headRef is required")
+	}
+	if c.baseRef == c.headRef {
+		return fmt.Errorf("baseRef and headRef cannot be the same")
+	}
```
heshanpadmasiri: I don't see a reason why we sholud restrict this

compiler-tools/benchmark/export.go
```diff
@@ -0,0 +1,343 @@
...
+	if rawRatio < 1.0 {
+		winnerRef = baseRef
+		ratio = 1.0 / rawRatio
+		// If q = 1/r then dq = dr / r^2.
+		ratioErr = rawRatioErr / (rawRatio * rawRatio)
+	}
+	return true, fmt.Sprintf("%.2f", ratio), fmt.Sprintf("%.2f", math.Abs(ratioErr)), winnerRef
+}
+
+const htmlTemplate = `<!doctype html>
```
heshanpadmasiri: I prefer template to be a seperate file instead

compiler-tools/benchmark/export.go
```diff
@@ -0,0 +1,343 @@
...
+		last := &rows[len(rows)-1]
+		if base != nil {
+			last.BaseMean = fmt.Sprintf("%.3f", base.Mean*1000.0)
+			last.BaseStddev = fmt.Sprintf("%.3f", base.Stddev*1000.0)
+		}
+		if head != nil {
+			last.HeadMean = fmt.Sprintf("%.3f", head.Mean*1000.0)
+			last.HeadStddev = fmt.Sprintf("%.3f", head.Stddev*1000.0)
+		}
+		last.DeltaAvailable, last.DeltaRatio, last.DeltaErr, last.DeltaWinnerRef = computeDelta(base, head, r.BaseRef, r.HeadRef)
```
heshanpadmasiri: What is the purpose of this. AFAIK you are not reading last anywhere

compiler-tools/benchmark/export.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	tpl := template.Must(template.New("report").Parse(htmlTemplate))
+	f, err := os.Create(outPath)
+	if err != nil {
+		return fmt.Errorf("create html report %q: %w", outPath, err)
```
heshanpadmasiri: shouldn't this say something like error creating html report

compiler-tools/benchmark/export.go
```diff
@@ -0,0 +1,343 @@
...
+	data := struct {
+		Report report
+		Rows   []row
+	}{
+		Report: *r,
+		Rows:   rows,
+	}
+
+	if err := tpl.Execute(f, data); err != nil {
+		return fmt.Errorf("render html report %q: %w", outPath, err)
```
heshanpadmasiri: ditto

compiler-tools/benchmark/go.mod
```diff
@@ -0,0 +1,3 @@
...
+
+go 1.25.0
```
heshanpadmasiri: Everything should use a same go version. I think interpreter is 1.26

heshanpadmasiri (commented):
Can you attach a sample html with the PR as well

compiler-tools/benchmark/go.mod
snelusha: > Why seperate go mod

This should be separate executable

compiler-tools/benchmark/export.go
```diff
@@ -0,0 +1,145 @@
...
+	winnerRef := headRef
+	ratio := rawRatio
+	ratioErr := rawRatioErr
+	if rawRatio < 1.0 {
+		winnerRef = baseRef
+		ratio = 1.0 / rawRatio
+		// If q = 1/r then dq = dr / r^2.
+		ratioErr = rawRatioErr / (rawRatio * rawRatio)
+	}
+	return true, fmt.Sprintf("%.2f", ratio), fmt.Sprintf("%.2f", math.Abs(ratioErr)), winnerRef
```
heshanpadmasiri: What is the statistic we are calculating here?

snelusha: We are calculating speedup ratio here.
Hyperfine's export json does not provide that.

compiler-tools/benchmark/export.go
```diff
@@ -0,0 +1,156 @@
...
+		Rows   []row
+	}{
+		Report: *r,
+		Rows:   rows,
+	}
+
+	if err := tpl.Execute(f, data); err != nil {
+		return fmt.Errorf("failed to render html report %q: %w", outPath, err)
+	}
+	return nil
```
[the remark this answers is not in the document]
snelusha: I think we don't need to capture or display the error during the defer for this specific use case.

heshanpadmasiri (approved):
LGTM

PR #379, opened 2026-04-16
Aggregate Codecov coverage from individual Go modules into root report

warunalakshitha:
## Purpose
ATM only root module tests are added to Codecov. This will cover all individual Go modules

## Summary by CodeRabbit

* **Tests**
  * Expanded fixture-driven CLI and generation tests for the tree-gen tool, covering success paths and many error cases.
  * Added integration-style coverage hooks to capture per-run coverage from test binaries.

* **Chores**
  * CI updated to run stricter cross-platform test workflows and aggregate per-module coverage for upload.
  * Coverage settings changed to include previously excluded test utilities.
  * Added an upstream tooling dependency and made benchmark output visible.


compiler-tools/benchmark/runner.go
```diff
@@ -214,6 +195,9 @@ func sanitize(ref string) string {
 }
 
 func shellQuote(s string) string {
+	if runtime.GOOS == "windows" {
+		return `"` + strings.ReplaceAll(s, `"`, `\"`) + `"`
```
warunalakshitha, line 193: benchmark tool fails in windows.this will fix it

compiler-tools/benchmark/config.go
```diff
@@ -83,7 +83,7 @@ func (c *config) validate() error {
 		return fmt.Errorf("runs must be greater than zero")
 	}
 	if c.exportPath == "" {
```
snelusha: `--export-html` should be an optional flag. Requiring it here was a mistake from my side.

compiler-tools/benchmark/runner.go
```diff
@@ -21,15 +21,13 @@ import (
...
 	"strconv"
 	"strings"
 	"time"
 )
 
-const (
-	baseInterpreter = "bal-base"
-	headInterpreter = "bal-base"
-)
+const builtInterpreter = "bal-base"
```
snelusha: This can be just `bal` right?

compiler-tools/benchmark/runner.go
```diff
@@ -154,42 +136,35 @@ func (b *benchmark) buildInterpreter(worktreePath, ref, output string) error {
...
+	args = append(args, "--runs", strconv.Itoa(b.runs))
 	return args
 }
 
 func (b *benchmark) runHyperfine(cmds []string, jsonExportPath string) (*benchExport, error) {
 	args := b.hyperfineFlags()
-	if b.config.exportPath != "" {
-		args = append(args, "--export-json", jsonExportPath)
-	}
+	args = append(args, "--export-json", jsonExportPath)
```
snelusha, line 149: Do we need to export JSON files if we're not generating an HTML report?

warunalakshitha: if we are not generating json what is default use case of tool?

snelusha: We are able to see results directly in the terminal

warunalakshitha: will revert then

heshanpadmasiri (approved):
LGTM

compiler-tools/benchmark/runner.go
warunalakshitha: Will fix this as seperate PR.

PR #380, opened 2026-04-16
Optimize runtime memory usage for type checker

heshanpadmasiri:
## Purpose
$subject

## Summary by CodeRabbit

* **Refactor**
  * Consolidated execution state into a shared Context used throughout the runtime.
  * Threaded the Context through execution paths and stack management for cleaner control-flow handling.
  * Internal operand and error stack handling routed via the Context.
  * No public APIs or observable behavior changes for users; runtime semantics remain unchanged.

PR #382, opened 2026-04-17
Change wasm GC and heap bounds

heshanpadmasiri:
## Purpose
Related to #341 (This is not a proper fix AFAIK but along with improvements to memory usage this seems to reduce the significantly reduce the likelyhood of hitting the GC bug)
Depends on #381 and #380


warunalakshitha (approved):
LGTM

PR #383, opened 2026-04-18
Fix float equality and division semantics in runtime

### LINKED ISSUE #364: [Bug]: Float and decimal division should never cause a panic
### Description

Relevant section of the [spec](https://ballerina.io/spec/lang/master/#section_6.23)

> float, decimal

> * performs the multiplication operation with the destination format being the same as the source format, as defined by IEEE 754-2008; this never causes a panic to occur
> / performs the division operation with the destination format being the same as the source format, as defined by IEEE 754-2008; this never causes a panic to occur
> % performs a remainder operation; the remainder is not the IEEE-defined remainder operation but is instead a remainder operation analogous to integer remainder; more precisely,
> if x is NaN or y is NaN or x is an infinity or y is a zero, then x % y is NaN
> for finite x, and infinite y, x % y is x
> for finite x and finite non-zero y, x % y is equal to the result of rounding x - (y × n) to the nearest representable value using the roundTiesToEven rounding mode, where n is the integer that is nearest to the mathematical quotient of x and y without exceeding it in magnitude; if the result is zero, then its sign is that of x
> no exceptions are thrown

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #391: [Bug]: Float ==/!= semantics were wrong for NaN in runtime
### Description

$Subject

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/364
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/391

## Summary by CodeRabbit

* **Bug Fixes**
  * Floating-point division and modulo no longer panic on divide-by-zero; results follow IEEE‑754 (Inf/NaN).

* **Behavior Changes**
  * Equality and reference comparisons revised for consistent NaN, signed-zero, and infinity handling.

* **New Features**
  * Added example programs illustrating IEEE‑754 edge cases and strict-equality behaviors.

* **Tests**
  * Updated and added integration fixtures to match new IEEE‑754 outputs; previous divide-by-zero failure expectations removed.


values/compare.go
```diff
@@ -88,14 +66,64 @@ func Compare(x, y BalValue) CompareResult {
...
+	} else if a == b {
+		return CmpEQ
+	}
+	if a < b {
+		return CmpLT
+	}
+	if a > b {
+		return CmpGT
+	}
+	return CmpUN
```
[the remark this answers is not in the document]
warunalakshitha: @coderabbitai compareFloat in values/compare.go already returns CmpUN when either operand is NaN, which is the correct ordering behavior

values/compare.go
```diff
@@ -88,14 +67,88 @@ func Compare(x, y BalValue) CompareResult {
...
+		return false
+	}
+	if xf, ok := x.(float64); ok {
+		yf := y.(float64)
+		if math.IsNaN(xf) && math.IsNaN(yf) {
+			return true
+		}
+	}
+	return Compare(x, y) == CmpEQ
+}
```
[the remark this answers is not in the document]
warunalakshitha: @coderabbitai structural types will be covered in separate pr.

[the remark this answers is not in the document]
warunalakshitha: @coderabbitai https://github.com/ballerina-platform/ballerina-lang-go/issues/394

heshanpadmasiri:
Shall we add test to validate nans are propagated across arithmetic operations as well (I assume this should just work)
``` ballerina
import ballerina/io;

public function main() {
    float nan1 = 0.0 / 0.0;
    float x = 100 * nan1;
    io:println(x == nan1);
}
```

corpus/bal/subset3/03-float/nan1-v.bal
```diff
@@ -0,0 +1,25 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+// @productions multiplicative-expr function-call-expr local-var-decl-stmt float-literal equal-expr
+import ballerina/io;
+
+public function main() {
+    float nan1 = 0.0 / 0.0;
+    float x = 100.0 * nan1;
+    io:println(x == nan1); // @output false
```
heshanpadmasiri, line 23: this should true?

values/compare.go
```diff
@@ -113,3 +147,20 @@ func compareList(x, y *List) CompareResult {
 	}
 	return CmpEQ
 }
+
+func compareFloat(a, b float64) CompareResult {
```
heshanpadmasiri, line 151: This should have a EQ case

values/compare.go
```diff
@@ -31,55 +32,32 @@ const (
 )
 
 func Compare(x, y BalValue) CompareResult {
-	if x == nil && y == nil {
+	if x == y {
```
heshanpadmasiri, line 35: Not sure if it's something we need to prevent (given fronend should detect these cases and give errors already) but this would allow comparision of things like map, table etc as long as they are the same pointer (EQ case). Maybe we should add a comment here at least saying it assume both values should belong to Comparable

values/compare.go
```diff
@@ -88,14 +66,70 @@ func Compare(x, y BalValue) CompareResult {
...
+	switch xv := x.(type) {
+	case *Function:
+		yFn, ok := y.(*Function)
+		return ok && xv.LookupKey == yFn.LookupKey
+	case float64:
+		yf, ok := y.(float64)
+		return ok && compareFloatRef(xv, yf)
+	case *big.Rat:
+		yDec, ok := y.(*big.Rat)
+		return ok && xv.Cmp(yDec) == 0
```
heshanpadmasiri, line 101: According to the spec 
"Note that === and == are the same for simple values except for floating point types.

For the float type, the operators differ as regards floating point zero: == treats positive and negative zero from the same basic type as equal whereas === treats them as unequal. Both == and === treat two NaN values from the same basic type as equal. This means that neither of these operators correspond to operations defined by IEEE 754-2008, because they do not treat NaN in the special way defined for those operations.

For the decimal type, the operators differ in whether they consider the precision of the value. For example, 1.0 == 1.00 is true but 1.0 === 1.00 is false.
"

heshanpadmasiri:
We need to check if these are fixed by the as well https://github.com/ballerina-platform/ballerina-lang-go/blob/0a5e463d88174fe1a15c35ce4d3e4bcd165658ac/test_util/skip.go#L154. We can create separate issue to track those that are not

warunalakshitha:
Sent a new PR with https://github.com/ballerina-platform/ballerina-lang-go/pull/458

PR #393, opened 2026-04-20
Implement initial version for order by and on conflict for lists and maps

### LINKED ISSUE #392: [Improvement]: Implement order by and on conflict clauses for lists and maps
### Description

Query expressions dont support order by and on conflict clauses at the moment. Currently query expressions support only lists and maps. We need to support order by and on conflicts for lists and maps. Later we can expand to other collections.

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

gimantha:
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/392

## Automation tests
- Unit tests
  > Code coverage information
- Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Query expressions: support for order-by sorting (ascending/descending) across primitives, nullable values, and nested lists
  * Map queries: support for on-conflict handlers to resolve duplicate-key cases with nil or error outcomes
  * Added a dedicated sorting implementation to efficiently order query results

* **Tests**
  * Added comprehensive unit and integration tests and fixtures covering order-by and on-conflict behaviors and runtime comparisons


desugar/expression.go
```diff
@@ -970,28 +923,586 @@ func walkQueryExpr(cx *functionContext, expr *ast.BLangQueryExpr) desugaredNode[
...
+	stageStores = append(stageStores, newLetStores...)
+	payloadRowsRef, ok := createQueryPayloadStore(cx, initStmts, basePos, stageStores)
+	if !ok {
+		return queryOrderStageInput{}, false
+	}
+
+	loopCounterRef := createQueryCounterRef(cx, initStmts, basePos)
+	baseIndexExpr := queryStageBaseIndexExpr(loopCounterRef, stageInput.indexRowsRef)
+	elementAccess := queryElementAccess(collRef, keysRef, baseIndexExpr, loopVarTy)
+	if elementAccess == nil {
```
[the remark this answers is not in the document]
gimantha: No fix needed here.

I checked the current lowering in desugar/expression.go (line 1117) and the runtime sort in lib/array/runtime/array.go (line 127). sortedIndexRowsRef is intentionally carrying original row IDs, not current-stage positions. querySort does not interpret indices semantically; it just reorders that list in parallel with keys and payloads. That means pushing baseIndexExpr is correct: after each stage, we still have a mapping from current sorted position back to the original collection row.

That original-row lineage is required by the next stage and by the final stage in desugar/expression.go (line 1250), where queryStageBaseIndexExpr uses stageInput.indexRowsRef to fetch the actual collection element from collRef. If we switched later ORDER BY stages to store loopCounterRef instead, we would lose that lineage and final element access would be wrong unless we introduced a separate parallel lineage structure.

lib/array/runtime/array.go
```diff
@@ -25,6 +43,145 @@ func initArrayModule(rt *runtime.Runtime) {
...
+
+		sort.SliceStable(order, func(i, j int) bool {
+			leftRow := order[i]
+			rightRow := order[j]
+			leftKeys, _ := keys.Get(leftRow).(*values.List)
+			rightKeys, _ := keys.Get(rightRow).(*values.List)
+			keyCount := directions.Len()
+			for keyIndex := range keyCount {
+				isAscending, _ := directions.Get(keyIndex).(bool)
+				cmp := compareQuerySortValues(leftKeys.Get(keyIndex), rightKeys.Get(keyIndex), isAscending)
```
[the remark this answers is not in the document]
gimantha: Verified against the current code. No fix is needed for that finding.

array.go (line 46) already type-checks keys, directions, indices, and payloads as *values.List.
array.go (line 65) already checks keys.Len() == indices.Len().
array.go (line 70) already validates every directions entry is bool.
array.go (line 79) already validates each keys row is a nested *values.List and matches directions.Len().
array.go (line 99) already uses the correct explicit loop: for i := 0; i < rowCount; i++.
I also verified with:

go test ./corpus -run 'TestIntegration/subset8/08-query-expression/order-by-query-sort'
go test ./lib/array/runtime
Both passed.

ast/clauses.go
```diff
@@ -36,10 +36,20 @@ type (
 	BLangFromClause struct {
 		BLangInputClause
 	}
+	BLangJoinClause struct {
+		BLangInputClause
+		OnClause        *BLangOnClause
```
heshanpadmasiri: I belive OnClause should be there always. In that case I would personally prefer to hold it by value instead by pointer (avoid a second allocation, gc lookup and heap segmentation).

ast/clauses.go
```diff
@@ -36,10 +36,20 @@ type (
...
+		IsOuterJoinFlag bool
+	}
 	BLangLetClause struct {
 		bLangNodeBase
 		LetVarDeclarations []model.VariableDefinitionNode
 	}
+	BLangOnClause struct {
+		bLangNodeBase
+		LhsExpr BLangExpression
+		RhsExpr BLangExpression
```
heshanpadmasiri: NIT: I wonder if it is better to call them On and Equals instead

ast/clauses.go
```diff
@@ -48,10 +58,23 @@ type (
...
 		Expression BLangExpression
 	}
+	BLangOrderByClause struct {
+		bLangNodeBase
+		OrderByKeyList []BLangOrderKey
+	}
+	BLangOrderKey struct {
+		bLangNodeBase
+		Expression  BLangExpression
+		IsAscending bool
```
heshanpadmasiri: NIT: I would personally make this IsDescending. Default value for bool in false and default for order is ascending. So we only need to explicitly set this when we have descending

ast/clauses.go
heshanpadmasiri: STNode as well (that's machine translated from java implementation)

desugar/query_expression.go
```diff
@@ -0,0 +1,1759 @@
...
+func cloneSimpleVariableDef(varDef *ast.BLangSimpleVariableDef) *ast.BLangSimpleVariableDef {
+	clone := *varDef
+	cloneVar := *varDef.Var
+	if varDef.Var.Name != nil {
+		cloneName := *varDef.Var.Name
+		cloneVar.Name = &cloneName
+	}
+	clone.Var = &cloneVar
+	return &clone
+}
```
heshanpadmasiri: I am bit suspicious about this. As far I understand this used inside the loop to redeclare the same variable inside the loop as one declared outside with the **same symbol**. Then inside the loop we assign to that second variable and exploit the fact in BIR gen given both variable declarations have the same symbol so it will assign to the same frame slot. 
I would expect instead use a variable reference to the outer variable and assign to that reference instead. 
Generally we shouldn't (and hopefully we don't at the momement) generate AST's that are invalid after desugaring.

desugar/query_expression.go
```diff
@@ -0,0 +1,1759 @@
...
+	for i := startClauseIndex; i < endClauseIndex; i++ {
+		if _, isOrderBy := queryExpr.QueryClauseList[i].(*ast.BLangOrderByClause); isOrderBy {
+			indices = append(indices, i)
+		}
+	}
+	return indices
+}
+
+type queryLetStore struct {
+	varDef   *ast.BLangSimpleVariableDef
```
heshanpadmasiri: Fallow on from above I don't think we should pass in reference to variable definition here (and in  `queryRowBinding`) because it makes it difficult to reason if we are going to mutate the variable definition anywhere deep inside (it seems not to be the case). I expect they should be variable reference expressions (or symbols + what ever information needed to create a variable reference) in that case. 

Otherwise they should ideally return a variable definition

lib/array/compile/array.go
```diff
@@ -50,6 +50,15 @@ func GetArraySymbols(ctx *context.CompilerContext) model.ExportedSymbolSpace {
 	space.AddSymbol("length", lengthSymbol)
 	lengthRef, _ := space.GetSymbol("length")
 	ctx.SetSymbolType(lengthRef, libcommon.FunctionSignatureToSemType(ctx.GetTypeEnv(), &lenghtSignature))
+
+	querySortSignature := model.FunctionSignature{
+		ParamTypes: []semtypes.SemType{semtypes.LIST, semtypes.LIST, semtypes.LIST, semtypes.LIST},
+		ReturnType: semtypes.NIL,
+	}
+	querySortSymbol := model.NewFunctionSymbol("querySort", querySortSignature, true)
```
heshanpadmasiri: I assume this new (not defined in `lang.array`) function added to support order by clause. In that case I wonder if we need to introduce a seperate library like `lang.internal` (I assume we will more functions like this in the future for desugar). Otherwise we will expose this function to user as well

semantics/semantic_analyzer.go
```diff
@@ -675,24 +675,65 @@ func analyzeCheckPanickedExpr[A analyzer](a A, expr *ast.BLangCheckPanickedExpr,
 	return validateResolvedType(a, expr, expectedType)
 }
 
-func analyzeQueryExpr[A analyzer](a A, queryExpr *ast.BLangQueryExpr, expectedType semtypes.SemType) bool {
+func queryExprClausesForAnalysis[A analyzer](
+	a A,
+	queryExpr *ast.BLangQueryExpr,
+) (*ast.BLangFromClause, *ast.BLangSelectClause, *ast.BLangOnConflictClause, int, bool) {
```
heshanpadmasiri: NIT: If you have to return so many values may be it is better to define a struct and return it instead

semantics/semantic_analyzer.go
```diff
@@ -675,24 +675,65 @@ func analyzeCheckPanickedExpr[A analyzer](a A, expr *ast.BLangCheckPanickedExpr,
...
+	for i := 1; i < lastClauseIndex; i++ {
 		switch clause := queryExpr.QueryClauseList[i].(type) {
+		case *ast.BLangJoinClause:
+			if !analyzeActionOrExpression(a, clause.Collection, nil) {
+				return false
+			}
+			if clause.OnClause == nil {
+				a.semanticErr("join clause requires an on clause", clause.GetPosition())
+				return false
+			}
```
heshanpadmasiri: I think parser already detects this and should give a parser error. Please check

semantics/semantic_analyzer.go
heshanpadmasiri: Also isn't from always guranteed to be the first clause?

semantics/semantic_analyzer.go
```diff
@@ -710,9 +751,13 @@ func analyzeQueryExpr[A analyzer](a A, queryExpr *ast.BLangQueryExpr, expectedTy
 			}
 		case *ast.BLangWhereClause, *ast.BLangLimitClause:
 			// Query clause type and shape validation already happen in type resolution.
-		default:
-			a.unimplementedErr("only let + where clauses are supported as intermediate query clauses", clause.GetPosition())
-			return false
+		case *ast.BLangOrderByClause:
+			for j := range clause.OrderByKeyList {
+				orderKey := &clause.OrderByKeyList[j]
+				if !analyzeActionOrExpression(a, orderKey.Expression, nil) {
```
heshanpadmasiri: I assume this has expected type [Comparable](https://ballerina.io/spec/lang/master/#ordering) (otherwise this will let uncomparable field access expressions like function references as well). I don't think we have defined the the top type anywhere so probably need to do it similar to how we have done things like JSON

semantics/type_resolver.go
```diff
@@ -1629,6 +1629,9 @@ func resolveNumericLiteral(t typeResolver, n *ast.BLangNumericLiteral, expectedT
 func updateSymbolType(t typeResolver, node ast.BLangNode, ty semtypes.SemType) {
 	if nodeWithSymbol, ok := node.(ast.BNodeWithSymbol); ok {
 		symbol := nodeWithSymbol.Symbol()
+		if symbol == (model.SymbolRef{}) {
+			return
+		}
```
heshanpadmasiri: Not sure why this is needed. SymbolRef 0,0 is usually a lang lib but is not deterministic as far as I can recall. Either way I don't think it should be special cased (0,0) don't have a special meaning it usually set to same (I think io:println because we have it in the first line usually) but not guranteed to be so.

semantics/type_resolver.go
```diff
@@ -2405,14 +2508,69 @@ func resolveQueryIntermediateClauses(t typeResolver, chain *binding, queryExpr *
...
 
+var queryOrderablePrimitiveTypes = []semtypes.SemType{
+	semtypes.BOOLEAN,
+	semtypes.INT,
+	semtypes.FLOAT,
+	semtypes.DECIMAL,
+	semtypes.STRING,
+}
+
+func isQueryOrderableType(t typeResolver, ty semtypes.SemType, depth int) bool {
```
heshanpadmasiri: I think the correct way to create a recursive type defintion in semtypes module (like createJSON) and use it here and do a simple is subtype.

lib/array/runtime/array.go
```diff
@@ -41,6 +43,179 @@ func initArrayModule(rt *runtime.Runtime) {
 		}
 		return nil, fmt.Errorf("first argument must be an array")
 	})
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "querySort", func(args []values.BalValue) (values.BalValue, error) {
+		keys, ok := args[0].(*values.List)
```
heshanpadmasiri: NIT: I think it is better if we can keep the names aligned with the terms defined in the spec it self, unless we have some sort of reason (performance usually) to do otherwise

lib/array/compile/array.go
gimantha: Introduced a new internal lib

heshanpadmasiri (approved):
LGTM

PR #403, opened 2026-04-24
Add support for enum

### LINKED ISSUE #402: [New Feature]: Add support for enum
### Description

Enum members are treated as constants, and the enum type is represented as the union of those member constants

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

snelusha:
## Purpose
Resolves #402
Depends on #419

## Summary by CodeRabbit

* **New Features**
  * Adds first-class public enum declarations with members (optional explicit string values); enum values print and convert to strings. Member markdown docs are preserved.
* **Bug Fixes**
  * Reports duplicate enum members and emits errors for enums with no valid members; improves related syntax/semantic diagnostics.
* **Tests**
  * New integration fixtures and examples validating enum behavior, output, and diagnostics.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -483,10 +483,12 @@ const (
 	flagConstant      nodeFlags = 1 << model.Flag_CONSTANT
 	flagDistinct      nodeFlags = 1 << model.Flag_DISTINCT
 	flagClass         nodeFlags = 1 << model.Flag_CLASS
+	flagEnum          nodeFlags = 1 << model.Flag_ENUM
 	flagRequiredParam nodeFlags = 1 << model.Flag_REQUIRED_PARAM
 	flagDefaultParam  nodeFlags = 1 << model.Flag_DEFAULTABLE_PARAM
 	flagRestParam     nodeFlags = 1 << model.Flag_REST_PARAM
 	flagAnyFunction   nodeFlags = 1 << model.Flag_ANY_FUNCTION
+	flagEnumMember    nodeFlags = 1 << model.Flag_ENUM_MEMBER
 )
 
 func (f nodeFlags) has(flag nodeFlags) bool { return f&flag != 0 }
@@ -532,6 +534,7 @@ func (b *BLangVariableBase) IsReadonly() bool     { return b.flags.has(flagReado
 func (b *BLangVariableBase) FlagsAsInt64() int64  { return b.flags.asInt64() }
 
 func (b *BLangConstant) FlagsAsInt64() int64 { return (b.flags | flagConstant).asInt64() }
+func (b *BLangConstant) SetEnumMember()      { b.flags |= flagEnumMember }
 
 // BLangClassDefinition flag methods
 func (b *BLangClassDefinition) IsPublic() bool   { return b.flags.has(flagPublic) }
@@ -555,6 +558,7 @@ func (b *BLangTypeDefinition) IsPublic() bool    { return b.flags.has(flagPublic
 func (b *BLangTypeDefinition) IsAnonymous() bool { return b.flags.has(flagAnonymous) }
 func (b *BLangTypeDefinition) SetPublic()        { b.flags |= flagPublic }
 func (b *BLangTypeDefinition) SetAnonymous()     { b.flags |= flagAnonymous }
+func (b *BLangTypeDefinition) SetEnum()          { b.flags |= flagEnum }
 
 // Stub IsPublic for types with no flags
 func (b *BLangAnnotation) IsPublic() bool     { return false }
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -483,12 +483,10 @@ const (
 	flagConstant      nodeFlags = 1 << model.Flag_CONSTANT
 	flagDistinct      nodeFlags = 1 << model.Flag_DISTINCT
 	flagClass         nodeFlags = 1 << model.Flag_CLASS
-	flagEnum          nodeFlags = 1 << model.Flag_ENUM
 	flagRequiredParam nodeFlags = 1 << model.Flag_REQUIRED_PARAM
 	flagDefaultParam  nodeFlags = 1 << model.Flag_DEFAULTABLE_PARAM
 	flagRestParam     nodeFlags = 1 << model.Flag_REST_PARAM
 	flagAnyFunction   nodeFlags = 1 << model.Flag_ANY_FUNCTION
-	flagEnumMember    nodeFlags = 1 << model.Flag_ENUM_MEMBER
 )
 
 func (f nodeFlags) has(flag nodeFlags) bool { return f&flag != 0 }
@@ -534,7 +532,6 @@ func (b *BLangVariableBase) IsReadonly() bool     { return b.flags.has(flagReado
 func (b *BLangVariableBase) FlagsAsInt64() int64  { return b.flags.asInt64() }
 
 func (b *BLangConstant) FlagsAsInt64() int64 { return (b.flags | flagConstant).asInt64() }
-func (b *BLangConstant) SetEnumMember()      { b.flags |= flagEnumMember }
 
 // BLangClassDefinition flag methods
 func (b *BLangClassDefinition) IsPublic() bool   { return b.flags.has(flagPublic) }
@@ -558,7 +555,6 @@ func (b *BLangTypeDefinition) IsPublic() bool    { return b.flags.has(flagPublic
 func (b *BLangTypeDefinition) IsAnonymous() bool { return b.flags.has(flagAnonymous) }
 func (b *BLangTypeDefinition) SetPublic()        { b.flags |= flagPublic }
 func (b *BLangTypeDefinition) SetAnonymous()     { b.flags |= flagAnonymous }
-func (b *BLangTypeDefinition) SetEnum()          { b.flags |= flagEnum }
 
 // Stub IsPublic for types with no flags
 func (b *BLangAnnotation) IsPublic() bool     { return false }
--- a/ast/node_builder.go
+++ b/ast/node_builder.go
@@ -84,7 +84,7 @@ type NodeBuilder struct {
 	isInLocalContext     bool
 	isInFiniteContext    bool
 	inCollectContext     bool
-	constantSet          map[string]bool // Track declared constants to detect redeclarations
+	constantSet          map[string]string // Track declared constants to detect redeclarations
 	cx                   *context.CompilerContext
 	types                typeTable
 }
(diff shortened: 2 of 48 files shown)


corpus/bal/subset8/08-enum/1-v.bal
```diff
@@ -0,0 +1,30 @@
...
+
+public enum Color {
+    RED,
+    GREEN = "green",
+    BLUE
+}
+
+public function main() {
+    Color c = RED;
+    io:println(c); // @output RED
```
heshanpadmasiri, line 27: Also add tests to validate typing is correct.
```ballerina
"RED"|"green"|"BLUE" x = c
string x = c
```

heshanpadmasiri (commented):
Also add project tests to validate importing works

corpus/project/module-enum-v/Ballerina.toml
```diff
@@ -0,0 +1,4 @@
+[package]
+org = "testorg"
+name = "moduleenum"
```
warunalakshitha:
```suggestion
name = "module_enum"
```

corpus/bal/subset8/08-enum/1-v.bal
```diff
@@ -0,0 +1,33 @@
...
+    io:println(GREEN); // @output green
+    io:println(BLUE); // @output BLUE
+
+    "RED"|"green"|"BLUE" x = c;
```
heshanpadmasiri, line 31: you need -e cases for this as well

ast/ast.go
```diff
@@ -482,10 +482,12 @@ const (
 	flagConstant      nodeFlags = 1 << model.Flag_CONSTANT
 	flagDistinct      nodeFlags = 1 << model.Flag_DISTINCT
 	flagClass         nodeFlags = 1 << model.Flag_CLASS
+	flagEnum          nodeFlags = 1 << model.Flag_ENUM
```
heshanpadmasiri: Why do we need these 2 new flags?

heshanpadmasiri: I assume we are directly converting the enum to a string union definition in the node_builder it self correct?

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+	memberNodes := enumDeclarationNode.EnumMemberList()
+	memberTypeNodes := make([]model.TypeDescriptor, 0)
+	for memberNode := range memberNodes.Iterator() {
+		if memberNode.Kind() != common.ENUM_MEMBER {
+			continue
+		}
+		enumMember := memberNode.(*tree.EnumMemberNode)
+		if enumMember.Identifier() == nil || enumMember.Identifier().IsMissing() {
+			continue
+		}
```
heshanpadmasiri, line 3258: shouldn't this be an error? not sure how to write a code like this though. Think carefully and if it can't be done then should be an internal error. Else should be a syntax error

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+		constantNode, redeclared := n.transformEnumMember(enumMember, publicQualifier)
+		if redeclared {
+			continue
+		}
+		if n.currentCompUnit == nil {
+			n.cx.InternalError("enum constants can only be added at module level", getPosition(n.de(), enumMember))
+			continue
+		}
```
heshanpadmasiri, line 3266: If I try to define an enum in a functino body, will I get an internal error or syntax error? It should be syntax error but it is possible parser already give the error. Validate and only update if needed

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+
+	identifierPos := getPosition(n.de(), enumDeclarationNode.Identifier())
+	identifier := createIdentifierFromToken(identifierPos, enumDeclarationNode.Identifier())
+	typeDef.Name = &identifier
+
+	if len(memberTypeNodes) == 1 {
+		typeDef.SetTypeData(model.TypeData{TypeDescriptor: memberTypeNodes[0]})
+	} else if len(memberTypeNodes) > 1 {
+		current := memberTypeNodes[len(memberTypeNodes)-1]
+		for i := len(memberTypeNodes) - 2; i >= 0; i-- {
```
heshanpadmasiri: Why can't we just start from one and move forward. Also I think you shoud be able to merge == 1 and > 1 cases.

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+			}
+			unionType.pos = typeDef.pos
+			current = unionType
+		}
+		typeDef.SetTypeData(model.TypeData{TypeDescriptor: current})
+	} else {
+		neverType := &BLangValueType{TypeKind: model.TypeKind_NEVER}
+		neverType.pos = diagnostics.NewBuiltinLocation()
+		typeDef.SetTypeData(model.TypeData{TypeDescriptor: neverType})
+		n.cx.SemanticError("missing enum member", typeDef.Name.GetPosition())
```
heshanpadmasiri: syntax error not semantic error

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+	if exprNode := enumMemberNode.ConstExprNode(); exprNode != nil {
+		constantNode.Expr = n.createExpression(exprNode)
+	} else {
+		literal := &BLangLiteral{}
+		literal.pos = identifierPos
+		literal.Value = identifier.Value
+		literal.OriginalValue = identifier.Value
+		literal.SetValueType(n.types.getTypeFromTag(model.TypeTags_STRING).(BType))
+		constantNode.Expr = literal
+	}
```
heshanpadmasiri, line 3328: I am suspicious about this given constant declarations didn't need this. please double check

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+	qualifier := enumDeclarationNode.Qualifier()
+	if qualifier != nil && qualifier.Kind() == common.PUBLIC_KEYWORD {
+		publicQualifier = true
+	}
+
+	memberNodes := enumDeclarationNode.EnumMemberList()
+	memberTypeNodes := make([]model.TypeDescriptor, 0)
+	for memberNode := range memberNodes.Iterator() {
+		if memberNode.Kind() != common.ENUM_MEMBER {
+			continue
```
heshanpadmasiri, line 3252: Do we detect cases where we you add random nonsense like var ref, function declarations in enum body?

ast/node_builder.go
```diff
@@ -3210,11 +3213,121 @@ func (n *NodeBuilder) TransformConditionalExpression(conditionalExpressionNode *
...
+
+	metadata := enumMemberNode.Metadata()
+	if metadata != nil && !metadata.IsMissing() {
+		docString := getDocumentationString(metadata)
+		constantNode.MarkdownDocumentationAttachment = n.createMarkdownDocumentationAttachment(docString)
+	}
+
+	constantName := constantNode.Name.GetValue()
+	if n.constantSet[constantName] {
+		n.cx.SemanticError("redeclared symbol '"+constantName+"'", constantNode.Name.GetPosition())
```
heshanpadmasiri, line 3342: Do we detect cases where user have already defined a constant and there is a enum conflicting with that?

ast/node_builder.go
snelusha: Currently we get panicked from `node_builder.go:4173`

ast/node_builder.go
snelusha: Enum members without explicit values require an implicit string literal, same as the JBallerina.

corpus/bal/subset8/08-enum/1-v.bal
snelusha: Added

ast/node_builder.go
snelusha: Yes, we do detect and also I added tests to cover this

ast/node_builder.go
snelusha: This should be handled in the parser and I added an internal error here!

ast/node_builder.go
snelusha: We get syntax errors

ast/node_builder.go
snelusha: Yes in here also we get syntax errors

ast/ast.go
snelusha: Removed enum flags

ast/node_builder.go
heshanpadmasiri: I understand that my point is we already have logic for creating string literals and this is duplicating it. extract to helper and reuse it

ast/node_builder.go
```diff
@@ -1335,6 +1338,10 @@ func (n *NodeBuilder) TransformModulePart(modulePartNode *tree.ModulePart) BLang
 	// Generate import declarations
 	imports := modulePartNode.Imports()
 	for importDecl := range imports.Iterator() {
+		if importDecl.HasDiagnostics() {
```
heshanpadmasiri, line 1345: Do we need to do this everywhere? If I recall original design was when we have a syntax error in some node we have a single function that get the "narrowest" node with the error and give it (not sure what happens when there are mulitple such nodes ideally if they are no overlapping then all should get errors). Then simply calling this at the top level node should be enough.

corpus/integration/subset8/08-enum/3-e.txtar
```diff
@@ -0,0 +1,7 @@
+-- stdout --
+-- stderr --
+error[SYNTAX_ERROR]: missing enum member
+  --> 3-e.bal:19:19
+   |
+19 | public enum Color { // @error
```
heshanpadmasiri: NIT: ideally I would like this be the whole enum decl not `{`

corpus/bal/subset8/08-enum/7-e.bal
```diff
@@ -0,0 +1,25 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+// @productions module-enum-decl
+import ballerina/io;
+
+string Color = "RED";
+
+public enum Color { // @error
```
heshanpadmasiri, line 21: We also need a test where there is module level variable that collide with enum variant

snelusha: Added

heshanpadmasiri:
This now crashes 
```ballerina
public enum Color {
    RED, GREEN, BLUE
}

public const string RED = "red";
```

snelusha:
@heshanpadmasiri 

> This now crashes
> 
> ```ballerina
> public enum Color {
>     RED, GREEN, BLUE
> }
> 
> public const string RED = "red";
> ```

Fixed

heshanpadmasiri (approved):
LGTM

PR #419, opened 2026-04-30
Prevent frontend from continuing if there is an error in any stage

### LINKED ISSUE #418: [Bug]: Prevent frontend from continuing if there is an error in any stage
### Description

$subject

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Purpose
$subject
Resolves #418 

## Summary by CodeRabbit

* **Bug Fixes**
  * Compiler stops earlier on module errors to prevent cascading failures; redeclaration diagnostics now report all relevant locations.

* **Documentation**
  * Added detailed compiler pipeline and error-propagation documentation.

* **Tests**
  * Updated/added integration and syntax fixtures to reflect improved diagnostics.

* **New Features**
  * Added example projects demonstrating redeclaration error scenarios and expected outputs.


azinneera (dismissed):
LGTM

semantics/symbol_resolver.go
```diff
@@ -57,11 +57,17 @@ type (
 		nextDefaultSymbolName() string
 	}
 
+	prevPos struct {
```
warunalakshitha, line 60: Why this change?

heshanpadmasiri: We have a race condition when there a duplicate symbol declarations in different files which one get reported as the error. This change in intended to fix it by report both positions as duplicates instead of second declaration we see

semantics/symbol_resolver.go
```diff
@@ -185,10 +192,17 @@ func (bs *blockSymbolResolver) GetTypeDefns() map[model.SymbolRef]model.TypeDefi
 
 func addTopLevelSymbol(resolver *moduleSymbolResolver, name string, symbol model.Symbol, pos diagnostics.Location) bool {
 	if _, _, exists := resolver.GetSymbol(name); exists {
-		semanticError(resolver, "redeclared symbol '"+name+"'", pos)
+		msg := "redeclared symbol '" + name + "'"
+		if prev, ok := resolver.prevPos[name]; ok && !prev.reported {
+			semanticError(resolver, msg, prev.pos)
```
warunalakshitha, line 197: What if we have redeclared symbol more than 2 . will that be not logged since reported.

heshanpadmasiri: No we are using reported to avoid duplicate declarations for the initial node we found. Given resolver doesn't try to recover and continue within given "kind" (we go over each top level node first if one fails we stop there; then we go over each function if we found an error we simply skip to the next function), it is unlikely you will get more than 2 symbol resolution error unless they are in different files.  In this case you will get all those errors (added collide3-e)

This is intentional to keep the number of states symbol resolver could be in to a minimum

warunalakshitha (approved):

PR #448, opened 2026-05-14
Add `ballerina/http` client support with full HTTP methods, TLS, and response API

### LINKED ISSUE #444: Add HTTP Client basic support
### Description

Add support for basic client operations with limited client configuration

### Describe your problem(s)

Currently, the implementation does not support HTTP client calls. We need to support the basic functionalities

### Describe your solution(s)

Add supports for basic HTTP client operations and configuration. Need to decide on the scope

### Related area

Other Area

_No response_

_No response_

_No response_

TharmiganK:
## Purpose
Implements the `ballerina/http` client module for the Go-based Ballerina runtime, enabling Ballerina programs to make outbound HTTP/HTTPS requests. This covers HTTP method dispatch, client configuration (timeout, redirect, HTTP version, TLS/mTLS), response header and payload APIs, and HTTP header parsing utilities.

Fixes: https://github.com/ballerina-platform/ballerina-lang-go/issues/444

## Approach
The implementation is split across two layers:
- **Compile-time** (`lib/http/compile/http.go`): type registration, extern function wiring, and semantic analysis hooks for `ballerina/http` symbols.
- **Runtime** (`lib/http/runtime/http.go`): PAL-backed Go HTTP client that executes remote method calls (`get`, `post`, `put`, `patch`, `delete`, `head`, `options`, `execute`) via a single `Execute(method, url, body, contentType, headers)` interface using Go's `net/http` package.
>
Key design decisions:
- `HttpVersion "2.0"` enables HTTP/2 (ALPN) and h2c via Go 1.24+ `Transport.Protocols`; `"1.1"` restricts to HTTP/1.x only.
- `secureSocket` maps to Go TLS config: `cert` sets a custom CA PEM, `key`/`certFile`/`keyFile` enables mTLS, `verifyHostName=false` sets `InsecureSkipVerify`.
- CN-based TLS hostname fallback is included for legacy self-signed certificates without SANs.
- `http:Response` header API (`hasHeader`, `getHeader`, `getHeaders`, `getHeaderNames`) stores headers as lists to preserve multi-value semantics; `TRAILING` position is accepted but silently ignored.
- `getJsonPayload()` uses a `UseNumber` JSON decoder to preserve integer precision.
- All outgoing requests carry a `User-Agent: ballerina` header.
>
A support reference document (`lib/http/client-support.md`) describes the supported API subset and known limitations.

## User stories
- As a Ballerina developer, I can create an `http:Client` and call `get`, `post`, `put`, `patch`, `delete`, `head`, `options`, and `execute` methods to make outbound HTTP requests.
- As a Ballerina developer, I can configure the HTTP client with a timeout, redirect policy, HTTP version (1.1 or 2.0), and TLS/mTLS settings via `ClientConfiguration`.
- As a Ballerina developer, I can read response headers using `hasHeader`, `getHeader`, `getHeaders`, and `getHeaderNames` on an `http:Response`.
- As a Ballerina developer, I can extract response payloads as `string`, `json`, or `byte[]` using `getTextPayload`, `getJsonPayload`, and `getBinaryPayload`.
- As a Ballerina developer, I can parse HTTP header values using `http:parseHeader`.

## Release note
Added `ballerina/http` client support to the Go-based Ballerina runtime. Programs can now make outbound HTTP and HTTPS requests using the standard `http:Client` API, including all HTTP methods, `ClientConfiguration` (timeout, redirect, HTTP version, TLS/mTLS via `secureSocket`), and a full `http:Response` header and payload API.

## Documentation
- `lib/http/client-support.md` — in-repo support reference documenting the supported API subset, type definitions, known limitations, and implementation notes.

## Training
N/A

## Certification
N/A — this change adds runtime support for a standard library module. No new exam-relevant certification questions arise from an implementation change of this nature.

## Marketing
N/A

## Automation tests
- Unit tests
  Corpus tests added for all supported `ballerina/http` scenarios: `http-client-v`, `http-client-methods-v`, `http-client-post-v`, `http-client-tls-v`, `http-client-response-headers-v`, `http-client-response-payload-v`, `http-parse-header-v`. Error cases covered: `http-client-e`, `http-client-config-e`, `http-client-tls-e`, `http-parse-header-e`. AST, BIR, CFG, and desugared corpus outputs are all updated.
- Integration tests
  Extern integration tests in `extern-test/http_client_test.go` cover: `Get`, `Post`, all HTTP methods (`Methods`), and TLS with insecure verify (`TLSInsecure`). Tests run against a live local HTTP/HTTPS test server spun up within the test harness.

## Security checks

## Samples
Ballerina source samples are provided as corpus test inputs under `corpus/bal/subset8/08-network/` and `extern-test/testdata/`, demonstrating client creation, HTTP method calls, TLS configuration, and response payload/header extraction.

## Related PRs
N/A

## Migrations (if applicable)
N/A — new feature; no existing API changed.

## Test environment
- Go 1.26.0, darwin/arm64 (Apple Silicon)
- Corpus tests: `go test ./corpus/...`
- Extern tests: `go test ./extern-test/...`

## Learning
- [Ballerina HTTP client spec](https://lib.ballerina.io/ballerina/http/latest) — reference for supported type definitions and method signatures.
- Go `net/http` transport documentation for HTTP/2 (`Transport.Protocols`) and TLS configuration (`tls.Config`).
- `encoding/json` `UseNumber` decoder for integer-preserving JSON parsing.

## Summary by CodeRabbit

* **New Features**
  * Full HTTP client support: create clients, common methods (GET/POST/PUT/PATCH/DELETE/HEAD/OPTIONS/execute), header parsing, payload accessors, and TLS configuration (including insecure mode).

* **Improvements**
  * Better constructor/default-parameter handling and stricter propagation of init/main errors.
  * Platform abstraction for IO/HTTP with a native platform implementation.

* **Tests**
  * Extensive unit and integration tests and fixtures covering client methods, headers, payloads, TLS, and interpreter behavior.

* **Documentation**
  * Expanded HTTP client docs with defaults, supported behaviors, and capability matrices.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ballerina-platform/ballerina-lang-go/pull/448?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


semantics/type_resolver.go
```diff
@@ -4018,6 +4070,17 @@ func resolveBTypeInner(t typeResolver, btype ast.BType, depth int) (semtypes.Sem
...
+			// json / anydata ≈ NIL|BOOLEAN|INT|FLOAT|DECIMAL|STRING|list|map
+			// (xml excluded — not representable at runtime)
+			return semtypes.Union(semtypes.SIMPLE_OR_STRING, semtypes.Union(semtypes.LIST, semtypes.MAPPING)), true
+		case model.TypeKind_ANY:
+			return semtypes.ANY, true
+		case model.TypeKind_XML:
+			return semtypes.XML, true
+		case model.TypeKind_STREAM, model.TypeKind_TABLE, model.TypeKind_FUTURE:
+			t.unimplemented("unsupported builtin type kind: "+string(ty.TypeKind), ty.GetPosition())
+			return nil, false
```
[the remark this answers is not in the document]
TharmiganK: BLangValueType uses the memoised recursive CreateAnydata (with cycle-breaking), while BLangBuiltInRefTypeNode uses a simpler approximation intentionally. Unifying them would either degrade the recursive type's correctness or silently change the semantics of anydata variable declarations.

runtime/runtime.go
```diff
@@ -78,3 +79,24 @@ func (rt *Runtime) GetTypeEnv() semtypes.Env {
...
+// RegisterModuleGlobals makes module-level constants accessible at runtime.
+// When Ballerina source code accesses an extern package's constant (e.g. http:LEADING),
+// the BIR executor looks it up as a global variable in that package's module. Without
+// registration, GetModule returns nil and causes a nil dereference panic.
+func RegisterModuleGlobals(rt *Runtime, pkgId *model.PackageID, globals map[string]values.BalValue) {
+	if existing := rt.registry.GetModule(pkgId); existing != nil {
+		for k, v := range globals {
+			existing.Globals[k] = v
+		}
+		return
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,731 @@
...
+
+// listToBytes converts a Ballerina byte[] (List of int64) to []byte.
+func listToBytes(list *values.List) []byte {
+	b := make([]byte, list.Len())
+	for i := range list.Len() {
+		if n, ok := list.Get(i).(int64); ok {
+			b[i] = byte(n)
+		}
+	}
+	return b
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,731 @@
...
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "Response.getHeader",
+		func(args []values.BalValue) (values.BalValue, error) {
+			self := args[0].(*values.Object)
+			name := strings.ToLower(args[1].(string))
+			v, ok := responseHeaders(self).Get(name)
+			if !ok {
+				return values.NewErrorWithMessage("header not found: " + name), nil
+			}
+			return v.(*values.List).Get(0), nil
+		})
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

lib/http/client-support.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Request message body
+
+The `message` parameter for body-carrying methods accepts:
+
+- `string` — sent as `text/plain`
+- `byte[]` — sent as `application/octet-stream`
+- `json`-compatible values (`nil`, `boolean`, `int`, `float`, `decimal`, nested maps and lists) — serialised and sent as `application/json`
+
+The `mediaType` parameter overrides the inferred `Content-Type` in all cases.
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

lib/http/client-support.md
```diff
@@ -0,0 +1,282 @@
...
+public isolated function parseHeader(string headerValue) returns string[]|error = external;
+```
+
+---
+
+## Supported
+
+### Client — remote methods
+
+All nine remote methods are supported: `get`, `post`, `put`, `patch`, `delete`, `head`, `options`, and `execute`. Each method accepts optional request headers as a `map<string|string[]>`. Methods that carry a body (`post`, `put`, `patch`, `delete`, `execute`) additionally accept an optional media type override.
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

extern-test/http_client_test.go
```diff
@@ -0,0 +1,348 @@
...
+	"ballerina-lang-go/test_util"
+)
+
+type rewritingHTTPClient struct {
+	serverURL string
+	client    *http.Client
+}
+
+func (c *rewritingHTTPClient) Execute(method, url string, body []byte, contentType string, reqHeaders map[string][]string) (int, map[string][]string, []byte, error) {
+	realURL := c.serverURL + url[len("http://testserver"):]
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

desugar/expression.go
```diff
@@ -780,12 +780,60 @@ func walkNewExpression(cx *functionContext, expr *ast.BLangNewExpression) desuga
...
+		defaultInv.ArgExprs = append([]ast.BLangExpression(nil), expr.ArgsExprs[:i]...)
+		defaultInv.SetSymbol(dp.Symbol)
+		defaultInv.SetDeterminedType(sig.ParamTypes[i])
+		setPositionIfMissing(defaultInv, pos)
+
+		varDef, varRef := assignToLocal(cx, defaultInv, pos)
+		initStmts = append(initStmts, varDef)
+		expr.ArgsExprs = append(expr.ArgsExprs, varRef)
+	}
+	return initStmts
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

cli/cmd/utils.go
```diff
@@ -40,6 +79,78 @@ var nativePal = pal.Platform{
...
+					// we do our own verification so CN-only certs are accepted as a fallback.
+					tlsConfig.InsecureSkipVerify = true //nolint:gosec
+					tlsConfig.VerifyConnection = tlsVerifyConnectionWithCNFallback(pool)
+				}
+			}
+			if len(cfg.TLS.ClientCertPEM) > 0 && len(cfg.TLS.ClientKeyPEM) > 0 {
+				if cert, err := tls.X509KeyPair(cfg.TLS.ClientCertPEM, cfg.TLS.ClientKeyPEM); err == nil {
+					tlsConfig.Certificates = []tls.Certificate{cert}
+				}
+			}
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [5701625](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/570162571287e7e75b70ce2471c7692d2f165c6c)

cli/cmd/utils.go
```diff
@@ -31,6 +35,41 @@ import (
...
+	}
+	for k, vals := range reqHeaders {
+		if len(vals) == 0 {
+			continue
+		}
+		req.Header.Set(k, vals[0])
+		for _, v := range vals[1:] {
+			req.Header.Add(k, v)
+		}
+	}
```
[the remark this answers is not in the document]
TharmiganK: We are not promising this, `ballerina` is the default value but users can overwrite it by setting the `User-Agent` header

desugar/expression.go
```diff
@@ -780,12 +780,72 @@ func walkNewExpression(cx *functionContext, expr *ast.BLangNewExpression) desuga
...
```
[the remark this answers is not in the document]
TharmiganK: Addressed via [ae63bc9](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/ae63bc954b79e4a1eb6057f85cfa8301b2c42cbf)

cli/cmd/utils.go
```diff
@@ -255,3 +369,65 @@ func buildPointer(lineContent string, startCol, highlightLen int) string {
...
+		if _, err := cs.PeerCertificates[0].Verify(opts); err != nil {
+			return err
+		}
+		// cs.ServerName is the SNI hostname (no port). Try SAN-based verification first;
+		// fall back to CN matching for legacy certificates.
+		leaf := cs.PeerCertificates[0]
+		if err := leaf.VerifyHostname(cs.ServerName); err == nil {
+			return nil
+		}
+		return tlsMatchCN(leaf.Subject.CommonName, cs.ServerName)
```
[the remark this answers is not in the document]
TharmiganK: Addressed via [ae63bc9](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/ae63bc954b79e4a1eb6057f85cfa8301b2c42cbf)

cli/cmd/utils.go
```diff
@@ -40,6 +79,81 @@ var nativePal = pal.Platform{
...
+				}
+			} else {
+				maxCount := cfg.FollowRedirects.MaxCount
+				if maxCount <= 0 {
+					maxCount = 5 // Ballerina default
+				}
+				allowAuth := cfg.FollowRedirects.AllowAuthHeaders
+				c.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+					if len(via) >= maxCount {
+						return http.ErrUseLastResponse
```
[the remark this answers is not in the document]
TharmiganK: Addressed via [ae63bc9](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/ae63bc954b79e4a1eb6057f85cfa8301b2c42cbf)

cli/cmd/utils.go
```diff
@@ -255,3 +371,69 @@ func buildPointer(lineContent string, startCol, highlightLen int) string {
...
+		// cs.ServerName is the SNI hostname (no port). Try SAN-based verification first.
+		// Only fall back to CN matching for certs that genuinely have no SANs — when SANs
+		// are present but don't match, that is a real mismatch and must not be bypassed.
+		leaf := cs.PeerCertificates[0]
+		if err := leaf.VerifyHostname(cs.ServerName); err != nil {
+			if len(leaf.DNSNames) > 0 || len(leaf.IPAddresses) > 0 {
+				return err
+			}
+			return tlsMatchCN(leaf.Subject.CommonName, cs.ServerName)
+		}
```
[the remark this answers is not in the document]
TharmiganK: The consolidation into a single Verify(opts) call with opts.DNSName set would create a security gap — a cert with no SANs that fails chain validation (expired, untrusted CA) would reach the CN-fallback branch instead of being rejected. The current two-step approach (chain verification first, hostname separately) is correct and intentional

cli/cmd/utils.go
```diff
@@ -40,6 +81,81 @@ var nativePal = pal.Platform{
 			return os.Stderr.Write(p)
 		},
 	},
+	HTTP: pal.HTTP{
+		NewClient: func(cfg pal.ClientConfig) pal.HTTPClient {
+			tlsConfig := &tls.Config{InsecureSkipVerify: cfg.TLS.InsecureSkipVerify} //nolint:gosec
+			if len(cfg.TLS.CACertPEM) > 0 {
+				pool := x509.NewCertPool()
+				pool.AppendCertsFromPEM(cfg.TLS.CACertPEM)
+				tlsConfig.RootCAs = pool
```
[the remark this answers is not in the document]
TharmiganK: Fixed with [40f19b2](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/40f19b230e743b00425eea1e1343c9d25fb83a7c)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,747 @@
...
+	if len(args) > 2 && args[2] != nil {
+		body, contentType = messageToBody(args[2])
+		if body == nil && contentType == "json_error" {
+			return values.NewErrorWithMessage("failed to serialize body to JSON"), nil
+		}
+	}
+
+	var reqHeaders map[string][]string
+	if len(args) > 3 {
+		reqHeaders = extractHeaders(args[3])
```
heshanpadmasiri: If there is a request header `Content-Type` don't we need to override the contentType here?

warunalakshitha:
Can u do basic perf test compared with jballerina simple http client and update the results here.

semantics/type_resolver.go
```diff
@@ -4490,6 +4547,17 @@ func resolveBTypeInner(t typeResolver, btype ast.BType, depth int) (semtypes.Sem
 		switch ty.TypeKind {
 		case model.TypeKind_MAP:
 			return semtypes.MAPPING, true
+		case model.TypeKind_JSON, model.TypeKind_ANYDATA:
+			// json / anydata ≈ NIL|BOOLEAN|INT|FLOAT|DECIMAL|STRING|list|map
+			// (xml excluded — not representable at runtime)
+			return semtypes.Union(semtypes.SIMPLE_OR_STRING, semtypes.Union(semtypes.LIST, semtypes.MAPPING)), true
```
heshanpadmasiri: should use `semtypes.CreateAnydata(t.typeContext())` and `semtypes.CreateJSON(t.typeContext())` (json needs to be made public)

semantics/type_resolver.go
```diff
@@ -2335,6 +2342,56 @@ func resolveNewExpr(t typeResolver, chain *binding, e *ast.BLangNewExpression, e
 	return e.GetDeterminedType(), defaultExpressionEffect(chain), true
 }
 
+// padMissingDefaultArgs adds typed placeholder expressions to e.ArgsExprs for each missing
+// param that is declared defaultable on the class's init method. It returns the count added
+// so the caller can remove them after determineObjectType runs.
+func padMissingDefaultArgs(t typeResolver, e *ast.BLangNewExpression, objectTy semtypes.SemType) int {
```
heshanpadmasiri: Instead of this I think you should use the `padArgTypesForDefaults` Something like 
```go
func padNewExprArgTypesForDefaults(t typeResolver, objectTy semtypes.SemType, argTys []semtypes.SemType, loc diagnostics.Location) []semtypes.SemType, bool {
	oat := semtypes.ToObjectAtomicType(t.typeContext(), objectTy)
	if oat == nil {
		return nil, false
	}
	classRef, ok := t.getClassAtomSymbol(oat)
	if !ok {
		return nil, false
	}
	classSym, ok := t.getSymbol(classRef).(*model.ClassSymbol)
	if !ok {
		return nil, false
	}
	initRef, ok := classSym.MethodSymbol("init")
	if !ok {
		return nil, false
	}
	return padArgTypesForDefaults(t, initRef, argTys, loc)
}
```

TharmiganK:
> Can u do basic perf test compared with jballerina simple http client and update the results here.

Do you mean comparing the execution time right?

platform/pal/platform.go
```diff
@@ -23,12 +23,49 @@
...
 	Platform struct {
-		IO IO
+		IO   IO
+		HTTP HTTP
 	}
 	IO struct {
 		Stdout func(p []byte) (n int, err error)
 		Stderr func(p []byte) (n int, err error)
 	}
+	// TLSConfig carries TLS settings derived from Ballerina's secureSocket config.
```
heshanpadmasiri, line 44: NIT: I would personally keep the http config related stuff in a seperate block and only put the HTTP struct here
```go
type (
	Platform struct {
		IO   IO
		HTTP HTTP
	}
	IO struct {
		Stdout func(p []byte) (n int, err error)
		Stderr func(p []byte) (n int, err error)
	}
	HTTP struct {
		NewClient func(cfg ClientConfig) HTTPClient
	}
)

// HTTP
type (
	// TLSConfig carries TLS settings derived from Ballerina's secureSocket config.
	TLSConfig struct {
		InsecureSkipVerify    bool          // secureSocket.enable=false OR verifyHostName=false
		CACertPEM             []byte        // secureSocket.cert (string PEM file path) → file contents
		ClientCertPEM         []byte        // secureSocket.key.certFile → file contents
		ClientKeyPEM          []byte        // secureSocket.key.keyFile  → file contents
		ServerName            string        // secureSocket.serverName → tls.Config.ServerName (SNI)
		CipherSuiteNames      []string      // secureSocket.ciphers → IANA names; platform resolves IDs
		MinVersion            uint16        // secureSocket.protocol.versions min → tls.Config.MinVersion
		MaxVersion            uint16        // secureSocket.protocol.versions max → tls.Config.MaxVersion
		HandshakeTimeout      time.Duration // secureSocket.handshakeTimeout → transport.TLSHandshakeTimeout
		DisableSessionTickets bool          // secureSocket.shareSession=false → tls.Config.SessionTicketsDisabled
	}
	// FollowRedirects controls HTTP redirect behaviour, matching Ballerina's http:FollowRedirects.
	FollowRedirects struct {
		Enabled          bool // default false — no redirects by default (Ballerina spec)
		MaxCount         int  // 0 uses Ballerina default of 5; only used when Enabled=true
		AllowAuthHeaders bool // if true, forward Authorization/Proxy-Authorization on redirect
	}
	// ClientConfig bundles all static options for a new HTTP client instance.
	ClientConfig struct {
		Timeout         time.Duration
		FollowRedirects FollowRedirects
		HTTPVersion     string // "1.1" or "2.0"; defaults to "2.0"
		TLS             TLSConfig
	}

	// HTTPClient is an opaque handle to an HTTP client created by the platform.
	// It is created once per Ballerina http:Client init and reused across requests.
	HTTPClient interface {
		Execute(method, url string, body []byte, contentType string, reqHeaders map[string][]string) (statusCode int, respHeaders map[string][]string, respBody []byte, err error)
	}
)
```

lib/http/runtime/http.go
```diff
@@ -0,0 +1,747 @@
...
+	// Remote method name uses the "$remote$" prefix (model.RemoteMethodName).
+	// The BIR gen emits `callInfo.Name = "$remote$get"` for c->get(...), which
+	// resolveObjectMethod then looks up in the object's methodKeys map.
+	clientClassDef := &bir.BIRClassDef{
+		Name:      model.Name("Client"),
+		LookupKey: "ballerina/http:Client",
+		Fields: []bir.ObjectField{
+			{Name: "url", Ty: semtypes.STRING},
+			{Name: "timeout", Ty: semtypes.DECIMAL},
+			{Name: "followRedirects", Ty: semtypes.MAPPING},
```
heshanpadmasiri: shouldn't this be `			{Name: "followRedirects", Ty: semtypes.Union(semtypes.MAPPING,semtypes.Nil)}`,?
Given this is a specific record type you may need to build the type using the type env from rt otherwise this will treated as `map` top type by the runtime

lib/http/runtime/http.go
```diff
@@ -0,0 +1,747 @@
...
+			}
+			return result, nil
+		})
+
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "Client.init",
+		func(args []values.BalValue) (values.BalValue, error) {
+			self := args[0].(*values.Object)
+			url := args[1].(string)
+
+			timeout := big.NewRat(30, 1)
```
heshanpadmasiri: You will need to use the Decimal wrapper (`decimal.go`) if you want this to be treated as decimal value by the runtime

lib/http/runtime/http.go
TharmiganK: We need to. Will fix it. Btw the media-type given in the function call gets the highest priority(default content-type < content-type from headers < media-type supplied in the function)

semantics/type_resolver.go
TharmiganK: Ack, will fix this

semantics/type_resolver.go
TharmiganK: Ack

platform/pal/platform.go
TharmiganK: Yes, makes sense. Will fix this

lib/http/runtime/http.go
TharmiganK: Yes, it should be union with nil. Will fix this

lib/http/runtime/http.go
TharmiganK: Ack, will fix this

TharmiganK: Fixed in [3c83cf6](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/3c83cf6345d6fbeb4b60286da1e08ebd9163b90f)

lib/http/runtime/http.go
TharmiganK: Fixed in [3c83cf6](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/3c83cf6345d6fbeb4b60286da1e08ebd9163b90f)

platform/pal/platform.go
TharmiganK: Fixed in [3c83cf6](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/3c83cf6345d6fbeb4b60286da1e08ebd9163b90f)

semantics/type_resolver.go
TharmiganK: Fixed in [3c83cf6](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/3c83cf6345d6fbeb4b60286da1e08ebd9163b90f)

semantics/type_resolver.go
TharmiganK: Fixed in [3c83cf6](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/3c83cf6345d6fbeb4b60286da1e08ebd9163b90f)

lib/http/runtime/http.go
TharmiganK: Fixed in [3c83cf6](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/3c83cf6345d6fbeb4b60286da1e08ebd9163b90f)

TharmiganK:
# Execution Benchmark Results

Tested on **2026-05-14** across **5 runs** on macOS (arm64).

Scenario: Calling a public endpoint in internet with GET requests

---

## Summary (5-run averages)

| Scenario | Avg Time | Avg Peak RSS |
|:---|---:|---:|
| `$BAL_GO run http-public-ep.bal` | 1.516 s | 26.3 MB |
| **`bal build` → `java -jar` (combined)** | **7.112 s** | **401.0 MB** |
| &nbsp;&nbsp;└─ `bal build http-public-ep.bal` | 4.886 s | 401.0 MB |
| &nbsp;&nbsp;└─ `java -jar http-public-ep.jar` | 2.226 s | 210.0 MB |
| `java -jar http-public-ep.jar` (pre-built) | 2.232 s | 209.4 MB |
| **`bal build --graalvm` → `./http-public-ep` (combined)** | **1m 11.6s** | **6821.0 MB** |
| &nbsp;&nbsp;└─ `bal build --graalvm http-public-ep.bal` | 1m 8.4s | 6821.0 MB |
| &nbsp;&nbsp;└─ `./http-public-ep` (after build) | 3.228 s | 132.2 MB |
| `./http-public-ep` (pre-built) | 1.252 s | 132.0 MB |

---

## Notes

- **Times include HTTP roundtrip** to `httpbin.org` — network latency varies between runs and contributes to execution time variance.
- **Peak RSS** = maximum resident set size for that process only (measured via `/usr/bin/time -l`).
- **Combined Peak RSS** = `max(build RSS, run RSS)` — the compiler dominates in both combined scenarios.
- The GraalVM native binary (`./http-public-ep`) produced at **~6.8 GB peak build memory** but runs at only **~132 MB**, making it the most memory-efficient runtime after the one-time build cost.
- `$BAL_GO run` is the **fastest runtime** at ~1.5 s with the **lowest memory** at ~26 MB — significantly lighter than the JVM path (~210 MB) and the GraalVM build step.
- `java -jar` (pre-built) and `java -jar` (after build) are statistically identical at ~2.2 s / ~210 MB, confirming build does not affect JVM startup or runtime behaviour.

cli/cmd/utils.go
```diff
@@ -31,6 +35,44 @@ import (
 	"golang.org/x/term"
 )
 
+type nativeHTTPClient struct {
```
warunalakshitha: why we have http code in cmd?

cli/cmd/utils.go
```diff
@@ -40,6 +82,84 @@ var nativePal = pal.Platform{
...
+					return http.ErrUseLastResponse
+				}
+			} else {
+				maxCount := cfg.FollowRedirects.MaxCount
+				if maxCount <= 0 {
+					maxCount = 5 // Ballerina default
+				}
+				allowAuth := cfg.FollowRedirects.AllowAuthHeaders
+				c.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+					if len(via) > maxCount {
```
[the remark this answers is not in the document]
TharmiganK: Go invokes CheckRedirect before making the next request, with via containing every request already sent. When we are about to follow redirect N, len(via) == N. With len(via) > maxCount and maxCount == 5:
- 5th redirect: len(via) == 5, 5 > 5 is false → allowed
- 6th redirect: len(via) == 6, 6 > 5 is true → denied

So we follow at most 5 redirects, which matches the Ballerina spec for maxCount = 5. Switching to >= would deny the 5th redirect and only allow 4 — that would be the off-by-one.

cli/cmd/utils.go
TharmiganK: The HTTP code lives in cmd because this file is the native-CLI's implementation of the PAL (pal.Platform). The HTTP library itself(lib/http/runtime/http.go) doesn't import net/http — it calls rt.Platform().HTTP.NewClient(...), same way lib/io calls rt.Platform().IO.Stdout(...).

I am not sure about this pattern though. @snelusha @heshanpadmasiri can you comment on this?

lib/http/runtime/http.go
```diff
@@ -0,0 +1,753 @@
...
+	case nil:
+		return nil
+	case bool:
+		return t
+	case int64:
+		return t
+	case float64:
+		return t
+	case *decimal.Decimal:
+		return t.Float64()
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

lib/http/client-support.md
```diff
@@ -0,0 +1,282 @@
...
+                                     string path,
+                                     json message,
+                                     map<string|string[]>? headers = (),
+                                     string? mediaType = ()) returns Response|error = external;
+}
+
+// ── Module-level functions ────────────────────────────────────────────────────
+
+public isolated function parseHeader(string headerValue) returns string[]|error = external;
+```
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

lib/http/compile/http.go
```diff
@@ -0,0 +1,612 @@
...
+	env := ctx.GetTypeEnv()
+
+	// headers: map<string|string[]>? — open mapping (any key, string|list values), optional.
+	// Build an explicit open mapping type so the field value type resolves to STRING|LIST
+	// rather than NEVER (which happens when the basic MAPPING atom is used directly).
+	headersMd := semtypes.NewMappingDefinition()
+	headersMapType := headersMd.DefineMappingTypeWrapped(env,
+		[]semtypes.Field{},
+		semtypes.Union(semtypes.STRING, semtypes.LIST))
+	headersOptType := semtypes.Union(headersMapType, semtypes.NIL)
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

cli/cmd/utils.go
```diff
@@ -40,6 +82,84 @@ var nativePal = pal.Platform{
...
+			}
+			if len(cfg.TLS.CipherSuiteNames) > 0 {
+				tlsConfig.CipherSuites = resolveCipherSuites(cfg.TLS.CipherSuiteNames)
+			}
+			transport := &http.Transport{
+				TLSClientConfig:     tlsConfig,
+				TLSHandshakeTimeout: cfg.TLS.HandshakeTimeout,
+			}
+			protocols := new(http.Protocols)
+			if cfg.HTTPVersion == "2.0" {
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,753 @@
...
+
+			var reqHeaders map[string][]string
+			if len(args) > 4 {
+				reqHeaders = extractHeaders(args[4])
+			}
+			if len(args) > 5 {
+				if mt, ok := args[5].(string); ok && mt != "" {
+					contentType = mt
+				}
+			}
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,753 @@
...
+// messageToBody converts a Ballerina message value to a []byte body and a default Content-Type.
+// Returns (nil, "json_error") on JSON serialization failure.
+func messageToBody(msg values.BalValue) ([]byte, string) {
+	switch v := msg.(type) {
+	case string:
+		return []byte(v), "text/plain"
+	case *values.List:
+		if b, ok := listToBytes(v); ok {
+			return b, "application/octet-stream"
+		}
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,753 @@
...
+)
+
+const (
+	orgName    = "ballerina"
+	moduleName = "http"
+)
+
+func parseHeader(input string) (*values.List, error) {
+	segments := strings.Split(input, ",")
+	list := values.NewList(0, semtypes.LIST, nil)
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1fa3a46](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1fa3a466be232f8e26799d80e65b6f2a462c2fc5)

cli/cmd/utils.go
snelusha: @warunalakshitha Currently `nativePal` and diagnostic printer are also in `cli/cmd/utils.go`. The basic idea was to keep native implementations within `cli/cmd` @heshanpadmasiri right?

heshanpadmasiri: Each environment (native/CLI, web-editor, etc) needs to provide an implementation of PAL. That way environment can control how we interact with it and runtime could be agnostic towards the environment. For example in the web-editor it will needs to use the browserAPIs to perform HTTP requests and use some UI to trigger services, file system etc and redirect standard in/out to some UI.  In the native implementation that will probably use the OS for these. 

I don't have strong opinion on where the CLI's (perhaps native runtime is better term) PAL implementation should live in other that not in the runtime package.

warunalakshitha: We can move PAL implementation (io + http) (palnative or host or platform). not sure the best name. may be palnative is best one

warunalakshitha: Basically palnative > cli > runtime

context/context.go
```diff
@@ -62,6 +62,7 @@ type CompilerContext struct {
 	diagnostics []diagnostics.Diagnostic
 	moduleStats *ModuleStats
 	stage       activeStage
+	tyCtx       semtypes.Context
```
heshanpadmasiri: You should not share type context within Compiler context. Type Context is expected to be not shared between threads compiler context will be shared by various analysis phases.

heshanpadmasiri: For anlaysis each analysis context should build it's own type context (or reuse it as far as the analysis contexts are not shared). See how we create functionTypeResolver in ResolveLocalNodes for example

heshanpadmasiri: For compiler logic it is safe to simply create it inline

heshanpadmasiri (changes requested):
Apart from type context rest LGTM

context/context.go
TharmiganK: Ack, will look into this

TharmiganK: Fixed via [e4e18e3](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/e4e18e3559f63a93a9bc89bb6aaf8ddce7af89fe)

cli/cmd/utils.go
TharmiganK: I have moved them to `palnative/cli/runtime`

TharmiganK: As discussed offline, moved `pal` and `palnative` inside the `platform` directory via [035d4ea](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/035d4ea8b519eb5f25fc6b08a8badb99c9dceb2d)

test_util/test_util.go
```diff
@@ -192,11 +192,22 @@ func computeExpectedPath(inputPath, inputBaseDir, outputBaseDir, outputExt strin
 	return filepath.Join(outputBaseDir, relPath)
 }
 
+type stubHTTPClient struct{}
+
+func (c *stubHTTPClient) Execute(_, _ string, _ []byte, _ string, _ map[string][]string) (int, map[string][]string, []byte, error) {
+	return 200, map[string][]string{}, []byte("test body"), nil
+}
+
 func TestPal(stdout io.Writer, stderr io.Writer) pal.Platform {
```
warunalakshitha, line 201: Can we directly use planative instead of having another pal impl for tests

corpus/cli_integration_test.go
```diff
@@ -343,24 +343,56 @@ func runCLICommand(t *testing.T, balBin, repoRoot, coverDir string, args ...stri
...
+		stderrStr,
 	)
 	return "", "", 0
 }
 
+// stripGoCoverageErrors removes lines emitted by Go's coverage runtime when GOCOVERDIR is set.
+// When a covered binary exits abnormally (e.g. via os.Exit after a Ballerina panic), the coverage
+// meta-data write can fail and print "error: coverage ..." to stderr. These messages are
+// infrastructure noise unrelated to the program's observable output.
+func stripGoCoverageErrors(s string) string {
```
warunalakshitha: why we need this?

lib/http/runtime/http.go
```diff
@@ -0,0 +1,792 @@
...
+		return nil
+	}
+}
+
+// toJSONBytes serializes a Ballerina value to JSON bytes.
+func toJSONBytes(v values.BalValue) ([]byte, error) {
+	return json.Marshal(balToGoJSON(v))
+}
+
+func init() {
```
warunalakshitha: shall we rearrange function definition has init function defined top of the file and other function based on call order.

warunalakshitha:
LGTM except minor improvements requested.

lib/http/runtime/http.go
TharmiganK: Ack, fixed with https://github.com/ballerina-platform/ballerina-lang-go/pull/448/

corpus/cli_integration_test.go
TharmiganK: Added to fix a CI test failure, let me try by removing it now

test_util/test_util.go
TharmiganK: The TestPal helper can't be replaced by palnative.NewPlatform() directly — palnative hardcodes os.Stdout/os.Stderr (tests need injectable writers for output capture) and its NewHTTPClient makes real HTTP requests (tests need the stub to avoid network calls). This is intentional

lib/http/runtime/http.go
```diff
@@ -0,0 +1,792 @@
...
+		func(args []values.BalValue) (values.BalValue, error) {
+			self := args[0].(*values.Object)
+			bodyVal, _ := self.Get("body")
+			body := bodyVal.(string)
+			raw := []byte(body)
+			list := values.NewList(len(raw), semtypes.LIST, nil)
+			for i, b := range raw {
+				list.FillingSet(i, int64(b))
+			}
+			return list, nil
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [1aef577](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/1aef57758cd342f2a9625670059e55cab6b3679b)

lib/http/runtime/http.go
```diff
@@ -0,0 +1,792 @@
...
+											tlsCfg.ClientCertPEM = data
+										}
+									}
+									if kv, ok := keyMap.Get("keyFile"); ok {
+										if p, ok := kv.(string); ok && p != "" {
+											data, err := os.ReadFile(p)
+											if err != nil {
+												return values.NewErrorWithMessage("secureSocket.key.keyFile: " + err.Error()), nil
+											}
+											tlsCfg.ClientKeyPEM = data
```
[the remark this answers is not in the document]
TharmiganK: This will be addressed separately. @snelusha is working on a fix

corpus/cli_integration_test.go
TharmiganK: It is working now, without this change

runtime/runtime.go
```diff
@@ -78,3 +79,27 @@ func (rt *Runtime) GetTypeEnv() semtypes.Env {
...
+// When Ballerina source code accesses an extern package's constant (e.g. http:LEADING),
+// the BIR executor looks it up as a global variable in that package's module. Without
+// registration, GetModule returns nil and causes a nil dereference panic.
+func RegisterModuleGlobals(rt *Runtime, pkgId *model.PackageID, globals map[string]values.BalValue) {
+	if existing := rt.registry.GetModule(pkgId); existing != nil {
+		if existing.Globals == nil {
+			existing.Globals = make(map[string]values.BalValue)
+		}
+		for k, v := range globals {
+			existing.Globals[k] = v
```
warunalakshitha, line 100: codecov says not cover.

runtime/internal/exec/interpreter.go
```diff
@@ -31,15 +32,24 @@ func Interpret(pkg bir.BIRPackage, reg *modules.Registry) (err error) {
 				err = getFormattedError(ctx, r)
 			}
 		}()
-		executeFunction(ctx, *pkg.InitFunction, nil, nil)
+		if result := executeFunction(ctx, *pkg.InitFunction, nil, nil); result != nil {
+			if _, ok := result.(*values.Error); ok {
+				err = getFormattedError(ctx, result)
+				return
```
warunalakshitha, line 38: new getFormattedError(ctx, result) lines not cover https://app.codecov.io/gh/ballerina-platform/ballerina-lang-go/pull/448?src=pr&el=tree&filepath=runtime%2Finternal%2Fexec%2Finterpreter.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ballerina-platform#diff-cnVudGltZS9pbnRlcm5hbC9leGVjL2ludGVycHJldGVyLmdv

desugar/expression.go
```diff
@@ -805,12 +805,80 @@ func walkNewExpression(cx *functionContext, expr *ast.BLangNewExpression) desuga
...
+		dp, ok := defaultableParams.Get(i)
+		if !ok {
+			break
+		}
+		if dp.Kind == model.DefaultableParamKindInferredTypedesc {
+			tdExpr := synthesizeInferredTypedescArg(cx, sig.ParamTypes[i], pos)
+			setPositionIfMissing(tdExpr, pos)
+			varDef, varRef := assignToLocal(cx, tdExpr, pos)
+			initStmts = append(initStmts, varDef)
+			expr.ArgsExprs = append(expr.ArgsExprs, varRef)
```
warunalakshitha, line 865: not covered by tests

runtime/runtime.go
TharmiganK: Ack, will look into this

runtime/internal/exec/interpreter.go
TharmiganK: Ack, will look into this

desugar/expression.go
TharmiganK: Ack, will look into this

TharmiganK: Covered via [f0b29fb](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/f0b29fbd68656110b11a27412ccf3641bddda0e4)

runtime/internal/exec/interpreter.go
TharmiganK: Covered via [f0b29fb](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/f0b29fbd68656110b11a27412ccf3641bddda0e4)

runtime/runtime.go
TharmiganK: Covered via [f0b29fb](https://github.com/ballerina-platform/ballerina-lang-go/pull/448/commits/f0b29fbd68656110b11a27412ccf3641bddda0e4)

heshanpadmasiri (dismissed):
Front end and semtype changes LGTM

runtime/runtime_test.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
```
warunalakshitha: We normally add all tests through corpus/bal since we need make sure all code actually run in runtime since we handle some cases in compile time which will not applicable at runtime. Can we add these cases using bal files in corpus. same with other newly added tests as well

TharmiganK: Ack, will change

TharmiganK: Did some refactoring and removed the previous tests and added them as corpus tests or external tests

warunalakshitha (approved):
LGTM

PR #456, opened 2026-05-14
Implement group by and collect clauses for maps and lists

### LINKED ISSUE #455: [Task]: Implement group by and collect for lists and maps
### Description

To support aggregations and grouping we need ti implement group by clause and collect clause for query expressions

### Describe your task(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

gimantha:
## Purpose
Fixes #455 

## Automation tests
- Unit tests
  > Code coverage information
- Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Query expressions now support GROUP BY and COLLECT, including grouping-key declarations and collect-based results.

* **Behavior/Compatibility**
  * Query finals may be SELECT or COLLECT; grouped queries aggregate non-key variables. List/tuple spread handling improved and integrated with query aggregation.

* **Tests**
  * Numerous new and updated example and integration tests covering group-by, collect, nested queries, spreads, and expected outputs/errors.

* **Diagnostics**
  * Improved error messages for intermediate and final query clause issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ballerina-platform/ballerina-lang-go/pull/456?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +117,167 @@ func initInternalModule(rt *runtime.Runtime) {
...
+			return nil, fmt.Errorf("query row %d must be a list", rowIndex)
+		}
+		if sourceRow.Len() != slotCount {
+			return nil, fmt.Errorf("query row %d length mismatch: got %d, expected %d", rowIndex, sourceRow.Len(), slotCount)
+		}
+		keyRow, ok := keyRows.Get(rowIndex).(*values.List)
+		if !ok {
+			return nil, fmt.Errorf("group key row %d must be a list", rowIndex)
+		}
+		groupIndex := findQueryGroup(groups, keyRow)
```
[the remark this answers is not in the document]
gimantha: Fixed with https://github.com/ballerina-platform/ballerina-lang-go/pull/456/commits/c52f7b2f6fc221ee72e46808281723fa981200ea

ast/clauses.go
```diff
@@ -230,6 +244,53 @@ func (b *BLangWhereClause) GetKind() model.NodeKind {
...
+	return result
+}
+
+func (b *BLangGroupingKey) GetKind() model.NodeKind {
+	return model.NodeKind_GROUPING_KEY
+}
+
+func (b *BLangGroupingKey) SetGroupingKey(groupingKey model.Node) {
+	switch key := groupingKey.(type) {
+	case *BLangSimpleVariableDef:
```
warunalakshitha: can we replace with if since default case cannot be happen

ast/clauses.go
```diff
@@ -230,6 +244,53 @@ func (b *BLangWhereClause) GetKind() model.NodeKind {
...
+}
+
+func (b *BLangGroupingKey) GetGroupingKey() model.Node {
+	if b.VariableRef != nil {
+		return b.VariableRef
+	}
+	if b.VariableDef != nil {
+		return b.VariableDef
+	}
+	return nil
```
warunalakshitha, line 275:
```suggestion
	if b.VariableRef != nil {
		return b.VariableRef
	}
	return b.VariableDef
```

lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +123,276 @@ func initInternalModule(rt *runtime.Runtime) {
 		}
 		return nil, nil
 	})
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "queryGroup", func(args []values.BalValue) (values.BalValue, error) {
+		if len(args) != 3 {
+			return nil, fmt.Errorf("queryGroup expects 3 arguments, got %d", len(args))
```
warunalakshitha: do we need if len(args) != 3 . since we desuger internally it will be always 3 args

lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +123,276 @@ func initInternalModule(rt *runtime.Runtime) {
...
+		}
+		rows, ok := args[0].(*values.List)
+		if !ok {
+			return nil, fmt.Errorf("first argument must be a list")
```
warunalakshitha: do we need this check. is it always true?

lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +123,276 @@ func initInternalModule(rt *runtime.Runtime) {
...
+		}
+		keyRows, ok := args[1].(*values.List)
+		if !ok {
```
warunalakshitha: same

lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +123,276 @@ func initInternalModule(rt *runtime.Runtime) {
...
+			return nil, fmt.Errorf("second argument must be a list")
+		}
+		scalarFlags, ok := args[2].(*values.List)
+		if !ok {
+			return nil, fmt.Errorf("third argument must be a list")
```
warunalakshitha: IMO we can scalarFlags := args[2].(*values.List) since we know them at compile time

lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +123,276 @@ func initInternalModule(rt *runtime.Runtime) {
...
+		}
+		return queryGroup(rows, keyRows, scalarFlags)
+	})
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "queryCollect", func(args []values.BalValue) (values.BalValue, error) {
+		if len(args) != 3 {
```
warunalakshitha: same as queryGroup. we can omit validations

lib/langinternal/runtime/internal.go
```diff
@@ -116,6 +123,276 @@ func initInternalModule(rt *runtime.Runtime) {
...
+	rowCount := rows.Len()
+	if keyRows.Len() != rowCount {
+		return nil, fmt.Errorf("group keys and rows length mismatch: keys=%d rows=%d", keyRows.Len(), rowCount)
+	}
+	slotCount := scalarFlags.Len()
+	scalarSlots := make([]bool, slotCount)
+	for slot := 0; slot < slotCount; slot++ {
+		isScalar, ok := scalarFlags.Get(slot).(bool)
+		if !ok {
+			return nil, fmt.Errorf("group scalar flag %d must be a bool", slot)
```
warunalakshitha: can these errors ever happen at runtime since we do compile time check. same with other errors too. But if they are for defensive code which can happen at runtime we can keep as it is.

desugar/query_expression.go
```diff
@@ -1753,6 +1990,58 @@ func createQuerySortInvocation(
...
+
+func createLangInternalInvocation(
+	cx *functionContext,
+	name string,
+	returnTy semtypes.SemType,
+	args []ast.BLangExpression,
+	pos diagnostics.Location,
+) *ast.BLangInvocation {
+	pkgName := langinternal.PackageName
+	space, ok := cx.getImportedSymbolSpace(pkgName)
```
warunalakshitha: can internal package not found happen since it will be pack always

ast/expressions.go
```diff
@@ -320,9 +320,10 @@ type (
 
 	BLangListConstructorExpr struct {
 		bLangExpressionBase
-		Exprs          []BLangExpression
-		IsTypedescExpr bool
-		AtomicType     semtypes.ListAtomicType
+		Exprs                 []BLangExpression
+		IsTypedescExpr        bool
+		AtomicType            semtypes.ListAtomicType
+		ImplicitSpreadMembers []bool
```
heshanpadmasiri: NIT: I can't think of a good reason why we need to differentiate between implicit and explict spread members here. I think just `IsSpread` should be enough

semantics/type_resolver.go
```diff
@@ -3201,30 +3500,45 @@ func resolveConstRef(t typeResolver, chain *binding, expr *ast.BLangConstRef) (s
 }
 
 func resolveListConstructorExpr(t typeResolver, chain *binding, expr *ast.BLangListConstructorExpr, expectedType semtypes.SemType) (semtypes.SemType, expressionEffect, bool) {
+	expr.ImplicitSpreadMembers = nil
```
heshanpadmasiri: Why do we need to clear this field here?

semantics/type_resolver.go
```diff
@@ -3201,30 +3500,45 @@ func resolveConstRef(t typeResolver, chain *binding, expr *ast.BLangConstRef) (s
...
 	for i, memberExpr := range expr.Exprs {
+		isImplicitSpread := isQueryAggregatedVariableReference(chain, memberExpr)
 		memberTy, _, ok := resolveActionOrExpression(t, chain, memberExpr, nil)
 		if !ok {
 			return nil, expressionEffect{}, false
 		}
+		if isImplicitSpread {
+			expr.SetImplicitSpreadMember(i)
+			spreadMemberTy := semtypes.ListProj(t.typeContext(), memberTy, semtypes.INT)
+			restTy = semtypes.Union(restTy, widenedListMemberType(spreadMemberTy))
```
heshanpadmasiri, line 3577: I don't think this is correct according to the spec "If there is no contextually expected type, then the inherent type will be a tuple-type-descriptor derived from the static type of each expression in the list-member-list. Each single-list-member that does not have a preceding spread-list-member contributes a member-type-descriptor to the tuple-type-descriptor; the type of the member-type-descriptor will be the broad type of the expression in the single-list-member. If there is a spread-list-member, then a tuple-rest-descriptor T... is added where T is the smallest type such that the broad type of the expression in every other single-list-member is a subtype of T, and the type of the expression in each spread-list-member is a subtype of T[]."
If we do it like this something like "[int, string, ...int] = [5, ...[1,2], "foo"]" will work. Instead I think what you need to do is go over members as normal until you read any spread member (implicit or other wise), these would become you fixed length members. All the members after that  you need to take the intersection of either int projection for spread and actual type of non spread. that gives you the rest type. Note sure if you need to explicitly detect if the result is never we'll detect this naturally given no expression can have type never

semantics/type_resolver.go
```diff
@@ -3234,6 +3548,10 @@ func resolveListConstructorInner(t typeResolver, chain *binding, expr *ast.BLang
 }
 
 func resolveListConstructorWithExpectedType(t typeResolver, chain *binding, expr *ast.BLangListConstructorExpr, expectedType semtypes.SemType) (semtypes.SemType, expressionEffect, bool) {
+	if resultType, lat, ok := expectedListAtomicType(t, expectedType); ok {
+		return resolveListConstructorWithAtomicType(t, chain, expr, resultType, lat)
+	}
+
```
heshanpadmasiri: It is not clear to me why this is needed?

semantics/type_resolver_narrowing.go
```diff
@@ -40,6 +40,9 @@ type binding struct {
 	// When lookupBinding crosses this marker and finds a narrowed binding beyond it,
 	// the variable is captured from the outer scope and should be treated as unnarrowed.
 	functionBoundary bool
+	// queryAggregated marks a query variable that was converted to an aggregated
+	// variable binding by a group by clause.
+	queryAggregated bool
```
heshanpadmasiri: NIT: Given we now have multiple flags may be now it's worth it turn in to bitset

heshanpadmasiri (changes requested):
Apart form list constructor rest of the changes LGTM

ast/clauses.go
gimantha: replaced with if

semantics/type_resolver.go
gimantha: Addressed with testcase spread1-v.bal and spread1-e.bal

heshanpadmasiri (dismissed):
LGTM

warunalakshitha (approved):
LGTM

PR #458, opened 2026-05-14
Fix float equality and non-panicking decimal division

### LINKED ISSUE #364: [Bug]: Float and decimal division should never cause a panic

### LINKED ISSUE #391: [Bug]: Float ==/!= semantics were wrong for NaN in runtime

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/364
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/391

## Summary by CodeRabbit

* **Tests**
  * Added new test cases for NaN float value handling and edge case scenarios.
  * Expanded integration test fixtures for float, comparable, and const operations.
  * Updated test expectations for float remainder operations.

* **Bug Fixes**
  * Improved float equality comparison semantics for edge cases involving NaN and signed zero values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ballerina-platform/ballerina-lang-go/pull/458)


values/equal.go
```diff
@@ -51,7 +51,7 @@ func deepEqual(v1, v2 BalValue, visited map[refPair]struct{}) bool {
 		return ok && a == b
 	case *decimal.Decimal:
 		b, ok := v2.(*decimal.Decimal)
-		return ok && a.Cmp(b) == 0
+		return ok && !a.IsNaN() && !b.IsNaN() && a.Cmp(b) == 0
```
heshanpadmasiri: I don't think this is needed according to spec "+0 and -0 are not distinguished; if the coefficent is zero, then the sign is also constrained to be zero;
NaN, infinities and subnormals are not supported; operations that would result in one of these values according to the normal rules of IEEE 754-2008 instead result in a panic."

values/compare.go
```diff
@@ -81,6 +81,9 @@ func Compare(x, y BalValue) CompareResult {
 		return CmpEQ
 	case *decimal.Decimal:
 		v2 := y.(*decimal.Decimal)
+		if v1.IsNaN() || v2.IsNaN() {
+			return CmpUN
+		}
```
heshanpadmasiri: I don't think this can happen as well (Decimals can never be NaN in Ballerina)

decimal/decimal.go
```diff
@@ -27,21 +27,34 @@ const (
...
+	c := newDecimalAPDContext()
 	c.Traps = apd.SystemOverflow | apd.SystemUnderflow | apd.Overflow |
 		apd.DivisionUndefined | apd.DivisionByZero | apd.DivisionImpossible |
 		apd.InvalidOperation
 	return c
 }()
 
+// ieeeContext matches context's precision and exponent range but has traps
+// disabled so IEEE arithmetic can yield NaN or infinity instead of *Error.
+var ieeeContext = func() *apd.Context {
```
heshanpadmasiri: Not sure these changes to decimal is needed.

values/equal.go
warunalakshitha: Removed decimal fixes

corpus/bal/subset8/08-float/25-v.bal
```diff
@@ -0,0 +1,24 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// @productions float multiplicative-expr equality-expr floating-point-literal return-stmt function-call-expr assign-stmt local-var-decl-stmt
+import ballerina/io;
+
+public function main() {
```
warunalakshitha, line 20: This was added part of last PR comment

heshanpadmasiri (approved):
LGTM

PR #501, opened 2026-05-26
Add support for string template expressions

### LINKED ISSUE #412: Add support for string template expressions

heshanpadmasiri:
## Purpose
$subject
resolves #412 

## Summary by CodeRabbit

* **New Features**
  * Added template expression support for string interpolation (embedded ${} expressions), with compile-time type validation (boolean, int, float, decimal, string) and runtime evaluation producing correct concatenated results.
  * XML interpolation remains unimplemented and reports a clear unimplemented error.

* **Tests**
  * Added comprehensive tests and fixtures: nested, adjacent, empty, boundary, type-error, constant/constness, panic, and integration expectations.


desugar/expression.go
```diff
@@ -430,6 +432,22 @@ func walkFieldBaseAccess(cx *functionContext, expr *ast.BLangFieldBaseAccess) de
...
+	if len(expr.Insertions) == 0 {
+		lit := &ast.BLangLiteral{Value: expr.Strings[0], OriginalValue: expr.Strings[0]}
+		lit.SetPosition(expr.GetPosition())
+		lit.SetDeterminedType(semtypes.StringConst(expr.Strings[0]))
+		return desugaredNode[ast.BLangActionOrExpression]{replacementNode: lit}
+	}
+	var initStmts []ast.StatementNode
+	for i, ins := range expr.Insertions {
+		r := walkExpression(cx, ins)
+		initStmts = append(initStmts, r.initStmts...)
```
gimantha, line 445: The string templates should be evaluated in the order they occur right?

It seems the below code executes in a different order.

```bal
import ballerina/io;

    io:println(string `${first()} ${check second()} ${third()}`);
}

function first() returns string {
    io:println("first");
    return "A";
}

function second() returns string|error {
    io:println("second");
    return "B";
}

function third() returns string {
    io:println("third");
    return "C";
}
```

output
```
second
first
third
A B C
```

semantics/type_resolver.go
```diff
@@ -2412,6 +2414,22 @@ func resolveXMLElementLiteral(t typeResolver, chain *binding, e *ast.BLangXMLEle
...
+	var ty semtypes.SemType
+	if len(e.Insertions) == 0 {
+		ty = semtypes.StringConst(e.Strings[0])
+	} else {
+		for _, ins := range e.Insertions {
+			if _, _, ok := resolveActionOrExpression(t, chain, ins, templateInsertionAllowedTypes); !ok {
+				return nil, expressionEffect{}, false
+			}
+		}
+		ty = semtypes.STRING
```
gimantha: Insertions can also be singleton right? in that case the following should work.

```bal
type Greeting "hello world";

public function main() {
    Greeting g = string `hello ${"world"}`;
}
```
at the moment it fails with incompatible types, `expected "hello world", got string`

gimantha: Also this should work too.

```bal
const GREETING = string `hello ${"world"}`;

public function main() {
}
```
At the moment this gives an error: ` expression is not a constant expression`

desugar/expression.go
heshanpadmasiri: This is due to a bug in desugar create https://github.com/ballerina-platform/ballerina-lang-go/issues/511 to track

heshanpadmasiri: We can add a fix (not necessarily a workaround given we don't give gurantees about the order in which eval expression for most expressions; template expression is an exception) for template expression in desugar (desugar each insertion in order they appear and assign them to desugared variable) not sure if that is a good idea though

semantics/type_resolver.go
heshanpadmasiri: fixed in b8bcfc0

heshanpadmasiri:
@coderabbitai review

gimantha (dismissed):
LGTM

runtime/internal/exec/non_terminators.go
```diff
@@ -345,6 +346,24 @@ func execNewXMLElement(ctx *extern.Context, instr *bir.NewXMLElement, frame *Fra
 	setOperandValue(ctx, instr.LhsOp, frame, &values.XMLElement{Name: name, Attributes: attrs, Namespaces: namespaces, Children: children})
 }
 
+func execEvalTemplateExpr(ctx *extern.Context, instr *bir.EvalTemplateExpr, frame *Frame) {
+	if instr.Kind != bir.TemplateKindString {
+		panic(fmt.Sprintf("unsupported template kind: %d", instr.Kind))
```
warunalakshitha, line 347: Can this be happen at runtime?

heshanpadmasiri: No it's a sanity check given we going to add more and more template kinds in the future

warunalakshitha (approved):
LGTM

PR #505, opened 2026-05-28
Migrate ballerina/io and ballerina/http to bala bundles

TharmiganK:
## Summary

- Replace the hard-coded Go intrinsic implementations of `ballerina/io` and `ballerina/http` with self-contained bala bundle packages under `lib/stdlibs/ballerina/`, removing ~1700 lines of compiler-internal scaffolding.
- The public API of each library is now written in Ballerina (`.bal` source + `Bala.toml`/`Ballerina.toml` metadata); the native backing code moves to `lib/stdlibs/ballerina/<name>/0.0.1/go1.2/native/`.
- Fix three compiler bugs that only surfaced once `http.Client` became a real Ballerina class:
  - `resolveClassDefinition` never added `init` to the class method table, breaking `new` expressions.
  - `desugarClassMethodDefaults` skipped `InitFunction`, so the default-parameter lambda for `Client.init` was never emitted.
  - The method loop in `resolveClassDefinition` iterated over a Go map, making default-param counter assignments non-deterministic.
- Add `parser.GetSyntaxTreeFromContent` and `context.CompilerContext.GetEnvironment` so test helpers and the new stdlib pre-compiler can parse in-memory sources without file I/O.
- Update test infrastructure (`testphases`, integration, package-resolution, extern-test, BIR serialization) and regenerate corpus snapshots.

## Test plan

- [x] `go build ./...` — no compilation errors
- [x] `go test ./corpus/... ./test_util/... ./extern-test/... ./corpus/package-resolution/...` — all pass
- [x] `go test ./semantics/... ./desugar/... ./bir/...` — all pass
- [x] Verify desugar corpus snapshots are deterministic across multiple runs:
      `go test ./corpus/ -run TestDesugar` twice and diff outputs
- [x] `go test ./extern-test/ -run TestHttpClient` — passes (requires network)

## Summary by CodeRabbit

* **New Features**
  * HTTP client API with remote methods, TLS/secure socket configuration, redirect handling, and response payload/header utilities.
  * Standard library `print()` and `println()` functions for variadic output.

* **Documentation**
  * Removed HTTP client support reference documentation.


test_util/testphases/phases.go
```diff
@@ -66,6 +68,75 @@ type PipelineResult struct {
...
+// stdlibEntry describes one embedded standard-library package to pre-compile.
+type stdlibEntry struct {
+	org     string
+	name    string
+	version string
+}
+
+// builtinStdlibs is the ordered list of standard-library packages baked into the
+// binary. They are compiled with no imports of their own, so order is irrelevant.
+var builtinStdlibs = []stdlibEntry{
```
warunalakshitha, line 80: can we walk through lib/stdlibs than hardcoding paths.

TharmiganK: But it requires reading `Dependencies.toml` per package and topologically sorting before compilation — otherwise alphabetical order will break when stdlibs start depending on each other

corpus/extern_test.go
```diff
@@ -627,6 +634,42 @@ func TestDependentlyTypedCrossModuleRoundtrip(t *testing.T) {
...
+	exported := semantics.ResolveSymbols(cx, pkg, importedSymbols)
+	if cx.HasErrors() {
+		return semantics.PackageIdentifier{}, model.ExportedSymbolSpace{}, false
+	}
+	semantics.ResolveTopLevelNodes(cx, pkg, importedSymbols)
+	if cx.HasErrors() {
+		return semantics.PackageIdentifier{}, model.ExportedSymbolSpace{}, false
+	}
+	return semantics.PackageIdentifier{OrgName: org, ModuleName: name}, exported, true
+}
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [c1aa68c](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/c1aa68c4a39e30cca7e5465a3e28f5e667a050bb)

cli/cmd/run.go
```diff
@@ -258,12 +258,17 @@ func runBallerina(cmd *cobra.Command, args []string) error {
...
-	// Dump BIR if requested
+	// Dump BIR if requested — skip stdlib/dependency packages (ballerina org)
+	// so the output only contains the user's own code.
 	tyEnv := project.Environment().TypeEnv()
 	if buildOpts.DumpBIR() {
 		prettyPrinter := bir.PrettyPrinter{}
 		tyCtx := semtypes.ContextFrom(tyEnv)
 		for _, birPkg := range birPkgs {
+			if birPkg.PackageID != nil && birPkg.PackageID.OrgName != nil &&
+				birPkg.PackageID.OrgName.Value() == "ballerina" {
```
azinneera: Why do we skip ballerina packages from dumpbir here?

TharmiganK: IIUC current tests only have the output with only user's code. I thought we need to exclude the ballerina packages bir. Should we include that as well?

TharmiganK: Had a chat with @warunalakshitha and he suggested to only keep the BIRs of the current package. Hence changed it like that via [68e6068](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/68e60687bedd828f0f9d0d95bc08371c39c1fadf)

corpus/extern_test.go
```diff
@@ -587,9 +590,13 @@ func TestDependentlyTypedCrossModuleRoundtrip(t *testing.T) {
 	// symbols. If dependent-fn metadata failed to survive serialization, type
 	// resolution of `int a = helper:inferred(0)` would widen to VAL (or error)
 	// and main compilation would fail.
+	// main.bal also imports ballerina/io, so compile io in env2 first.
 	publicSymbols := map[semantics.PackageIdentifier]model.ExportedSymbolSpace{
 		{OrgName: org, ModuleName: helperMod}: deserializedHelperExported,
 	}
+	if ioID, ioExported, ok := compileBuiltinInEnv(env2, "ballerina", "io", "0.0.1"); ok {
```
warunalakshitha, line 598: why we hanlde only for io?

test_util/testphases/phases.go
warunalakshitha, line 80: should this be public. Dont we ned this order when building the executer?

corpus/bir_reserialization_test.go
```diff
@@ -35,7 +36,19 @@ import (
 // list, on top of the shared test_util.UnsupportedTests baseline (which is
 // applied via isTestSkipped below). Currently empty -- every known failure
 // is already covered by the shared baseline.
-var birSerializationRoundtripSkipList = []string{}
+// birSerializationRoundtripSkipList skips tests whose BIR cannot be
+// round-tripped in isolation because they depend on stdlib modules whose
+// Ballerina-level functions (e.g. http:Client.init) live in the http BIR
+// rather than being registered as native Go functions. Running only the
```
warunalakshitha: Do we need this comment? Should those be generic

parser/parser.go
```diff
@@ -14676,6 +14676,22 @@ func (b *BallerinaParser) isSpecialMethodName(token tree.STToken) bool {
 	return (((token.Kind() == common.MAP_KEYWORD) || (token.Kind() == common.START_KEYWORD)) || (token.Kind() == common.JOIN_KEYWORD))
 }
 
+// GetSyntaxTreeFromContent parses Ballerina source from an in-memory string.
+// virtualPath is used only for diagnostic messages; no file I/O is performed.
```
warunalakshitha, line 14680: Why we need this change in the parser?

corpus/extern_test.go
TharmiganK: This particular test only has `ballerina/io` import. That's why only handling io

test_util/testphases/phases.go
TharmiganK: No, it is only required for running the tests. So no need to be public

parser/parser.go
TharmiganK: We need this to load the syntax tree from the string content(the existing GetSyntaxTree method loads it from file system). This is only need to implement - `loadBuiltinPublicSymbols` in the test phases package

corpus/bir_reserialization_test.go
TharmiganK: Ack

TharmiganK: Changed to a generic function via [f47fc57](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/f47fc57788d43791702fe33e548fd607e80c8ab3)

cli/cmd/run.go
```diff
@@ -258,12 +258,19 @@ func runBallerina(cmd *cobra.Command, args []string) error {
...
+	// (same org + package name), so all sub-modules are covered while external
+	// imports are excluded.
 	tyEnv := project.Environment().TypeEnv()
 	if buildOpts.DumpBIR() {
 		prettyPrinter := bir.PrettyPrinter{}
 		tyCtx := semtypes.ContextFrom(tyEnv)
+		rootOrgName := pkg.PackageOrg().Value()
+		rootPkgName := pkg.PackageName().Value()
 		for _, birPkg := range birPkgs {
+			if birPkg.PackageID.OrgName.Value() != rootOrgName || birPkg.PackageID.PkgName.Value() != rootPkgName {
```
warunalakshitha, line 271: @azinneera is this best way check given module is part of root package?

cli/cmd/run.go
```diff
@@ -258,12 +258,19 @@ func runBallerina(cmd *cobra.Command, args []string) error {
 		fmt.Fprint(os.Stderr, compilation.StatsReport())
 	}
 
-	// Dump BIR if requested
+	// Dump BIR if requested — only include packages belonging to the root package
```
heshanpadmasiri, line 261: I am not sure if this filtering is needed given this flag is for us to help with debugging

context/context.go
```diff
@@ -182,6 +182,11 @@ func (c *CompilerContext) GetTypeEnv() semtypes.Env {
 	return c.env.GetTypeEnv()
 }
 
+// GetEnvironment returns the shared compiler environment backing this context.
+func (c *CompilerContext) GetEnvironment() *CompilerEnvironment {
```
heshanpadmasiri: As far as I can see this is used only for test suite. If so testsuite should figure out how to get this information (it should somewhere have the env given it should use it to create the context), instead of modifying the frontend API

corpus/desugared/subset8/08-const/2-v.txt
```diff
@@ -6,15 +6,15 @@
   (const N3 () ())
   (function init () ()
     (block-function-body
-      (assignment
```
heshanpadmasiri, line 12: This change is not needed (desugars toplogical sort is non deterministic by design) and test harness is smart enought to handle it

corpus/bir_reserialization_test.go
```diff
@@ -31,11 +32,16 @@ import (
...
-// applied via isTestSkipped below). Currently empty -- every known failure
-// is already covered by the shared baseline.
-var birSerializationRoundtripSkipList = []string{}
+// applied via isTestSkipped below). It skips tests whose BIR cannot be
+// round-tripped in isolation because they depend on stdlib modules whose
+// Ballerina-level functions (e.g. http:Client.init) live in the http BIR
+// rather than being registered as native Go functions. Running only the
+// deserialized consumer BIR in a fresh env would panic on those calls.
+var birSerializationRoundtripSkipPrefixes = []string{
+	"08-network/http-client",
```
heshanpadmasiri: Do we have an issue to track this? typically I add a comment with the issue number next to all skips

cli/cmd/run.go
azinneera: Yes

azinneera (dismissed):
LGTM

parser/parser.go
```diff
@@ -14676,6 +14676,24 @@ func (b *BallerinaParser) isSpecialMethodName(token tree.STToken) bool {
 	return (((token.Kind() == common.MAP_KEYWORD) || (token.Kind() == common.START_KEYWORD)) || (token.Kind() == common.JOIN_KEYWORD))
 }
 
+// GetSyntaxTreeFromContent parses Ballerina source from an in-memory string.
+// It is the counterpart to GetSyntaxTree for sources that have no real
+// filesystem path (e.g. stdlib .bal files embedded via embed.FS).
+// virtualPath is used only for diagnostic messages; no file I/O is performed.
+func GetSyntaxTreeFromContent(ctx *context.CompilerContext, virtualPath, content string) (*tree.SyntaxTree, error) {
```
heshanpadmasiri: I think the correct fix is to change GetSyntax tree to get content string and pass the responsibility of how to read the file to driver
```go
// GetSyntaxTree parses content into a syntax tree, attributing it to fileName
// (used for diagnostics and the syntax tree's text document).
func GetSyntaxTree(ctx *context.CompilerContext, fileName string, content string) (*tree.SyntaxTree, error) {
	// Create CharReader from file content
	reader := text.CharReaderFromText(content)

	// Create Lexer
	lexer := NewLexer(reader)

	// Create TokenReader from Lexer
	tokenReader := CreateTokenReader(lexer)

	// Create Parser from TokenReader
	ballerinaParser := NewBallerinaParserFromTokenReader(tokenReader)

	// Parse the entire file (parser will internally call tokenizer)
	rootNode := ballerinaParser.Parse().(*tree.STModulePart)

	moduleNode := tree.CreateUnlinkedFacade[*tree.STModulePart, *tree.ModulePart](rootNode)
	syntaxTree := tree.NewSyntaxTreeFromNodeTextDocument(moduleNode, nil, fileName, false)
	if syntaxTree.HasDiagnostics() {
		ctx.SyntaxError("syntax error at", diagnostics.Location{})
	}
	return &syntaxTree, nil
}
```

semantics/symbol_resolver.go
```diff
@@ -1236,7 +1227,10 @@ func resolveClassDefinition(ms *moduleSymbolResolver, classDef *ast.BLangClassDe
 		allocateDefaultParamSymbols(ms, ms.scope, classDef.InitFunction)
```
heshanpadmasiri, line 1244: looking at this in retrospecitive I would first build an array (whose order is deterministic) to have all the functions and do the symbol resolution, default param allocation, and method table update all in a single pass to avoid touching the memory in mulitple places with map iterators and sort. Something like 
```go
type namedClassMethod struct {
	name     string
	function *ast.BLangFunction
}

// classMethodsInResolutionOrder returns the class's init function (if any)
// followed by its methods in sorted order. The deterministic order keeps
// default-param symbol counter assignments stable across runs.
func classMethodsInResolutionOrder(classDef *ast.BLangClassDefinition) []namedClassMethod {
	ordered := make([]namedClassMethod, 0, len(classDef.Methods)+1)
	if classDef.InitFunction != nil {
		ordered = append(ordered, namedClassMethod{"init", classDef.InitFunction})
	}
	for _, methodName := range slices.Sorted(maps.Keys(classDef.Methods)) {
		ordered = append(ordered, namedClassMethod{methodName, classDef.Methods[methodName]})
	}
	return ordered
}
...

	for _, field := range classDef.Fields {
		ast.Walk(classResolver, field.(ast.BLangNode))
	}

	// Resolve init and methods in a single deterministic pass so that
	// default-param symbol counter assignments (allocated from a shared module
	// counter) are stable regardless of Go's map iteration order. init is placed
	// first, followed by methods in sorted order.
	orderedMethods := classMethodsInResolutionOrder(classDef)

	methodTable := make(map[string]model.SymbolRef, len(orderedMethods))
	for _, named := range orderedMethods {
		methodResolver := newFunctionResolver(classResolver, named.function)
		named.function.SetScope(methodResolver.scope)
		resolveFunction(methodResolver, named.function)
		allocateDefaultParamSymbols(ms, ms.scope, named.function)
		methodTable[named.name] = named.function.Symbol()
	}

	classSym := ms.ctx.GetSymbol(classDef.Symbol()).(*model.ClassSymbol)
	classSym.SetMethods(methodTable)
```

heshanpadmasiri (commented):
Few nits: overall changes LTGM

cli/cmd/run.go
TharmiganK: @warunalakshitha FYI

corpus/bir_reserialization_test.go
TharmiganK: Fixed the root cause instead via [8c3f70f](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/8c3f70f2e54b2c498455a3b1675d7462af804d42). The skip was introduced because the roundtrip test was only serializing the consumer BIR (the last package in topological order), leaving the http stdlib BIR absent from the fresh env. When the deserialized consumer called http:Client.init — a Ballerina-level wrapper function whose BIR body lives in the http package — the interpreter panicked.

The fix is to round-trip all BIR packages (in topological order), serialize and deserialize each into the same fresh CompilerEnvironment, then run them all. This restores the original intent of the test: verify that whatever was compiled survives serialization and produces correct output. The http-client tests now pass without any skip list.

semantics/symbol_resolver.go
TharmiganK: Addressed via [8c3f70f](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/8c3f70f2e54b2c498455a3b1675d7462af804d42)

parser/parser.go
TharmiganK: Addressed via [8c3f70f](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/8c3f70f2e54b2c498455a3b1675d7462af804d42)

corpus/desugared/subset8/08-const/2-v.txt
TharmiganK: Addressed via [8c3f70f](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/8c3f70f2e54b2c498455a3b1675d7462af804d42)

context/context.go
TharmiganK: Addressed via [8c3f70f](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/8c3f70f2e54b2c498455a3b1675d7462af804d42)

cli/cmd/run.go
warunalakshitha: I think dump-bir should only dump modules in the given package. Dumping all imported modules adds unnecessary noise when debugging. This is similar to the behavior we had for code coverage /runnig tests in the current JBallerina implementation.

test_util/skip.go
```diff
@@ -190,3 +190,16 @@ func MatchesSkip(path string, entries []string) bool {
 	}
 	return false
 }
+
+// MatchesSkipSubstr reports whether path contains any of the given substrings.
+// Use this for directory-level or prefix-based skip rules where all affected
+// tests share a common path segment (e.g. "08-network/http-client").
+func MatchesSkipSubstr(path string, substrings []string) bool {
```
warunalakshitha: This seems not calling.

heshanpadmasiri (dismissed):
FE changes LGTM

test_util/skip.go
TharmiganK: Ack, will remove this

TharmiganK: Removed via [8c2103a](https://github.com/ballerina-platform/ballerina-lang-go/pull/505/commits/8c2103a63c894a91d43c43a42082baa53a0a1e5b)

warunalakshitha (dismissed):

warunalakshitha (approved):
LGTM

PR #506, opened 2026-05-28
Support cloneWithType and fromJsonWithType lang libs

### LINKED ISSUE #502: Add `fromJsonWithType` runtime API support(both Ballerina and Go version)
### Description

> $Subject

### Describe your problem(s)

The initial data-binding support in the standard libraries requires this API. 

### Describe your solution(s)

Introduce the APIs which can be called from Ballerina side and Go side.

### Related area

Runtime

_No response_

_No response_

_No response_

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/502

## Summary

* **New Features**
  * Added `cloneWithType` and `fromJsonWithType` for type-driven deep cloning and JSON deserialization using `typedesc` targets.
* **Bug Fixes**
  * Improved conversion/cast diagnostics with clearer “bad type cast” messaging.
  * Refined dependently-typed `typedesc` parameter inference behavior.
* **Tests**
  * Added new corpus and integration cases covering function/object/stream `tostring`, `typedesc` errors, numeric conversion edges, unions, negatives, and cycle scenarios.
* **Chores**
  * Updated expected outputs for adjusted diagnostics and stringification formatting.


values/convertible.go
```diff
@@ -0,0 +1,211 @@
...
+) bool {
+	atomic := semtypes.ToMappingAtomicType(tc, target)
+
+	for _, name := range atomic.Names {
+		if _, ok := source.Get(name); ok {
+			continue
+		}
+		if !fieldMayOmitKey(tc, target, name, atomic) {
+			return false
+		}
```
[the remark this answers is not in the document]
warunalakshitha: Fixed

corpus/cli/output/run/workspaces/missing-member.txtar
```diff
@@ -1,6 +1,11 @@
 -- stdout --
 -- stderr --
 ballerina: no package found at path
+notamember
```
heshanpadmasiri: Is this change intentionally (I am fine with including if so; just to check)

values/convert.go
```diff
@@ -0,0 +1,223 @@
...
+package values
+
+import (
+	"math"
+
+	"ballerina-lang-go/decimal"
+	"ballerina-lang-go/semtypes"
+)
+
+// Convert converts a value to the given target type.
```
heshanpadmasiri: I think this need to give a specific contract on how it converts a given value to another. AFAIK this don't correspond to any abstract operation defined in spec

values/convert.go
```diff
@@ -0,0 +1,223 @@
...
+// On failure it returns a lang.value ConversionError as *Error.
+func Convert(tc semtypes.Context, value BalValue, targetType semtypes.SemType) (BalValue, *Error) {
+	var unionErrors []string
+	convertibleType, err := getConvertibleType(tc, value, targetType, &unionErrors, true)
```
heshanpadmasiri: Wouldn't it be more efficient to create a list of candidate types by breaking target type to atomic types and then against each try to convert value as a fallable operation. Against each convertable type we allow numeric conversion in convert operation. 

Current approch seems to walk the whole value against each candidate type ones do some checks then again for each such type try finish the conversion again walking each candidate type and value twice

values/convert.go
```diff
@@ -0,0 +1,223 @@
...
+	case float64:
+		d, err := decimal.FromFloat64(v)
+		if err != nil {
+			return nil, newConversionFailure(err.Error())
+		}
+		return d, nil
+	default: // int64
+		return decimal.FromInt64(value.(int64)), nil
+	}
+}
```
heshanpadmasiri: Ideally numeric conversion and is valid numeric conversion checking logic should be shared with cast instruction handling

values/never.go
```diff
@@ -14,7 +14,6 @@
 // specific language governing permissions and limitations
 // under the License.
```
heshanpadmasiri, line 16: We need to have package comment somewhere. No strong opinion on where it should be

semantics/type_resolver.go
```diff
@@ -5490,7 +5511,13 @@ func argArray(t typeResolver, sym model.FunctionSymbol, paramTypes []semtypes.Se
 				}
 				ctx := t.typeContext()
 				T := semtypes.TypedescConstraint(ctx, paramTypes[i])
-				S := semtypes.Intersect(T, callExpectedType)
+				S := semtypes.Diff(callExpectedType, semtypes.ERROR)
```
heshanpadmasiri: I don't think this is correct consider
```ballerina
   error|int withError = inferredErrorOrInt();
   if withError is error {
       io:println("error"); // @output error
   } else {
       io:println(withError);
   }

   function inferredErrorOrInt(typedesc<any|error> retTy = <>) returns retTy = external;

```
The core issue here as far as I can see is if you have

T v = check <call-exp>

currently we treat the contextually expected type of <call-exp> to be T|error. 

According to spec "If the default value for a defaultable-param was specified as <>, then the value for the parameter will be a typedesc value determined at compile-time from the contextually expected type of the function-call-expr as follows. Let the name of the parameter for which a default of <> was specified be t. Let the type of t be typedesc<T>. Let the return type descriptor for the function be R, which will have references to t. Let the contextually expected type for the function-call-expr be C. Then the default value is found by unifying C and R: the default value is a typedesc value that describes the maximal type S, such that S is a subtype of T, and R with S substituted for t is a subtype of C." I don't fully understand how you get the maximal part. 

I can think of a way to tighten the approximation better than what we have now by, in TypeOp (symbol.go) in addition to apply you add a `FixedPart() Semtype`(need a better name for this) this gives you the "non-dependent" part of the function. For Identity this is the type for ref `never` and binary can work from that. You then remove from the contextually expected type that fixed part let's call this t'. Then as before take intersection of t' with contextually expected type.

heshanpadmasiri (changes requested):
I couldn't find a spec that says what is the behaviour of fromJson should be but these cases have different behaviours with jballerina
```ballerina
type Node record {|
    int value;
    Node? next;
|};

public function main() {
    map<json> cyclic = {value: 1};
    cyclic["next"] = cyclic;
    json j = cyclic;
    Node _ = checkpanic j.fromJsonWithType(Node); // @panic
}
```

```ballerina
import ballerina/io;

isolated function defaultAge() returns int {
    io:println("default called");
    return 42;
}

type Person record {|
    string name;
    int age = defaultAge();
|};

    json j = {"name": "Alice"};
    Person p = check j.fromJsonWithType(Person); // @output default called
    io:println(p); // @output {"name":"Alice","age":42}
}
```

```ballerina
type PersonNilableRequired record {|
    string name;
    int? age;
|};

public function main() {
    json missingNilableRequired = {"name": "Bob"};
    PersonNilableRequired _ = checkpanic missingNilableRequired.fromJsonWithType(PersonNilableRequired); // @panic
}
```

warunalakshitha:
> I couldn't find a spec that says what is the behaviour of fromJson should be but these cases have different behaviours with jballerina
> 
> ```ballerina
> type Node record {|
>     int value;
>     Node? next;
> |};
> 
> public function main() {
>     map<json> cyclic = {value: 1};
>     cyclic["next"] = cyclic;
>     json j = cyclic;
>     Node _ = checkpanic j.fromJsonWithType(Node); // @panic
> }
> ```
> 
> ```ballerina
> import ballerina/io;
> 
> isolated function defaultAge() returns int {
>     io:println("default called");
>     return 42;
> }
> 
> type Person record {|
>     string name;
>     int age = defaultAge();
> |};
> 
> public function main() returns error? {
>     json j = {"name": "Alice"};
>     Person p = check j.fromJsonWithType(Person); // @output default called
>     io:println(p); // @output {"name":"Alice","age":42}
> }
> ```
> 
> ```ballerina
> type PersonNilableRequired record {|
>     string name;
>     int? age;
> |};
> 
> public function main() {
>     json missingNilableRequired = {"name": "Bob"};
>     PersonNilableRequired _ = checkpanic missingNilableRequired.fromJsonWithType(PersonNilableRequired); // @panic
> }
> ```

Case 1 — cyclic value handling: ✅ Fixed and added a test
Case 2 — missing record field with default initializer: we can support once we have record default value runtime API. Will send as a separate PR.
Case 3 — missing required nilable field → error: ✅ Fixed and added a test.

corpus/cli/output/run/workspaces/missing-member.txtar
warunalakshitha: Fixed

values/convert.go
warunalakshitha: Initially, I followed the same approach used by JBallerina, which performed two passes: one to identify the candidate type and another to perform the actual conversion. I reduced this to a single pass by directly attempting the conversion for each candidate type as a fallible operation.

With the new implementation, performance has improved by 1×.

values/convert.go
warunalakshitha: Now both are using same numeric conversion logic.

values/never.go
warunalakshitha: added to values.go

semantics/type_resolver.go
warunalakshitha: Fixed as per the comment

values/convert.go
warunalakshitha: exposed cloneWithType
```ballerina
// CloneWithType implements the cloneWithType abstract operation defined in the Ballerina spec
// (https://ballerina.io/spec/lang/master/#section_16.6).
//
// It constructs a value of targetType by deep-cloning value, applying the following conversions:
//   - the inherent type of any structural value comes from targetType
//   - numeric values may be converted between int, float, and decimal via NumericConvert
//   - missing required fields (with or without defaults) cause a ConversionError; default injection
//     is not yet implemented (tracked as a separate work item)
//
// Cyclic values return a ConversionError (matching jballerina behaviour).
//
// On failure it returns a ConversionError wrapped as *Error.
func CloneWithType(tc semtypes.Context, value BalValue, targetType semtypes.SemType) (BalValue, *Error) {
```

semantics/type_resolver.go
warunalakshitha: Added a test as well

warunalakshitha:
@coderabbitai review

values/convert.go
```diff
@@ -0,0 +1,332 @@
...
+	for i := 0; i < source.Len(); i++ {
+		memberTy := atomic.MemberAtInnerVal(i)
+		converted, err := tryConvert(tc, source.Get(i), memberTy, allowNumeric, visiting)
+		if err != nil {
+			return nil, err
+		}
+		items[i] = converted
+	}
+
+	restFiller, _ := FillerFactoryFor(tc, atomic.Rest())
```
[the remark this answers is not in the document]
warunalakshitha: FillerFactoryFor (values/values.go:71) returns (FillerFactory, bool), not (FillerFactory, error). There's no error to propagate here.

Ignoring the bool is the established, correct pattern, not an oversight.

warunalakshitha:
@coderabbitai review

corpus/bal/subset9/09-value/clonewithtype-numeric-v.bal
```diff
@@ -0,0 +1,142 @@
...
+    // float[] → byte[]
+    anydata floatBytes = [1.0, 200.0, 255.0];
+    ByteArray ba = check floatBytes.cloneWithType(ByteArray);
+    io:println(ba); // @output [1,200,255]
+
+    // decimal[] → int[] (truncates each element)
+    decimal[] decArr = [1.23, 2.34, 3.45];
+    anydata decArrAny = decArr;
+    IntArray fromDecArr = check decArrAny.cloneWithType(IntArray);
+    io:println(fromDecArr); // @output [1,2,3]
```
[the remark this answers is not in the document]
warunalakshitha: @coderabbitai fixed

corpus/bal/subset9/09-value/fromjsonwithtype-closed-union-v.bal
```diff
@@ -0,0 +1,46 @@
...
+    io:println(ageOnlyPet); // @output {"age":5}
+
+    // Only PetByType fields match — single member selected.
+    json typeOnly = {"pet_type": "Cat"};
+    Pet typeOnlyPet = check typeOnly.fromJsonWithType(Pet);
+    io:println(typeOnlyPet); // @output {"pet_type":"Cat"}
+
+    // Both PetByAge and PetByType required fields present — ambiguous closed union → error.
+    json petJson = {"nickname": "Fido", "pet_type": "Dog", "age": 4};
+    io:println(petJson.fromJsonWithType(Pet) is error); // @output true
```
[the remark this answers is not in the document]
warunalakshitha: @coderabbitai fixed

lib/langlibs/ballerina/lang.value/0.0.1/any/lang.value.bal
```diff
@@ -16,3 +16,55 @@
 
 // This module currently exposes no symbols; it exists so that the lang.value
 // langlib resolves as a real bundle.
+
+# Constructs a value with a specified type by cloning another value.
```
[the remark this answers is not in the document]
warunalakshitha: Fixed

values/convert.go
```diff
@@ -0,0 +1,332 @@
...
+		return nil, cannotConvertNil(tc, target)
+	}
+
+	if members := unionMemberTypes(tc, target); len(members) > 1 {
+		return tryConvertUnion(tc, value, target, members, allowNumeric, visiting)
+	}
+
+	switch v := value.(type) {
+	case *Map:
+		if semtypes.IsSubtypeSimple(target, semtypes.MAPPING) {
```
heshanpadmasiri: It is better to use IsSubtype instead given you can't control how you get the target type

values/convert.go
```diff
@@ -0,0 +1,332 @@
...
+// On failure it returns a ConversionError wrapped as *Error.
+func CloneWithType(tc semtypes.Context, value BalValue, targetType semtypes.SemType) (BalValue, *Error) {
+	result, err := tryConvert(tc, value, targetType, true, nil)
+	if err != nil {
+		return nil, wrapConversionError(err)
+	}
+	return result, nil
+}
+
+func tryConvert(tc semtypes.Context, value BalValue, target semtypes.SemType,
```
heshanpadmasiri: NIT: I feel this can be much cleaner if you decompose this into the two orthogonal concerns.
1. Given a single semtype what are the candidate semtypes (unionMemberTypes) return a `[]semtype`
2. Then for each try to a single candiadate

I don't know if the spec have some gurantees about the ordering about candidates (I recall it has for numeric literals; don't know for others). candiate list should be ordered as such and convert purely try to convert to just that type.

corpus/bal/subset9/09-value/clonewithtype-numeric-v.bal
```diff
@@ -0,0 +1,143 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/io;
+
```
heshanpadmasiri, line 18: I think we need to test for more complicated types such as

```ballerina

type R R1|R2;
type R1 {|
    R3|R4 f1;
|};
type R2 {|
    R5|R6 f2;
|};

type R3 {|
   int f3;
|};

type R4 {|
   float f3;
|};

type R5 {|
  int f3;
|};

type R6 {|
  string f3;
|}
```
Where the candidate decision tree is more than one level deep

heshanpadmasiri (commented):
Couple of minor nits, front end LGTM

values/convert.go
warunalakshitha: As per the spec, we need to use  first (leftmost) such type descriptor. 

values.CloneWithType only ever sees a semtypes.SemType for the target — and SemType is a canonical, order-independent representation. int|float and float|int compile down to the exact same SemType value; there is no "member order" left to read at the point candidateTypes runs. it's that the information needed to do it right has already been discarded several stages upstream. 

Created https://github.com/ballerina-nutcracker/ballerina/issues/657

warunalakshitha: Refactor the code first get candidateList then try convert to each candidate

corpus/bal/subset9/09-value/clonewithtype-numeric-v.bal
warunalakshitha: Added a test

values/convert.go
```diff
@@ -0,0 +1,340 @@
...
+func CloneWithType(tc semtypes.Context, value BalValue, targetType semtypes.SemType) (BalValue, *Error) {
+	result, err := convert(tc, value, targetType, true, nil)
+	if err != nil {
+		return nil, wrapConversionError(err)
+	}
+	return result, nil
+}
+
+func convert(tc semtypes.Context, value BalValue, target semtypes.SemType,
+	allowNumeric bool, visiting map[BalValue]struct{}) (BalValue, *conversionFailure) {
```
heshanpadmasiri: This still feels off to me. The shape I would have expected is something like this
```go
func convert(ctx semtypes.Context, value BalValue, target semtypes.SemType, allowNumericConversion bool) (BalValue, error) {
     candidates := candidateTypes(ctx, target);
     for _, candidate := range(candidates) { 
            res, ok := tryConvert(ctx, value, candidate, allowNumericConversion, make(map[ValValue]struct{}));
            if ok {
                return res, nil
            }
     }
     return nil, error("failed to convert")
}
```

I am also not entierly sure if this don't need context (given stack trace in case of panic; if the error it self don't have the stack trace and we need to patch it after I'll just pass the error up and build a proper error value there)

heshanpadmasiri: candidates should given you possible candidates including basic types such that you only need a single tryConvert. Internally it could call a tryConvertBasicType that will do the numeric conversion. (and separate methods for each other basic type like tryConvertMap etc. So if the expected type is a map for a given part you call tryConvertMap with that part of the value and that part of the type). All tryConvert do is look at the value and target type if can be done dispatch to the correct fn. Also do cycle detection there.

semtypes/predefined_type.go
```diff
@@ -75,6 +75,13 @@ var (
 	NUMBER           = numberBits.semType()
 	SIMPLE_BASIC     = simpleBasicBits.semType()
 
+	// SimpleBasicTypes lists, as individual SemTypes, the basic types outside MAPPING and
+	// LIST that a caller decomposing a union may need to try one at a time (e.g. via
+	// Intersect). MAPPING and LIST decompose differently, through MappingAlternatives and
+	// ListAlternatives, since a single union can hold multiple distinct record/tuple shapes.
+	// Callers must treat this slice as read-only.
+	SimpleBasicTypes = []SemType{NIL, BOOLEAN, INT, FLOAT, DECIMAL, STRING, XML, ERROR}
```
heshanpadmasiri: NIT: Accroding to spec these are not basic types. String and XML are sequence types, Error is a behavioural value. Others are simple.
Anyway I feel their usage is specific enough that there will be little utility in giving this as public API in semtypes.

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+// (scalar) constituents, ones value already satisfies without numeric conversion are ordered
+// first, so a caller trying candidates in order finds an exact match before any that would widen
+// or narrow value numerically.
+func candidateTypes(tc semtypes.Context, value BalValue, ty semtypes.SemType) []semtypes.SemType {
+	var members []semtypes.SemType
+	basic := semtypes.WidenToBasicTypes(ty)
+
+	if semtypes.ContainsBasicType(basic, semtypes.MAPPING) {
+		mappingTy := semtypes.Intersect(ty, semtypes.MAPPING)
+		if !semtypes.IsEmpty(tc, mappingTy) {
```
heshanpadmasiri: NIT: Isn't this alway true? same for list as well

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+	if semtypes.ContainsBasicType(basic, semtypes.LIST) {
+		listTy := semtypes.Intersect(ty, semtypes.LIST)
+		if !semtypes.IsEmpty(tc, listTy) {
+			for _, alt := range semtypes.ListAlternatives(tc, listTy) {
+				members = append(members, alt.SemType)
+			}
+		}
+	}
+
+	valueTy := SemTypeForValue(value)
```
heshanpadmasiri: I would keep this independent of value, just a function of targetType. Also it is not clear to me the reason for having exact and coerced arrays

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+	return nil, incompatibleConversion(tc, value, target)
+}
+
+func tryConvertMap(tc semtypes.Context, source *Map, target semtypes.SemType, visiting map[BalValue]struct{}) (BalValue, *conversionFailure) {
+	var cycleErr *conversionFailure
+	visiting, cycleErr = enterCycleCheck(tc, source.Type, source, visiting)
+	if cycleErr != nil {
+		return nil, cycleErr
+	}
+	defer delete(visiting, source)
```
heshanpadmasiri: I think you call pull cycle detection in to tryConvert (common for both list and mapping)

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+
+	atomic := semtypes.ToMappingAtomicType(tc, target)
```
heshanpadmasiri: I would think all the candidates should be atomic. If we have concrete cases where *Alternative returns a non atomic type please create an issue

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func mappingFieldType(tc semtypes.Context, target semtypes.SemType, atomic *semtypes.MappingAtomicType, key string) semtypes.SemType {
+	if atomic != nil {
+		for _, name := range atomic.Names {
+			if name == key {
+				return atomic.FieldInnerVal(key)
+			}
+		}
+	}
```
heshanpadmasiri: As far as I can see you have already filtered non atomic cases. Why would you get atomic == nil?

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, cycleErr
+	}
+	defer delete(visiting, source)
+
+	atomic := semtypes.ToMappingAtomicType(tc, target)
+	if atomic == nil {
+		return nil, incompatibleConversion(tc, source, target)
+	}
+
+	closed := isClosedRecord(atomic)
```
heshanpadmasiri: I think you should have a simple projection on the atomic type (I think `mappingFieldType` already do most of this) that takes the field name and give you the expected type, you don't need to special case closed records. They'll get NEVER for that field and your value can't be converted to never

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+		for _, name := range atomic.Names {
+			if name == key {
+				return atomic.FieldInnerVal(key)
+			}
+		}
+	}
+	return semtypes.MappingMemberTypeInnerVal(tc, target, semtypes.StringConst(key))
+}
+
+func fieldMayOmitKey(tc semtypes.Context, target semtypes.SemType, name string) bool {
```
heshanpadmasiri: I assume this is checking if the field is optional, if so mapping atom have a method for this

values/convert.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, wrapConversionError(err)
+	}
+	return result, nil
+}
+
+// convert converts value to target, trying each of target's basic-type constituents in turn;
+// candidateTypes orders them so any constituent value already satisfies precedes one that would
+// need numeric conversion, so a single pass suffices to prefer an exact match. See
+// conversionFailureFor for how a total failure is reported.
+func convert(tc semtypes.Context, value BalValue, target semtypes.SemType, visiting map[BalValue]struct{}) (BalValue, *conversionFailure) {
```
heshanpadmasiri, line 63: Doesn't this still need allowNumeric convert, or are we always allowing it?

values/convert.go
heshanpadmasiri: Or you can use FieldInnerVal in atomic type

heshanpadmasiri (commented):
Shape of convert now looks good to me. Few minor details needs to be checked

semtypes/predefined_type.go
warunalakshitha: Initially this was added inside convert.go. I searched for similar function in semtypes, but couldn't. I'll revert this can added as public API if we get any requirement in the future.

values/convert.go
warunalakshitha: Yes.We allow it. cloneWithType and fromJsonWithType both spec operations permit numeric conversion

for a target like int|float, both INT and FLOAT are separate entries in nonStructuralBasicTypes, so candidateTypes builds a member for each one that's non-empty against ty. Whichever one value already satisfies goes in exact, everything else needing a numeric conversion goes in coerced, and exact is tried first — so an int64 value against int|float gets INT tried first (exact, no conversion) and only falls to FLOAT (coerced) if the INT member turns out empty.

values/convert.go
warunalakshitha: two arrays used to separate out numeric conversion and without numeric conversion.

ex.  a JSON value 12.0 (runtime type float) converted to IntFloat = int|float must stay float, not get coerced to int.

semtypes/predefined_type.go
warunalakshitha: Fixed

values/convert.go
warunalakshitha: Removed and Fixed

values/convert.go
warunalakshitha: Fixed

values/convert.go
warunalakshitha: Removed unnecessary if condition to check atomic.

values/convert.go
warunalakshitha: Removed.

values/convert.go
warunalakshitha: Fixed

values/convert.go
warunalakshitha: Fixed

values/convert.go
warunalakshitha: we do not need value but need type to decide which numeric candidate value already satisfies without conversion, so it's tried before one that would coerce it (e.g. keeping 12.0 as float instead of wrongly narrowing it to int when the target is int|float);

We can move to tryConvert or separate step, but this is clear. refactored candiateTypes get only value type.

heshanpadmasiri (approved):
LGTM

PR #520, opened 2026-06-03
Add unused variable detection to the frontend  

### LINKED ISSUE #439: Add unused variable detection to the frontend
### Description

$subject

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
Resolves #439


heshanpadmasiri:
@coderabbit review --dir semantics

semantics/symbol_resolver.go
```diff
@@ -858,7 +1017,8 @@ type variableNode interface {
 }
 
 func referVariable[T symbolResolver](resolver T, variable variableNode) {
-	resolveSymbolRef(resolver, variable.GetName().GetValue(), "", variable.GetPosition(), variable)
+	name := variable.GetName().GetValue()
+	resolveSymbolRef(resolver, name, "", variable.GetPosition(), variable)
```
warunalakshitha, line 1025: dont we need to add markUnprefixedRefUsed here?

semantics/symbol_resolver.go
```diff
@@ -520,6 +673,7 @@ func resolveLambdaFunction(functionResolver *blockSymbolResolver, parent *blockS
 		}
 		symbol := model.NewValueSymbol(name, false, false, true)
 		addSymbolAndSetOnNode(functionResolver, name, &symbol, param)
+		markInit(functionResolver, name, param.Symbol(), param.GetPosition())
```
warunalakshitha, line 679: can we have unused lambda functions?

corpus/integration/project/import-const6-e.txtar
```diff
@@ -11,3 +11,9 @@ error[SEMANTIC_ERROR]: Unknown symbol: X
    |
 19 | const Y = root:X; // @error
    |           ^^^^^^
+
+error[SEMANTIC_ERROR]: unused variable
```
warunalakshitha: should we add variable name here?

semantics/symbol_resolver.go
heshanpadmasiri: you can do something like 
```ballerina
var _ = function (int x) returns int {...}
```
Not sure if there are any other forms, either way we are not trying to detect them

semantics/symbol_resolver.go
heshanpadmasiri: `referVariable` is reached for `ast.VariableNode`, which in includes declaration nodes such as the `BLangSimpleVariable` inside `BLangSimpleVariableDef` after `defineVariable` has already registered the symbol. Adding `markUnprefixedRefUsed` here would make a declaration mark itself as used during the normal AST walk. Instead we are handling this explicitly in VarRef nodes

corpus/integration/project/import-const6-e.txtar
heshanpadmasiri: fixed

warunalakshitha (approved):
LGTM

PR #507, opened 2026-05-29
Add runtime support for graceful and immediate stop

heshanpadmasiri:
## Purpose
$subject
Resolves #475
Depends on #484

## Summary by CodeRabbit

* **New Features**
  * Added listener lifecycle hooks (`init`, `'start()`, `gracefulStop()`, `immediateStop()`) and improved lifecycle dispatch order.
  * Introduced `runtime:onGracefulStop` to register shutdown handlers.
  * Improved runtime startup/shutdown flow for CLI execution (initializes packages before listening and reports exit status after completion).
* **Bug Fixes**
  * Improved handling of deserialized non-public symbol duplicates to avoid inconsistent symbol resolution.
* **Tests**
  * Added a unified integration/extern test harness plus signal/lifecycle test utilities.

PR #526, opened 2026-06-04
Add support for ephimaral type atoms for runtime 

### LINKED ISSUE #471: [Bug]: Deal with list type creation in runtime for rest args
### Description

We create a List for rest arguments (both rest args and rest path segment). For it's type we create a list atom which measn repeated call to the same function with rest args will leak. We need a better way to handle it. Possible options I see are

1. Introduce some sort of ephemeral list atom type that is not tracked by the env. This solves the memory leak issue. We can always gurantee that these types are not cyclic
2. Create the list type once per call site and reuse it (list type here don't change so shouldn't be a problem)

I am more inclined towards first because we may need these with data binding as well

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Purpose
Resolves #471 
Depends on #507 

## Summary by CodeRabbit

* **Bug Fixes**
  * Shortened type-cast failure diagnostics to a concise `error: bad type cast`.
  * Updated quoted-identifier duplicate reporting to show a duplicate-definition error (`Variable already defined: foo`) instead of an unused-variable message.
* **Improvements**
  * Made the lifecycle listener instance publicly accessible for easier integration.
  * Improved runtime type environment initialization and semtype handling to reduce overhead and avoid unnecessary lazy setup.

### DIFF
--- a/corpus/ast/subset9/09-object/listener-lifecycle1-v.txt
+++ b/corpus/ast/subset9/09-object/listener-lifecycle1-v.txt
@@ -56,8 +56,6 @@
   (variable l (type
     (user-defined-type LifecycleListener)) (expr
     (new ())))
-  (variable _ (expr
-    (simple-var-ref l)))
   (function main () (
     (value-type null))
     (block-function-body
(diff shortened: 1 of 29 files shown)


warunalakshitha (approved):
LGTM

PR #527, opened 2026-06-08
Add support for XML template expressions

### LINKED ISSUE #413: Add support for xml template expressions

heshanpadmasiri:
## Purpose
$subject
Resolves #413 

## Summary by CodeRabbit

* **New Features**
  * Added XML template expression support, enabling dynamic XML construction with interpolated values using `${...}` syntax
  * Supports content and attribute interpolation with automatic XML escaping
  * Includes namespace resolution and declaration for templated XML elements
  * Enforces validation to prevent invalid interpolations in comments, processing instructions, CDATA sections, and xmlns attributes

### DIFF
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -114,10 +114,17 @@ const (
 
 type TemplateExprKind uint8
 
+type XMLTemplateInsertionKind uint8
+
 const (
 	TemplateExprKindString TemplateExprKind = iota
 	TemplateExprKindXML
 	TemplateExprKindRaw
+)
+
+const (
+	XMLTemplateInsertionKindContent XMLTemplateInsertionKind = iota
+	XMLTemplateInsertionKindAttribute
 )
 
 type (
@@ -398,6 +405,19 @@ type (
 		Kind       TemplateExprKind
 		Strings    []string
 		Insertions []BLangExpression
+	}
+
+	XMLTemplateNamespaceInsertion struct {
+		Offset         int // this is the offest in the string where we need to insert the namespace declarations when we desugar to BLangTemplateExpr
+		UsedPrefixes   map[string]struct{}
+		NeedsDefaultNS bool
+		Namespaces     map[string]string
+	}
+
+	BLangXMLTemplateExpr struct {
+		BLangTemplateExpr
+		InsertionKinds      []XMLTemplateInsertionKind        // This tracks where do we do the insertion for each expression
+		NamespaceInsertions [][]XMLTemplateNamespaceInsertion // namespace insertion points for each template string
 	}
 
 	BLangXMLElementLiteral struct {
@@ -1035,6 +1055,7 @@ func createBLangUnaryExpr(location diagnostics.Location, operator model.Operator
 var (
 	_ BLangExpression = &BLangXMLSequenceLiteral{}
 	_ BLangExpression = &BLangTemplateExpr{}
+	_ BLangExpression = &BLangXMLTemplateExpr{}
 	_ BLangExpression = &BLangXMLElementLiteral{}
 	_ BLangExpression = &BLangXMLAttribute{}
 	_ BLangExpression = &BLangXMLPILiteral{}
@@ -1042,6 +1063,7 @@ var (
 	_ BLangExpression = &BLangXMLTextLiteral{}
 	_ BLangNode       = &BLangXMLSequenceLiteral{}
 	_ BLangNode       = &BLangTemplateExpr{}
+	_ BLangNode       = &BLangXMLTemplateExpr{}
 	_ BLangNode       = &BLangXMLElementLiteral{}
 	_ BLangNode       = &BLangXMLAttribute{}
 	_ BLangNode       = &BLangXMLPILiteral{}
--- a/desugar/expression.go
+++ b/desugar/expression.go
@@ -19,11 +19,14 @@ package desugar
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
 	"ballerina-lang-go/ast"
 	"ballerina-lang-go/model"
 	"ballerina-lang-go/semtypes"
 	"ballerina-lang-go/tools/diagnostics"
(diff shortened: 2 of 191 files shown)


ast/expressions.go
```diff
@@ -376,6 +383,19 @@ type (
 		Insertions []BLangExpression
 	}
 
+	XMLTemplateNamespaceInsertion struct {
+		Offset         int // this is the offest in the string where we need to insert the namespace declarations when we desugar to BLangTemplateExpr
+		UsedPrefixes   map[string]struct{}
+		NeedsDefaultNS bool
+		Namespaces     map[string]string
+	}
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

runtime/internal/exec/non_terminators.go
gimantha, line 366: Now this appending happens before the re-parsing right?

This code seems failing. `<child/>` is taken as a element and `a&b` cause parser panic

```bal
import ballerina/io;

public function main() {
    string markup = "<child/>";
    string amp = "a&b";
    xml x = xml <root>${markup}|${amp}</root>;
    io:println(x);
}
```

semantics/type_resolver.go
```diff
@@ -2503,6 +2505,21 @@ func resolveStringTemplateType(t typeResolver, chain *binding, e *ast.BLangTempl
 	return semtypes.STRING, true
 }
 
+func resolveXMLTemplateExpr(t typeResolver, chain *binding, e *ast.BLangXMLTemplateExpr) (semtypes.SemType, expressionEffect, bool) {
+	if len(e.InsertionKinds) != len(e.Insertions) {
+		t.internalError(fmt.Sprintf("xml template insertion kind count mismatch: got %d kinds for %d insertions", len(e.InsertionKinds), len(e.Insertions)), e.GetPosition())
+		return nil, expressionEffect{}, false
+	}
+	for i, ins := range e.Insertions {
+		allowed := xmlTemplateInsertionAllowedTypes(e.InsertionKinds[i])
```
gimantha, line 2514: Ideally following code shoud return a xml, but it returns [xml...]

```bal
import ballerina/io;

public function main() {
    int[] values = [1, 2];
    xml x = xml `<root>${from var n in values select xml `<n>${n}</n>`}</root>`;
    io:println(x);
}
```

gimantha:
```bal
import ballerina/io;

xml XML_VALUE = xml `<root>${"x"}</root>`;

public function main() {
    io:println(XML_VALUE);
}

```
This code also fails. If this is not supposed to be implemented in this PR, we can consider it seperately

runtime/internal/exec/non_terminators.go
gimantha:
```bal
import ballerina/io;

public function main() {
    string value = "a&b\"c<d";
    xml x = xml `<root value="${value}"/>`;
    io:println(x);
}
```
This code also has the same problem. We might have to call `xmlTemplateAttrValueEscaper:` in expression.go:503

semantics/xmlns_resolver.go
```diff
@@ -288,3 +288,71 @@ func mergeNamespaces(root *ast.BLangXMLElementLiteral, extras map[string]string)
...
+					insn.Namespaces["xmlns"] = uri
+				}
+			}
+			for prefix := range insn.UsedPrefixes {
+				uri, _, ok := scope.LookupXMLNS(prefix)
+				if !ok {
+					semanticError(resolver, "undefined XML namespace prefix '"+prefix+"'", e.GetPosition())
+					continue
+				}
+				for p, u := range xmlTemplatePrefixesForURI(scope, uri) {
```
gimantha: check this code.

```bal
import ballerina/io;

xmlns "urn:test" as a;
xmlns "urn:test" as z;

public function main() {
    string value = "x";
    xml x = xml `<a:root>${value}</a:root>`;
    io:println(x);
}
```
Output is: `<z:root xmlns:a="urn:test" xmlns:z="urn:test">x</z:root>`

the template uses a:root, but the output serializes it as z:root. Both a and z map to urn:test, so they are semantically equivalent, but the prefix is not preserved from the source. Also using one alias should be fine.

heshanpadmasiri: This should get fixed with https://github.com/ballerina-platform/ballerina-lang-go/issues/532

runtime/internal/exec/non_terminators.go
heshanpadmasiri: fixed

semantics/type_resolver.go
heshanpadmasiri: Created https://github.com/ballerina-platform/ballerina-lang-go/issues/533 to track. After fixing it this should just work

gimantha (dismissed):
LGTM

desugar/expression.go
```diff
@@ -482,6 +487,113 @@ func walkClientResourceAccessAction(cx *functionContext, expr *ast.BLangClientRe
...
+				b.WriteByte('"')
+			}
+			b.WriteString(part[insn.Offset:])
+			out[stringIndex] = b.String()
+		}
+	}
+	return out
+}
+
+func xmlTemplateStartTagInsertOffset(s string) int {
```
warunalakshitha: this function code coverage is low. shall we improve it.

semantics/xmlns_resolver.go
```diff
@@ -288,3 +288,69 @@ func mergeNamespaces(root *ast.BLangXMLElementLiteral, extras map[string]string)
...
+				if prefix == "" || prefix == model.XMLNSReservedPrefix {
+					continue
+				}
+				insn.Namespaces["xmlns:"+prefix] = uri
+			}
+		}
+	}
+}
+
+func xmlTemplatePrefixesForURI(scope model.Scope, uri string) map[string]string {
```
warunalakshitha: this seems unused. Shall we check

gimantha (approved):

PR #484, opened 2026-05-25
Add support for declaring service objects

heshanpadmasiri:
## Purpose
$subject
Resolves https://github.com/ballerina-platform/ballerina-lang-go/issues/464

## Summary by CodeRabbit

* **New Features**
  * Enhanced service & listener model (attach-point literal handling, service init/attach/start flow, multi-listener support) and stronger listener/service validation.
  * Runtime extern dispatch hooks for pluggable object/remote/resource method lookup and invocation.
  * Pretty-printer extensibility via a fallback hook for unsupported node types.

* **Tests**
  * Extensive new/updated corpus and integration tests covering listener lifecycle, attach/init failure diagnostics, multi-listener rules, and isolation/lock error messages.

* **Documentation**
  * Updated guidance for maintaining corpus golden files and test marker conventions.

PR #529, opened 2026-06-09
Implement Annotations and annotation access expressions for compile time and runtime

### LINKED ISSUE #528: [New Feature]: Implement Annotations for compile time and runtime and Annotation access expression
### Description

Refer to the Ballerina spec for annotations: https://ballerina.io/spec/lang/master/#section_9.1
Annotation Access expressions: https://ballerina.io/spec/lang/master/#section_6.13

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

gimantha:
Fixes https://github.com/ballerina-platform/ballerina-lang-go/issues/528

## Automation tests
- Unit tests
  > Code coverage information
- Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Full annotation support: declare, attach and read annotations on types, fields, constants, functions (including return types) with runtime access via typedescs.
  * Typedesc values now carry annotation metadata and can resolve runtime-referenced annotation values.
  * Constant evaluation extended to support richer compile-time constants (maps, lists, typedescs, operators, conversions, templates).

* **Tests**
  * Large suite of annotation and constant-evaluation fixtures and integration tests added/updated.


bir/bir_gen.go
```diff
@@ -466,6 +466,10 @@ func transformConstantAsGlobal(ctx *Context, c *ast.BLangConstant) BIRGlobalVari
 	dcl.Type = ctx.CompilerContext.SymbolType(c.Symbol())
 	dcl.Flags = c.Flags()
 	dcl.GlobalVarLookupKey = buildGlobalVarLookupKey(ctx.packageID, name)
+	if c.ConstantValueKnown {
```
heshanpadmasiri: Why would this be false at this point?

ast/expressions.go
```diff
@@ -750,6 +762,21 @@ func (b *BLangLiteral) SetOriginalValue(originalValue string) {
 	b.OriginalValue = originalValue
 }
 
+var sharedBooleanBType = &BTypeBasic{tag: TypeTags_BOOLEAN, flags: model.FlagReadonly}
+
+// NewBooleanLiteral creates a synthesized boolean literal with a resolved type.
+// Used for implicit marker annotation values.
+func NewBooleanLiteral(value bool, pos diagnostics.Location) *BLangLiteral {
```
heshanpadmasiri: Shall we make this a private function in the pacakge it uses it. It don't need any private APIs as far as I can see

ast/node_builder.go
```diff
@@ -2661,11 +2687,74 @@ func (n *NodeBuilder) TransformNilLiteral(nilLiteralNode *tree.NilLiteralNode) B
...
+	if !ok {
+		n.cx.SyntaxError("unknown annotation attach point '"+strings.Join(parts, " ")+"'", getPosition(n.de(), annotationAttachPointNode))
+	}
+	return AttachPoint{
+		Point:  point,
+		Source: annotationAttachPointNode.SourceKeyword() != nil,
+	}
+}
+
+var annotationAttachPointByName = map[string]Point{
```
heshanpadmasiri: I belive Point was something we just either mechanically (or using llms) migrated from java without much though put into. Maybe now we are using them it is worth it to think of a better way to represent this. I would personally use a single byte (should be plently enough given there is less than 20 mutually exclusive possible values; faster equality checks move effecient to move them around) to represent the point it self as some sort of enum. For string -> enum and enum -> string conversions I would wirte a giant switch statement with default panic (compiler is usually smart enough to reduce these to length checks fallowed by byte checks see https://godbolt.org/z/1W91sM73E)

semantics/symbol_resolver.go
```diff
@@ -961,7 +1015,7 @@ func referUserDefinedType[T symbolResolver](resolver T, n *ast.BLangUserDefinedT
 	if n.GetPackageAlias() != nil {
 		prefix = n.GetPackageAlias().GetValue()
 	}
-	resolveSymbolRef(resolver, name, prefix, n.GetPosition(), n)
+	resolveTypeRef(resolver, name, prefix, n.GetPosition(), n)
```
heshanpadmasiri: It is not clear to me why this change is needed. As far as I can see resolveTypeRef is identical to old `resolveSymbolRef`

semantics/type_resolver.go
```diff
@@ -853,6 +863,462 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+	cloneableMap := annotationMapType(t)
+	ld := semtypes.NewListDefinition()
+	cloneableMapList := ld.DefineListTypeWrappedWithEnvSemType(t.typeEnv(), cloneableMap)
+	return semtypes.IsSubtype(cx, ty, semtypes.BooleanConst(true)) ||
+		semtypes.IsSubtype(cx, ty, cloneableMap) ||
+		semtypes.IsSubtype(cx, ty, cloneableMapList)
+}
+
+func annotationMapType(t typeResolver) semtypes.SemType {
+	return semtypes.Intersect(semtypes.MAPPING, semtypes.CreateCloneable(t.typeContext()))
```
heshanpadmasiri, line 1062: Perhaps it is worth it to cache this in semtype context as well. I think if a package have annoations they will have lot of annotations

semantics/type_resolver.go
```diff
@@ -853,6 +863,462 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+func annotationHasSourceAttachPoint(annotation *ast.BLangAnnotation) bool {
+	for _, attachPoint := range annotation.AttachPoints() {
+		if attachPoint.Source {
+			return true
+		}
+	}
+	return false
+}
+
+func resolveTopLevelAnnotationAttachments(t typeResolver, pkg *ast.BLangPackage) {
```
heshanpadmasiri, line 1101: Is there any perticular reason why we can't resolve annotations while resolving the perticular ast node itself. I would imagine you will need to be able to resolve types out of order which should already work via ensureResolved. Not sure about constants at the moment but should be straight forward to do. If that is going to be very complicated I am fine with this with perhaps comment explaining why we are visiting the tree again

semantics/type_resolver.go
```diff
@@ -853,6 +863,462 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+	initialGlobalCount := len(pkg.GlobalVars)
```
heshanpadmasiri, line 1102: Equally it is not entierly clear to me why we need to do this in 2 phases one collecting annotations and another one evaluating. Perhaps it will become clear after fixing above

ast/ast.go
```diff
@@ -253,7 +260,9 @@ type (
 
 	BLangConstant struct {
 		BLangVariableBase
-		Name *BLangIdentifier
+		Name               *BLangIdentifier
+		ConstantValue      values.BalValue
+		ConstantValueKnown bool
 	}
```
heshanpadmasiri, line 235: I don't fully grok how the constant folding is implemented but as far as I can understand way it works for imports is via semtype and that's why you need this flag. (you can't recover the BalValue for non simple types using the semtype) and as far as I understand there is workaround in BIR gen for such values. 

The way I would have expected constant folding to work is at symbol level not type (you can't restor all possible values using the type) or AST (AST mutations are ephemeral so incremental compiler can't support them). We would extend ValueSymbol to a new ConstantValueSymbol that will have an addition `E value`.  What this `E` should be is not entierly clear to me but I would assume this needs to be something akin to how we store dependently typed functions return type. This E needs to represent a restricted set of BalValue in a way that is serializable. BalValue are recursive which makes it painful to serialize them but constant expressions are gurnteed to be non cyclic. 

Then in type resolver after resolving the types of constnat expressions (you need to do this after type resolution because resolving something like `const float f = 1;` needs to properly resolve the type of 1 based on the contextually expected type). You need to topologically sort all constant declaration in the module (All your dependencies would have been foldeded at this point and there can't be cycles). Then you can iterate them one by one a fold them (probably need something like `foldExpr(constExpression ast.ExpressionNode) E`). 

For `E` you will need a apply function (similar to TypeOp) that returns an ExpressionNode. Then in desugar (not sure what needs to be done for annotations) you can replace all references to const symbol with that expression. Thus this should require no changes to BIRGen but after this change you can drop adding constants to BIRPackage entirely and all code should continue to work. 

Note this is restrictive in that folding stops the moment we are not dealing with constants. For example
```ballerina
var a = C1 + C2
```
will not be folded even if C1 and C2 are constants where as 
```ballerina
const a = C1 + C2
``` 
will be folded. But I think that is sufficient for our use case

heshanpadmasiri (changes requested):

warunalakshitha:
With new Service PR , service annotations are not covered. https://github.com/ballerina-platform/ballerina-lang-go/pull/484/changes?file-filters%5B%5D=.go#diff-9e02e2f1085b1a0ba57b7a1248d018dc99894d620c2af509437143a222b99892R1643-R1644

bir/bir_gen.go
gimantha: Fixed

ast/expressions.go
gimantha: fixed

semantics/type_resolver.go
gimantha: Fixed in type_resolver.go:915-926

semantics/symbol_resolver.go
gimantha: fixed

semantics/type_resolver.go
gimantha: This is due to parallel evaluation of constant expressions. If there are modules with lots of annotations, I have made constant expressions evaluation parallel

semantics/type_resolver.go
gimantha: This is due to parallel evaluation of constant expressions. If there are modules with lots of annotations, I have made constant expressions evaluation parallel

ast/node_builder.go
gimantha: Fixed now

ast/ast.go
gimantha: Constant Symbol is implemented and removed from the BIR now

ast/ast.go
```diff
@@ -1011,6 +1063,44 @@ func (b *bLangInvokableNodeBase) SetMarkdownDocumentationAttachment(markdownDocu
...
 }
 
+// ReturnTypeAnnotatable wraps the return-type annotation slice of an invokable
+// node and exposes it as an AnnotatableNode so the type resolver can process
+// return-type annotations uniformly.
+type ReturnTypeAnnotatable struct {
```
heshanpadmasiri: NIT: I assume this is used to represent annotations for the return type descriptor of function declaration. Personally I don't like idea of giving views/adaptor pattern when it comes to AST, it sort of makes ASTs behaviour dynamic/context dependent where as I would personally like to be able to treat the AST as pure data as much as possible. Would it be better to simply have ReturnTypeDesc Node where you can store both TypeDescriptor and it's annotations. So it can be treated both as an annotatable node and a type descriptor node

bir/bir_gen.go
```diff
@@ -386,6 +386,13 @@ func GenBir(ctx *compilerctx.CompilerContext, ast *ast.BLangPackage) *BIRPackage
 		addGlobalVar(birPkg, TransformGlobalVariableDcl(genCtx, &globalVar))
 	}
 	for _, constant := range ast.Constants {
+		// Constants folded to a serializable value are inlined at their use
```
heshanpadmasiri: Spec says "A const-expr is evaluated at compile-time"https://ballerina.io/spec/lang/master/#section_6.4, and I don't necessary see a reason why cast (given the reference it self is const expression) can't be done in front end with folding. We it can't be cast then that can be a semantic error saying $ can't be T

bir/bir_gen.go
```diff
@@ -1024,6 +1044,8 @@ func handleActionOrExpression(ctx context, curBB *BIRBasicBlock, expr ast.BLangA
 		return binaryExpression(ctx, curBB, expr)
 	case *ast.BLangSimpleVarRef:
 		return simpleVariableReference(ctx, curBB, expr)
+	case *ast.BLangConstRef:
+		return simpleVariableReference(ctx, curBB, &expr.BLangSimpleVarRef)
```
heshanpadmasiri: If we have folded these away shouldn't we replace all references to constants with the value it self. (not sure if this should have been done in desugar or this should be done in BIRGen; I prefer desugar)

bir/model.go
```diff
@@ -119,6 +119,8 @@ type (
 		Flags              model.Flag
 		PkgId              *model.PackageID
 		GlobalVarLookupKey string
+		InitialValue       any
+		HasInitialValue    bool
```
heshanpadmasiri: Not sure what this boolean means all module level variables must be initialized?

bir/model.go
```diff
@@ -119,6 +119,8 @@ type (
 		Flags              model.Flag
 		PkgId              *model.PackageID
 		GlobalVarLookupKey string
+		InitialValue       any
```
heshanpadmasiri:
```suggestion
		InitialValue       values.BValue
```

model/symbol.go
```diff
@@ -66,6 +67,20 @@ type Symbol interface {
 	Copy() Symbol
 }
 
+// ValueSymbolView is satisfied by both *ValueSymbol and *ConstantValueSymbol.
+// Callers that only care about a symbol's value-symbol facets (whether it is a
+// constant, parameter, final, etc.) should type-assert against this interface
+// rather than the concrete *ValueSymbol, so that constants — which are stored as
+// *ConstantValueSymbol — are matched too.
+type ValueSymbolView interface {
```
heshanpadmasiri: nit: I would rather have ValueSymbol called VariableSymbol and this ValueSymbol (view gives a different meaning IMHO)

model/symbol.go
```diff
@@ -285,6 +310,22 @@ type (
...
+	// Folding lives at the symbol level rather than on the type (a type cannot
+	// recover every possible value, e.g. compound shapes) or on the AST (AST
+	// mutations are ephemeral and would not survive incremental compilation).
+	// The value is materialized back into an expression node by the desugar
+	// package (the "apply", analogous to TypeOp.Apply) — that step cannot live
+	// here because model must not import ast.
+	ConstantValueSymbol struct {
+		ValueSymbol
+		value      values.BalValue
+		valueKnown bool
```
heshanpadmasiri: I assume this would dissapear when we fix constant folding for casts

model/symbol.go
```diff
@@ -285,6 +310,22 @@ type (
 		flags valueSymbolFlags
 	}
 
+	// ConstantValueSymbol is a ValueSymbol for a module constant whose
```
heshanpadmasiri: I think we can clean up this comment

runtime/internal/modules/module.go
```diff
@@ -32,10 +32,30 @@ type ExternFunction struct {
 	Impl extern.NativeFunc
 }
 
-func NewBIRModule(ctx *extern.Context, pkg *bir.BIRPackage) *BIRModule {
+func NewBIRModule(ctx *extern.Context, pkg *bir.BIRPackage) (*BIRModule, error) {
 	globals := make(map[string]values.BalValue, len(pkg.GlobalVars))
+	for key, gv := range pkg.GlobalVars {
```
heshanpadmasiri, line 38: This doesn't looks okay to be but I assume after fixing the constant type cast this would no longer be needed anyway

model/symbol.go
```diff
@@ -241,6 +259,13 @@ type (
 
 	TypeSymbol struct {
 		symbolBase
+		annotations values.AnnotationValues
```
heshanpadmasiri: Are we adding this to TypeSymbols because currently we only care about Type annotations (I assume others like FunctionSymbols) may need annotations as well. Either way I would much prefer annotations are kept at compiler env. Basically you store them by symbol ref and get things by symbol ref. This way it is much efficient to lookup annotations by reference and given most symbols may not have annotations we are not wasting a word in each struct. 

I agree when serializing it would be better to store them together but than can be done in serializer/deserilizer

semantics/type_resolver.go
```diff
@@ -882,6 +908,514 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+	cx := t.typeContext()
+	cloneableMap := annotationMapType(t)
+	cloneableMapList := annotationMapListType(t)
+	return semtypes.IsSubtype(cx, ty, semtypes.BooleanConst(true)) ||
+		semtypes.IsSubtype(cx, ty, cloneableMap) ||
+		semtypes.IsSubtype(cx, ty, cloneableMapList)
+}
+
+func annotationMapType(t typeResolver) semtypes.SemType {
+	if resolver := packageResolverOf(t); resolver != nil {
```
heshanpadmasiri: It is bit sucpicious to me that you are handling a case where there is no pkgResolver. Also I am not sure writing to pkgResolver like this is thread safe. when resolving local nodes I belive we run multiple type resolvers concurrently from the same pkgResolver

semantics/type_resolver.go
```diff
@@ -882,6 +908,514 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+				semtypes.CreateCloneable(resolver.typeContext()),
+			)
+		}
+		return resolver.annotationMapType
+	}
+	return semtypes.Intersect(semtypes.MAPPING, semtypes.CreateCloneable(t.typeContext()))
+}
+
+func annotationMapListType(t typeResolver) semtypes.SemType {
+	if resolver := packageResolverOf(t); resolver != nil {
```
heshanpadmasiri: same as above

semantics/type_resolver.go
```diff
@@ -882,6 +908,514 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+}
+
+func packageResolverOf(t typeResolver) *packageTypeResolver {
+	for t != nil {
+		if resolver, ok := t.(*packageTypeResolver); ok {
+			return resolver
+		}
+		t = t.parent()
+	}
+	return nil
```
heshanpadmasiri: It is not clear to me why this case needs to be supported

semantics/type_resolver.go
heshanpadmasiri: In retrospect I am not entierly convinced that it is a good call. Currently things we evaluate in parallel (such as function nodes after resolving top level nodes), language it self gurantee we can handle them in parallel without a need for synchronization (ignoring some of our own mistakes like how type narrowing worked previously). But language makes no such gurantee about constant expressions other than they are acyclic. Looking at the complexity we have taken in here and in constant_evaluator perhaps it's not worth it. It would be much simpler and much easier to reason about if we simply toplogically sort all the module constant declarations and fold then in that order, we would no longer needs any of these caches and other stuff.

semantics/semantic_analyzer.go
```diff
@@ -984,6 +986,11 @@ func analyzeActionOrExpression[A analyzer](a A, expr ast.BLangActionOrExpression
 		return validateResolvedType(a, expr, expectedType)
 	case *ast.BLangTypedescExpr:
 		return validateResolvedType(a, expr, expectedType)
+	case *ast.BLangAnnotAccessExpr:
+		if !analyzeActionOrExpression(a, expr.Expr, semtypes.SemType{}) {
```
heshanpadmasiri: Isn't expected type `typedesc`

semantics/type_resolver.go
gimantha: this is sequentially evaluated now

.claude/settings.local.json
warunalakshitha, line 1: Do we need this file?

coverage_report.html
warunalakshitha, line 1: We can remove this too

runtime/internal/modules/module.go
warunalakshitha: Shall we create an issue to track that

values/values.go
```diff
@@ -26,6 +26,39 @@ import (
 // Currently this is just an alias on any but I think we will need to add methods to this like type
 type BalValue any
 
+// AnnotationValue is the evaluated Ballerina value stored for an annotation.
+type AnnotationValue = BalValue
```
warunalakshitha: Shall we move annotation code to separate file ex. annotations.go

values/values.go
```diff
@@ -26,6 +26,39 @@ import (
...
+	return r.Organization + "/" + r.Module + ":" + r.GlobalName
+}
+
+// NewAnnotationValues returns an initialized annotation value map.
+func NewAnnotationValues() AnnotationValues {
+	return make(AnnotationValues)
+}
+
+// Clone returns a shallow copy of the annotation value map.
+func (values AnnotationValues) Clone() AnnotationValues {
```
warunalakshitha: Why we need clone map?

values/values.go
gimantha: fixed

values/values.go
gimantha: this is removed now

ast/ast.go
```diff
@@ -532,6 +562,27 @@ func (b *BLangAnnotation) SetTypeDescriptor(typeDescriptor TypeDescriptor) {
...
+	for attachPoint := range b.attachPoints.Values() {
+		result = append(result, attachPoint)
+	}
+	sort.Slice(result, func(i, j int) bool {
+		if result[i].Point != result[j].Point {
+			// Sort by the canonical key (alphabetical) to keep ordering stable
+			// across the byte-valued Point enum.
+			return result[i].Point.String() < result[j].Point.String()
+		}
+		return !result[i].Source && result[j].Source
```
heshanpadmasiri: Why do we need to keep this order stable? If this is to handle pretty printers then they should sort results at their end

ast/ast.go
```diff
@@ -1012,6 +1063,48 @@ func (b *bLangInvokableNodeBase) SetMarkdownDocumentationAttachment(markdownDocu
 	}
 }
 
+// BLangReturnTypeDescriptor is the return type descriptor of an invokable node.
+// It holds the return type and its annotation attachments as real AST data, so a
+// single node serves as both a type-descriptor holder and an AnnotatableNode.
+// This replaces the former ReturnTypeAnnotatable view, keeping the AST pure data
+// rather than synthesizing an annotatable adaptor on demand.
```
heshanpadmasiri: I think this comment is wrong (it captures the history of PR not the codebase)

ast/ast.go
```diff
@@ -1012,6 +1063,48 @@ func (b *bLangInvokableNodeBase) SetMarkdownDocumentationAttachment(markdownDocu
...
+	for i := range r.AnnAttachments {
+		result[i] = &r.AnnAttachments[i]
+	}
+	return result
+}
+
+// ReturnTypeAnnotatableOf returns the invokable node's return type descriptor as
+// an AnnotatableNode (it carries the return type's annotation attachments), or
+// nil if the node has no return type descriptor.
+func ReturnTypeAnnotatableOf(fn InvokableNode) AnnotatableNode {
```
heshanpadmasiri: This makes me feel that we haven't go the BLangReturnTypeDescriptor type correctly. It should implement both AnnotatbleNode and TypeDescriptor (I assume also BType). In my mind it should be straightforward as long as we maintain the invariant BLangReturnTypeDescriptor can't have a nil TypeDescriptor node. Then InvokableNode.returnTypeDescriptor can return a suitable interface (Say ReturnTypeDescriptor) that satisy both TypeDescriptor and AnnotatableNode interfaces.

bir/bir_gen.go
```diff
@@ -1724,6 +1740,8 @@ func simpleVariableReference(ctx context, curBB *BIRBasicBlock, expr *ast.BLangS
 			block:  curBB,
 		}
 	}
+	// Folded constants are inlined during desugar, so a constant reference that
+	// reaches BIR is a non-foldable constant lowered to a global variable.
```
heshanpadmasiri: This is no longer true?

desugar/constants.go
```diff
@@ -0,0 +1,103 @@
...
+// (the E in the design), realized here because model cannot import ast. Returns
+// nil for non-foldable constants (e.g. casts that panic at runtime), which keep
+// flowing through BIR as global variables.
+func materializeConstantRef(cx *functionContext, ref *ast.BLangSimpleVarRef) ast.BLangExpression {
+	constSym, ok := cx.getSymbol(ref.Symbol()).(*model.ConstantValueSymbol)
+	if !ok {
+		return nil
+	}
+	value, ok := constSym.ConstantValue()
+	if !ok || !values.IsSerializableConstValue(value) {
```
heshanpadmasiri: Can it be non serializable? (I am under the impression all possible constant values are acyclic thus always serializable)

desugar/constants.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil
+	}
+	return constantValueLiteral(value, ref.GetPosition(), ref.GetDeterminedType())
+}
+
+func constantValueLiteral(value values.BalValue, pos diagnostics.Location, ty semtypes.SemType) ast.BLangExpression {
+	if semtypes.IsZero(ty) {
```
heshanpadmasiri, line 37: I am bit suspicious that ty could be zero. given you get the type form a symbol ref, I would have assumed all of them to have types by the type we pass type resolution

desugar/constants.go
```diff
@@ -0,0 +1,103 @@
...
+	case float64, float32:
+		return ast.TypeTags_FLOAT, true
+	case *decimal.Decimal:
+		return ast.TypeTags_DECIMAL, true
+	case string, *string:
+		return ast.TypeTags_STRING, true
+	default:
+		return 0, false
+	}
+}
```
heshanpadmasiri: If seems to be we are switching on the type of the value multiple times. I think it is better to have a single function that gives us all this information at the same time

desugar/corpus_desugar_test.go
```diff
@@ -239,17 +239,39 @@ func canonicaliseInitFnBodies(e *sexp) {
...
+		rightIsImport := isImportPackageSExp(right)
+		if leftIsImport != rightIsImport {
+			return leftIsImport
+		}
+		if !leftIsImport {
+			return false
+		}
+		return printSExp(left) < printSExp(right)
+	})
+}
```
heshanpadmasiri: I assume this is fixed in main now please check

desugar/desugar.go
```diff
@@ -362,16 +356,38 @@ func (v *dependencyVisitor) Visit(node ast.BLangNode) ast.Visitor {
...
 	return v
 }
 
 func (v *dependencyVisitor) VisitTypeData(_ *ast.TypeData) ast.Visitor { return v }
 
+func (v *dependencyVisitor) dependsOnRuntimeAnnotation(expr *ast.BLangAnnotAccessExpr) {
+	receiver, ok := expr.Expr.(ast.BNodeWithSymbol)
+	if !ok {
+		return
+	}
```
heshanpadmasiri, line 371: Shouldn't this be always false? I generally prefer to make them internal errors in that case, so we know if suddenly our assumptions became false rather than silently ignoring them

desugar/desugar.go
```diff
@@ -504,6 +521,7 @@ func desugarInitFn(pkgCtx *packageContext, compilerCtx *context.CompilerContext,
 		}
 	}
 	if !needInit {
+		clearModuleInitExprs(pkg)
```
heshanpadmasiri: I am somewhat suspicious why this change is needed. It seems it is added purely to satisfy corpus tests. I think the correct changes is to simply set pkg.Constants = nil after desugaring all the constants refs

desugar/statement.go
```diff
@@ -1008,6 +1008,19 @@ func walkMatchStatement(cx *functionContext, stmt *ast.BLangMatchStatement) desu
 
 	for i := range stmt.MatchClauses {
 		clause := &stmt.MatchClauses[i]
+		// Const-pattern expressions are lowered directly by BIR generation, so
+		// they must be walked here too — otherwise a reference to a folded
+		// constant would survive as a dangling global once the constant is
+		// dropped from the BIR package.
+		for _, pattern := range clause.Patterns {
+			constPattern, ok := pattern.(*ast.BLangConstPattern)
+			if !ok || constPattern.Expr == nil {
```
heshanpadmasiri: Not sure how Expr could be nil

model/symbol.go
```diff
@@ -732,6 +811,64 @@ func (ts *TypeSymbol) Copy() Symbol {
 	panic("TypeSymbol cannot be copied")
 }
 
+func (as *AnnotationSymbol) Kind() SymbolKind {
+	return SymbolKindAnnotation
+}
+
+func (as *AnnotationSymbol) Copy() Symbol {
+	cp := *as
```
heshanpadmasiri: I don't think these symbols need to support Copy. We use copy to support type narrowing a symbol if I recall

model/symbol.go
```diff
@@ -732,6 +811,64 @@ func (ts *TypeSymbol) Copy() Symbol {
...
+	if as.attachPoints[point] {
+		return true
+	}
+	if point == "recordfield" || point == "objectfield" {
+		if as.attachPoints["field"] || as.attachPoints[SourceAnnotationAttachPointKey("field")] {
+			return true
+		}
+	}
+	return as.attachPoints[SourceAnnotationAttachPointKey(point)]
+}
```
heshanpadmasiri, line 867: Isn't the possible set of AttachPoints a fixed set. If so I would rather use a bit set probably a u32 or 64 should be enough

runtime/internal/modules/module.go
heshanpadmasiri: I am under the impression we are not going to special case type casts with folding anymore in the PR

semantics/constant_evaluator.go
```diff
@@ -0,0 +1,659 @@
...
+}
+
+type constantEvaluationCache struct {
+	results map[model.SymbolRef]constantEvaluationResult
+}
+
+func newConstantEvaluationCache() *constantEvaluationCache {
+	return &constantEvaluationCache{
+		results: make(map[model.SymbolRef]constantEvaluationResult),
+	}
```
heshanpadmasiri: Do we still need this? I though we are going to toplogically sorting all the contant expression and fold them in that order in which case we can directly get the value from the symbol

heshanpadmasiri: It seems we are trying to do both sorting and folding in the same pass, I feel that breaking them up would make the code much simpler and you don't need to maintain much of the state

semantics/constant_evaluator.go
```diff
@@ -0,0 +1,659 @@
...
+	default:
+		return nil, fmt.Errorf("unsupported constant conversion from %T", value)
+	}
+	if err != nil {
+		return nil, err
+	}
+	if !semtypes.IsSubtype(e.resolver.typeContext(), values.SemTypeForValue(converted), targetType) {
+		return nil, fmt.Errorf("converted constant does not belong to target type")
+	}
+	return converted, nil
```
heshanpadmasiri, line 494: Given this is common for both runtime and here I wonder would it be better to move the logic to values and reuse it in both places

semantics/type_resolver.go
```diff
@@ -893,6 +920,438 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+	return true
+}
+
+func annotationHasSourceAttachPoint(annotation *ast.BLangAnnotation) bool {
+	for _, attachPoint := range annotation.AttachPoints() {
+		if attachPoint.Source {
+			return true
+		}
+	}
+	return false
```
heshanpadmasiri: I wonder would it be better to give a method to directly check if an annotation have a given source. In the future that should allow us to make the actual representation of attachpoints and their lookup more efficient if we have to deal with lot of attachpoints

semantics/type_resolver.go
```diff
@@ -893,6 +920,438 @@ func (t *packageTypeResolver) resolveTopLevelTypes(pkg *ast.BLangPackage) {
...
+}
+
+func resolveTopLevelAnnotationAttachments(t typeResolver, pkg *ast.BLangPackage) {
+	initialGlobalCount := len(pkg.GlobalVars)
+	var tasks []annotationEvaluationTask
```
heshanpadmasiri: Given we are now resolving constants sequentially do we still need this? I would have assumed after that we can eval annotations as part of the normal type resolution of the given node

ast/ast.go
gimantha: fixed

ast/ast.go
gimantha: Changed the comment

ast/ast.go
gimantha: Fixed. ReturnTypeDescriptor interface now includes `BType` and `AnnotatableNode`

bir/bir_gen.go
gimantha: fixed

desugar/constants.go
gimantha: Removed  that check

desugar/constants.go
gimantha: This check is now removed

desugar/constants.go
gimantha: Fxied by reducing to 1 function

desugar/corpus_desugar_test.go
gimantha: yes. removed

desugar/desugar.go
gimantha: Added an internal error

desugar/desugar.go
gimantha: Removed this. We are later setting pkg.Constants to nil

desugar/statement.go
gimantha: nil-check removed

model/symbol.go
gimantha: changed to bitset u64

semantics/constant_evaluator.go
gimantha: This is not needed. Removed

semantics/constant_evaluator.go
gimantha: Fixed. Now reusing the values casting

semantics/type_resolver.go
gimantha: ast.go has a function called `HasSourceAttachPoint` now

semantics/type_resolver.go
gimantha: removed

heshanpadmasiri (dismissed):
LGTM

corpus/cli_integration_test.go
```diff
@@ -517,6 +520,34 @@ func runCLICommand(t *testing.T, balBin, repoRoot, coverDir string, args ...stri
 	return "", "", 0
 }
 
+func mergeCLICoverageDir(t *testing.T, fromDir, toDir string) {
```
warunalakshitha, line 523: Do we need this merge coverage now?

warunalakshitha: Resolving as per the offline discusstion.

warunalakshitha (approved):
LGTM

PR #525, opened 2026-06-04
Add Level 1 standard library support - Phase 1

TharmiganK:
## Summary

Closes #479.
Closes #515

Adds Go-native implementations of the Level 1 Ballerina standard libraries as embedded bala bundles (Phase 1).

### Libraries Added / Extended

| Package | Supported | Partially Supported | Not Yet Supported | Support % |
|---|---|---|---|---|
| `ballerina/http` | 24 | 2 | 46 | 33% |
| `ballerina/io` | 14 | 1 | 11 | 54% |
| `ballerina/math.vector` | 5 | 0 | 0 | 100% |
| `ballerina/time` | 31 | 1 | 0 | 97% |
| `ballerina/url` | 3 | 0 | 1 | 75% |
| **Total** | **77** | **4** | **58** | **55%** |

#### `ballerina/http`
HTTP client subset only (no listener/service side). Supports all nine remote methods (`get`, `post`, `put`, `patch`, `delete`, `head`, `options`, `execute`, `forward`), TLS/mTLS, redirect following, connection pooling, compression negotiation, proxy configuration, and response limit configuration.

#### `ballerina/io`
Console output (`print`, `println`, `fprint`, `fprintln` to stdout/stderr) plus full file I/O: `fileRead/WriteString`, `fileRead/WriteLines`, `fileRead/WriteBytes`, `fileRead/WriteJson`, `fileRead/WriteXml`, and the `FileWriteOption` enum.

#### `ballerina/math.vector`
100% coverage — all five vector functions (`fromArray`, `toArray`, `add`, `subtract`, `dot`) implemented in pure Ballerina.

#### `ballerina/time`
Near-complete: `Utc`, `Civil`, `Duration`, `Zone`, `TimeZone`, all conversion functions (`utcNow`, `utcFromCivil`, `civilFromUtc`, `utcToString`, `civilToString`, etc.), monotonic clock, IANA zone loading (`loadSystemZone`, `getZone`), and timezone fixed-offset construction.

#### `ballerina/url`
`encode`, `decode`, and `encodeUserInfo`/`decodeUserInfo`; multi-charset support (UTF-8, ISO-8859-1, UTF-16BE).

### Implementation Details

- **PAL (Platform Abstraction Layer):** HTTP and IO native operations are wired through `platform/pal` so the interpreter can swap implementations (e.g. test doubles).
- **Module resolver:** `projects/module_resolver.go` updated to resolve the new embedded bala packages.
[...]

PR #548, opened 2026-06-19
Add Level 2 standard library support - Phase 1

### LINKED ISSUE #480: Add level 2 standard libraries with basic support
### Description

Standard libraries:
- [ ] ~avro~
- [ ] ~constraint~
- [x] crypto
- [x] log
- [x] os
- [ ] ~protobuf~
- [x] random
- [ ] ~xslt~

### Describe your problem(s)

Level 2 standard libraries are not supported

### Describe your solution(s)

Add support for Level 2 standard libraries

### Related area

Other Area

_No response_

_No response_

_No response_

TharmiganK:
## Summary

Closes https://github.com/ballerina-platform/ballerina-lang-go/issues/480

Depends on https://github.com/ballerina-platform/ballerina-lang-go/pull/525

Adds Go-native implementations of the Level 2 Ballerina standard libraries as embedded bala bundles (Phase 2).

### Libraries Added

| Package | Supported | Partially Supported | Not Yet Supported | Support % |
|---|---|---|---|---|
| `ballerina/crypto` | 26 | 1 | 5 | 81% |
| `ballerina/log` | 7 | 2 | 15 | 29% |
| `ballerina/os` | 11 | 1 | 0 | 92% |
| `ballerina/random` | 3 | 1 | 1 | 60% |
| **New in this PR** | **47** | **5** | **21** | **72%** |

#### `ballerina/crypto`
Core cryptographic operations. Covers all standard hash functions (MD5, SHA-1/256/384/512, Keccak-256, CRC32B), all five HMAC variants, three password-hashing schemes (BCrypt, Argon2id, PBKDF2), symmetric AES encryption (CBC, ECB, GCM), asymmetric RSA and ECDSA sign/verify, PKCS12 keystore/truststore loading, PEM key loading with encrypted-key support, HKDF-SHA256 key derivation, and constant-time hash comparison. Not yet supported: post-quantum primitives (ML-KEM, ML-DSA, HPKE) and PGP operations.

#### `ballerina/log`
Structured LOGFMT logging to stderr at four severity levels (`printDebug`, `printInfo`, `printWarn`, `printError`). Supports optional `'error` parameter and arbitrary key-value pair annotations via rest-record syntax. Level filtering is hardcoded to `INFO`. Not yet supported: configurable levels/formats, JSON output, file destinations, log rotation, the `Logger` object API, and sensitive-data masking.

#### `ballerina/os`
Full surface of the jBallerina `os` module. Environment variable management (`getEnv`, `setEnv`, `unsetEnv`, `listEnv`), current-user queries (`getUsername`, `getUserHome`), and subprocess execution (`exec` with the `Process` object: `waitForExit`, `output`, `exit`). Validates that environment property keys cannot shadow the `command` field via a `never command?;` exclusion guard.

#### `ballerina/random`
`createDecimal()` for a cryptographically secure random `float` in [0.0, 1.0) and `createIntInRange(startRange, endRange)` for a uniform random integer in [startRange, endRange). Returns a typed `random:Error` on invalid range arguments.

### Implementation Details

- **PAL (Platform Abstraction Layer):** OS operations (environment variables, user info, subprocess execution) are wired through `platform/pal` via a new `OS` struct and a `ProcessHandle` interface, keeping the interpreter swappable for test doubles.
- **New dependency:** `golang.org/x/crypto` added for Keccak-256 (`golang.org/x/crypto/sha3`), Argon2id (`golang.org/x/crypto/argon2`), BCrypt (`golang.org/x/crypto/bcrypt`), and HKDF (`golang.org/x/crypto/hkdf`).
- **Module registration:** The four new packages are registered in `lib/rt/libs.go` alongside the Level 1 packages.
- **Corpus tests:** Golden-file corpus tests added for all new functions under `corpus/bal/`, `corpus/ast/`, `corpus/bir/`, `corpus/cfg/`, `corpus/desugared/`, and `corpus/integration/`.
- **Per-package READMEs:** Each library includes a `README.md` under `lib/stdlibs/ballerina/<name>/0.0.1/go1.2/` with a full feature-by-feature support table, behavioural notes, and usage examples. The consolidated summary at `lib/stdlibs/ballerina/README.md` is updated.
- **Agent skills refined:** `add-stdlib-support` and `fill-stdlib-gap` skills under `.agents/skills/` updated with lessons learned from the Level 2 migration.

### Notable Behavioural Differences from jBallerina

- **`crypto`:** AES-CBC and AES-ECB always apply PKCS7 padding regardless of the `padding` parameter — Go's `cipher` package exposes no no-padding mode. Programs relying on `NONE` padding will produce incorrect output.
- **`log`:** Module name is always empty (`module=""`) because Go has no equivalent to the JVM `StackWalker`. Error values are formatted as `error("message")` rather than a full `FullErrorDetails` record.
- **`os`:** Environment mutations (`setEnv`, `unsetEnv`) call `os.Setenv`/`os.Unsetenv` directly and are process-wide, not strand-isolated as in jBallerina. Safe for single-threaded programs.
- **`random`:** `createDecimal()` reads 53 bits from `crypto/rand`, producing a full-precision IEEE 754 `float64` (vs. jBallerina's 24-bit `SecureRandom.nextFloat()` widened to `float64`). `createIntInRange()` fixes an off-by-one bug present in the original jBallerina implementation — the Go version correctly covers the full `[startRange, endRange)` range.

## Test Plan

- [x] Run the full corpus test suite: `go test ./...`
- [x] Verify corpus golden files for log, os, random, and crypto pass
- [x] Confirm `ballerina/os` subprocess tests pass (`exec`, `waitForExit`, `output`)
- [x] Verify `ballerina/crypto` hash / HMAC / AES-GCM / RSA / ECDSA round-trip tests pass
- [x] Confirm key-value pair annotations are emitted correctly in `ballerina/log` output (`log-kv1-v.txtar`)
- [x] Check that `ballerina/random` range-error path returns a typed `random:Error` (`random-range1-v`)

## Summary by CodeRabbit

* **New Features**
  * Added/extended Subset 2 corpus programs and integration fixtures covering crypto, logging, OS environment/process utilities, random generation, and HTTP request/response.
  * Introduced new Subset 2 documentation for supported module capabilities and HTTP/crypto behaviors.
* **Bug Fixes**
  * Improved `-v` stderr assertions to tolerate non-diagnostic stderr while still failing on compiler diagnostics.
  * Updated float `round()` behavior to ties-to-even semantics.
  * Enhanced HTTP handling with lazy gzip decompression and expanded HTTP client tests (HTTP/1.1, header forwarding, malformed gzip).
* **Documentation**
  * Updated skill guides for stdlib porting workflows and cross-stdlib `Dependencies.toml` requirements.


lib/langlibs/go/lang.float/float.go
```diff
@@ -46,7 +46,8 @@ func initFloatModule(rt *runtime.Runtime) {
 		return math.Floor(args[0].(float64)), nil
 	})
 	reg("round", func(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
-		return math.Round(args[0].(float64)), nil
+		// Ballerina float:round is IEEE roundToIntegralTiesToEven, not half-away-from-zero.
+		return math.RoundToEven(args[0].(float64)), nil
```
warunalakshitha, line 50: Shall we write a test for this in language subset.

corpus/extern/testdata/crypto-keystore-v.bal
```diff
@@ -0,0 +1,43 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/crypto;
+import ballerina/io;
+
+public function main() returns error? {
+    crypto:KeyStore ks = {path: "testdata/crypto-keystore.p12", password: "secret"};
```
warunalakshitha, line 21: Why we cannot add these extern tests to corpus/bal/library

warunalakshitha: same with http client tests too

corpus/assertions_test.go
```diff
@@ -88,8 +88,10 @@ func hasAnySuffix(name string, suffixes ...string) bool {
 func assertOutputAnnotations(t *testing.T, anns annotations, stdout, stderr string) {
 	t.Helper()
 
-	if strings.TrimRight(stderr, "\n") != "" {
-		t.Errorf("expected empty stderr for -v test, got:\n%s", stderr)
+	// A -v test may write to stderr intentionally (e.g. ballerina/log records),
```
warunalakshitha, line 91: i hope this is with log:error not due to bug

lib/langlibs/go/lang.float/float.go
warunalakshitha: Do we need specific comment since this is a bug and other place we do not have a comment

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/helpers.go
```diff
@@ -0,0 +1,67 @@
...
+		items[i] = int64(b)
+	}
+	return values.NewList(byteArrTy, semtypes.ToListAtomicType(ctx.TypeCtx, byteArrTy), false, nil, 0, items)
+}
+
+// mapString reads a string field from a Ballerina Map.
+func mapString(m *values.Map, key string) string {
+	v, ok := m.Get(key)
+	if !ok {
+		return ""
```
warunalakshitha: is returning "" correct? can we confirm this path never happen

values/map.go
```diff
@@ -43,6 +43,18 @@ type Map struct {
 
 	data       map[string]*mapEntry
 	head, tail *mapEntry
+
+	nativeData any
+}
+
+// SetNativeData stores an opaque Go value alongside the map for use by native externs.
```
warunalakshitha: @heshanpadmasiri I hope this should be fine

lib/stdlibs/ballerina/log/0.0.1/go1.2/native/log.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+const (
+	orgName    = "ballerina"
+	moduleName = "log"
+)
+
+// logLevelWeight maps level name → numeric weight for filtering.
+// Higher weight = higher severity (matches jBallerina's logLevelWeight map).
+var logLevelWeight = map[string]int{
```
warunalakshitha: Is these values are required? Can we do something like this

const (
	Debug LogLevel = iota
	Info
	Warn
	Error
)

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/keys.go
```diff
@@ -0,0 +1,344 @@
...
+			if err != nil {
+				// Try with key password if store password failed
+				key, _, err = pkcs12.Decode(data, keyPwd)
+				if err != nil {
+					return cryptoError(fmt.Sprintf("Key cannot be recovered by using given key alias: %s", alias)), nil
+				}
+			}
+			rsaKey, ok := key.(*rsa.PrivateKey)
+			if !ok {
+				return cryptoError("Not a valid RSA key"), nil
```
warunalakshitha, line 58: Can we write some crypto error cases using bal files. If these are just defensive code we can keep without tests.

platform/palnative/pal.go
```diff
@@ -60,6 +63,44 @@ func NewPlatform() (pal.Platform, func()) {
 				return err
 			},
 		},
+		OS: pal.OS{
```
warunalakshitha, line 66: These are not covered as per https://github.com/ballerina-platform/ballerina-lang-go/pull/548#issuecomment-4748374213

Since this is production pal impl can we write some cli_integration tests which run with executable

lib/langlibs/go/lang.float/float.go
TharmiganK: There is already a test for this:  corpus/bal/subset8/08-float/29-v.bal
Also this is not a bug, as per the Ballerina specification, float:round is IEEE roundToIntegralTiesToEven, so the correct Go call is math.RoundToEven, not the more intuitive math.Round (half-away-from-zero). Just added that as a comment to stop any swapping with math.Round in the future

Reference: https://github.com/ballerina-platform/ballerina-lang-go/pull/585#discussion_r3497952995

corpus/extern/testdata/crypto-keystore-v.bal
TharmiganK: The corpus/bal/library harness runs .bal files under `testharness.NewTestPal()`, whose:
- HTTP client is stubHTTP — it returns a dummy 200 / "test body" for every request (test_harness.go:187). That's why http-client-compression-v.bal can live in subset2: it only checks that client config parses and a call returns 200. It cannot assert real status codes, headers, bodies, redirects, TLS, or connection errors.
- FS is the real os.ReadFile/WriteFile, so tests can materialize text fixtures (e.g. crypto-keys-v.bal writes PEM strings to /tmp and reads them back).

corpus/assertions_test.go
TharmiganK: Not a bug, and not limited to `log:error`. `ballerina/log` writes all enabled levels (INFO/WARN/ERROR by default; DEBUG filtered) to **stderr**, matching jBallerina. So any log-*-v.bal case legitimately emits stderr.

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/helpers.go
TharmiganK: Confirmed unreachable for the current callers. `mapString` is only used on `crypto:KeyStore/crypto:TrustStore`, both closed records with required string `path` and string `password`. Required fields on a closed record are guaranteed present, so the `""` branch never fires — it's purely a defensive default. I can add a one-line comment saying so if you'd like.

lib/stdlibs/ballerina/log/0.0.1/go1.2/native/log.go
TharmiganK: The numeric values aren't required, only the ordering, so an `iota` enum works. Note the level comes in as a string from the bal side, so we'd still need a string→level lookup, and the current weights intentionally mirror jBallerina's `logLevelWeight` to ease adding configurable per-module levels later. Happy to switch to the iota form if you prefer it for readability

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/keys.go
TharmiganK: The two reachable error cases are already covered by crypto-keystore-v.bal — wrong keystore password (L35-37) and missing keystore file (L40-42), both asserted is crypto:Error. 

The "Not a valid RSA key" branch on L57-58 only triggers when the store decodes but holds a non-RSA (e.g. EC) key passed to the RSA decoder; that's defensive and needs a dedicated non-RSA .p12 fixture. 

I'm inclined to leave it as defensive code — but if you want it covered I'll generate an EC keystore fixture and add a case that decodes it via decodeRsaPrivateKeyFromKeyStore.

platform/palnative/pal.go
TharmiganK: Added via [91ec024](https://github.com/ballerina-platform/ballerina-lang-go/pull/548/commits/91ec024da1057bf7878f7ae4afb2c943b716d8b4)

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/helpers.go
warunalakshitha: Can we remove that then.

lib/stdlibs/ballerina/log/0.0.1/go1.2/native/log.go
warunalakshitha: yes. better to proceed with go style

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/keys.go
warunalakshitha: Lets add

TharmiganK: Added a test via: [c28fea2](https://github.com/ballerina-platform/ballerina-lang-go/pull/548/commits/c28fea2d80720f08359e52ef1e5a68872f47f54a)

lib/stdlibs/ballerina/log/0.0.1/go1.2/native/log.go
TharmiganK: Changed via [c28fea2](https://github.com/ballerina-platform/ballerina-lang-go/pull/548/commits/c28fea2d80720f08359e52ef1e5a68872f47f54a)

lib/stdlibs/ballerina/crypto/0.0.1/go1.2/native/helpers.go
TharmiganK: Removed via [c28fea2](https://github.com/ballerina-platform/ballerina-lang-go/pull/548/commits/c28fea2d80720f08359e52ef1e5a68872f47f54a)

values/map.go
heshanpadmasiri: IMO no due to couple of reasons
1. Mapping values are pure data without encapsulations (you can't make a field private without making the whole type private) this means there is no way to control how a given record value is created unlike objects nor how it is passed around. User (or any other library) can create a record with the correct shape and what ever function that expect the value should be able to handle that. This gives ability to add a temporal contract to a ballerina value that is completely invisible to people looking at the code. Ideally this should never be needed, but in case it is needed due to very good reason, I don't think runtime/values should bear the burden of that. Instead we should push it to the specific library
2. With this perticular design there is no way for us to prevent another library from overriding this value as well.

IMHO correct solution (**If they must store information out of the ballerina value**) is to make it possible to store string -> any value map in env (and context) what ever out of bound data they want to store. That particular method needs to clearly state we provide no gurante against 2.  and that it is the responsibility of the writer to clean up that data. (we clear for context when strand goes out of bound)

heshanpadmasiri: However I don't necessarily think that is also needed in most cases. If the library needs to track a mapping from a value to some internal state, in a GC friendly way they can have an internal weak map that maps map value to what ever there internal state and have all their externs capture that internal map. Only reason they need to store and get data from env (or context) is when they are different packages.

TharmiganK: Ack, replaced with a weak-pointer key-data map in here: [3f7ae55](https://github.com/ballerina-platform/ballerina-lang-go/pull/548/commits/3f7ae551e478fc05858146ef6f6206addba0c8f5)

warunalakshitha (approved):
LGTM

PR #549, opened 2026-06-19
respect compilation units when resolving symbols

### LINKED ISSUE #523: [Bug]: Import aliases should be for file not package
### Description

Consider fallowing two files in same package 
```
import ballerina/io;
import ballerina/lang.array as foo;

function foo(int[] arr) {
    io:println(foo:length(arr));
}
```

```
import ballerina/io;
import ballerina/lang.map as foo;

function bar(map<int>  a) {
    io:println(foo:length(a));
}

public function main() {
     foo([1, 2]);
     bar({});
}
```
This should work given import alias is per file not per package. But this fails because we treat import alias per package. 

I think in order to fix this case front end should be for compilation unit not for package upto symbol resolution. At which point we can combine compilation units to a package.

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #524: [Improvement]: Move module level xmlns to prefix namespace
### Description

After fixing #523 we no longer need to special case xmlns handling we can move them directly to prefix namespace

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Purpose
Resolves #523.
Resolves #524.

## Approach
- Resolve imports and prefixes per compilation unit so aliases do not leak across files in the same package.
- Scope module-level XML namespace declarations to the declaring compilation unit.

## Automation tests
- Corpus project tests cover per-compilation-unit prefix/import scoping and XML namespace scoping.

## Test environment
- `go test ./...`

## Summary by CodeRabbit

* **Refactor**
  * Updated internal package compilation unit processing
  * Restructured namespace and scope management
  * Simplified import module handling in intermediate representation

* **Chores**
  * Updated BIR format version to 77


warunalakshitha (approved):
LGTM

PR #562, opened 2026-06-24
Add project tests to benchmark

heshanpadmasiri:
## Purpose
$subject


warunalakshitha:
Benchmark tests time increased by 7minutes. is it expected? Should we have separate test type for large projects which does not require benchmark
ex.
https://github.com/ballerina-platform/ballerina-lang/tree/master/tests/jballerina-unit-test/src/test/resources/test-src/jvm/largePackage/modules

heshanpadmasiri:
> Benchmark tests time increased by 7minutes. is it expected? Should we have separate test type for large projects which does not require benchmark ex. https://github.com/ballerina-platform/ballerina-lang/tree/master/tests/jballerina-unit-test/src/test/resources/test-src/jvm/largePackage/modules

Yes that is expected. I am trying to see how long it takes for us to handle very large (I would imagine 1M lines to be closer to what the maximum useful size be for a fresh compilation with dependencies) project so by definition it is a benchmark, not necessarily whether frontend crashes. More crucially these would become benchmarks for future language server as well.

warunalakshitha (approved):
LGTM

PR #589, opened 2026-06-25
Implement `lang.float`

### LINKED ISSUE #569: Migrate `lang.float`

snelusha:
## Purpose
Resolves #569

I ran https://github.com/snelusha/ballerina-lang-go/blob/feat/lang.float/corpus/bal/subset9/09-langlibs/float1-v.bal against both the Go impl and the JBallerina and got this diff. I think this is fine.

|`float1-v.bal`| Go Impl | JBallerina |
   |---|---:|---:|
   | `io:println(floats:fromBitsInt(4));` | `2e-323` | `2.0E-323` |
   | `io:println((12.456).toExpString(2));` | `1.25e+01` | `1.25e+1` |
   | `io:println((12.456).toExpString(()));` | `1.2456e+01` | `1.2456e+1` |

## Summary by CodeRabbit

* **New Features**
  * Added the `lang.float` library with `NaN`, `Infinity`, `PI`, and `E`, plus predicate, math, vararg aggregations, string/hex conversions (with error handling), numeric formatting, and bit-level conversions.
  * Float operations are now available via method-style calls.
* **Bug Fixes**
  * Corrected method-call resolution for float receivers to use the `lang.float` API.
* **Tests**
  * Added/updated float corpus programs and fixtures, including `NaN`/`Infinity`/hex parse outputs and error reporting for negative `fractionDigits`.


heshanpadmasiri:
Given https://github.com/ballerina-platform/ballerina-spec/issues/909 it is okay for our to string to not be exactly same as jBallerina, as long as it is reasonable

model/symbol.go
```diff
@@ -561,6 +564,8 @@ func mapToLangPrefixIfNeeded(prefix string) string {
 		return "lang.int"
 	case "boolean":
 		return "lang.boolean"
+	case "float":
+		return "lang.float"
```
warunalakshitha, line 649: Not sure why this is not covered. Dont we need that

lib/langlibs/go/lang.float/float.go
```diff
@@ -0,0 +1,271 @@
...
+	x := args[0].(float64)
+	if math.IsNaN(x) || math.IsInf(x, 0) {
+		return values.FormatFloat(x), nil
+	}
+	if args[1] == nil {
+		return strconv.FormatFloat(x, 'f', -1, 64), nil
+	}
+	digits := args[1].(int64)
+	if digits < 0 {
+		panic(values.NewErrorWithMessage("fractionDigits must be non-negative"))
```
warunalakshitha: Shall we add a float test with panic or error cases as well.  we can cover some error scenarios with that

lib/langlibs/go/lang.float/float.go
```diff
@@ -0,0 +1,271 @@
...
+	if math.IsNaN(x) || math.IsInf(x, 0) {
+		return values.FormatFloat(x), nil
+	}
+	return normalizeHexFloat(strconv.FormatFloat(x, 'x', -1, 64)), nil
+}
+
+func normalizeHexFloat(s string) string {
+	idx := strings.LastIndexByte(s, 'p')
+	if idx < 0 || idx == len(s)-1 {
+		return s
```
warunalakshitha: Can we cover these missing lines as well with tests.

snelusha: Addressed in [07a320e](https://github.com/ballerina-platform/ballerina-lang-go/pull/589/commits/07a320e784639da3e3785f4225e59f32c4adf06d)

lib/langlibs/go/lang.float/float.go
snelusha: Addressed in [07a320e](https://github.com/ballerina-platform/ballerina-lang-go/pull/589/commits/07a320e784639da3e3785f4225e59f32c4adf06d)

model/symbol.go
snelusha: We need, now this covered

lib/langlibs/go/lang.float/float.go
```diff
@@ -0,0 +1,268 @@
...
+
+func floatPow(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	return math.Pow(args[0].(float64), args[1].(float64)), nil
+}
+
+func floatAtan2(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	return math.Atan2(args[0].(float64), args[1].(float64)), nil
+}
+
+func floatFromString(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
```
gimantha, line 152: Can you double check this implementation? as per Ballerina spec, "1_000.0" should not be a valid float, but this is accepted by Go implementation. Also "Inf", "inf" should also be rejected as they are not valid values.

lib/langlibs/go/lang.float/float.go
```diff
@@ -0,0 +1,268 @@
...
+	s := args[0].(string)
+	unsigned := strings.TrimPrefix(strings.TrimPrefix(s, "+"), "-")
+	if strings.HasPrefix(unsigned, "0x") || strings.HasPrefix(unsigned, "0X") {
+		return values.NewErrorWithMessage("invalid decimal floating point string: " + s), nil
+	}
+	return parseFloat(s)
+}
+
+func floatFromHexString(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
```
gimantha, line 160: "0x1.0" is a valid Hex value, but this is rejected by this method

lib/langlibs/go/lang.float/float.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return out, nil
+}
+
+func floatToFixedString(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	x := args[0].(float64)
+	if math.IsNaN(x) || math.IsInf(x, 0) {
+		return values.FormatFloat(x), nil
+	}
+	if args[1] == nil {
```
gimantha: `io:println((1.0).toFixedString(()));` this should print 1.0

lib/langlibs/go/lang.float/float.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return fmt.Sprintf("%.*f", int(digits), x), nil
+}
+
+func floatToExpString(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	x := args[0].(float64)
+	if math.IsNaN(x) || math.IsInf(x, 0) {
+		return values.FormatFloat(x), nil
+	}
+	if args[1] == nil {
```
gimantha: `io:println((1.0).toExpString(()));` should print 1.0e+00

lib/langlibs/go/lang.float/float.go
snelusha: Fixed and added tests as well.

lib/langlibs/go/lang.float/float.go
snelusha: Fixed and added tests as well.

lib/langlibs/go/lang.float/float.go
snelusha: Fixed and added tests as well.

lib/langlibs/go/lang.float/float.go
snelusha: Fixed and added tests as well.

lib/langlibs/go/lang.float/float.go
gimantha: The API docs say () means “the minimum number of digits required to accurately represent the value” 

```
(1e7).toFixedString(()) -> 10000000.0
(1e20).toFixedString(()) -> 100000000000000000000.0
(1e7).toExpString(()) -> 1.0e+07
```

These should stay minimal: 10000000, 100000000000000000000, and 1e+07. jBallerina gives the same minimal mantissas, modulo its one-digit exponent formatting.

snelusha: Addressed in [`848e6da`](https://github.com/ballerina-platform/ballerina-lang-go/pull/589/changes/848e6dadfbac3760032f0db1e650b8a0432748fa)

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,267 @@ const (
...
+
+func formatMinimalFixed(x float64) string {
+	out := strconv.FormatFloat(x, 'f', -1, 64)
+	if math.Abs(x) < 1e7 && math.Trunc(x) == x && !strings.ContainsRune(out, '.') {
+		return out + ".0"
+	}
+	return out
+}
+
+func formatMinimalExp(x float64) string {
```
gimantha: This still drops required significant fractional digits for toExpString(()) when the value is an integral plain float with a non-zero exponent. The helper only adds .0 when the exponent is 0, so:
```
(10.0).toExpString(()) -> 1e+01
(100.0).toExpString(()) -> 1e+02
(1000000.0).toExpString(()) -> 1e+06
```
But toExpString(()) is supposed to use the same digits as value:toString, only shifted into scientific notation. With the PR’s two-digit exponent style, these should be 1.0e+01, 1.00e+02, and 1.000000e+06. jBallerina gives the same mantissas, with one-digit exponents.

snelusha: Fixed

gimantha:
LGTM

warunalakshitha:
Shall we run same float tests in jbal and confirm same outputs. when i run some outputs were mismatched

snelusha:
> Shall we run same float tests in jbal and confirm same outputs. when i run some outputs were mismatched

I ran the float corpus test with jBallerina and compared the outputs.

The remaining mismatches are formatting differences in float string output, e.g. exponent formatting (`1.79e308` vs `1.79E308`, `1.25e+01` vs `1.25e+1`) and hex formatting (`0x1p0` vs `0x1.0p0`). As Heshan mentioned earlier, these differences are acceptable if the output is reasonable.

For the parsing-related mismatches, I also checked the spec sections [5.2.4](https://ballerina.io/spec/lang/master/#section_5.2.4), [5.2.4.1](https://ballerina.io/spec/lang/master/#section_5.2.4.1), and [5.2.4.2](https://ballerina.io/spec/lang/master/#section_5.2.4.2).

   - `float:fromString("0x1p0")` should be rejected since `fromString` expects a `DecimalFloatingPointNumber`.
   - `float:fromString("1.")` should be rejected since `DottedDecimalNumber` requires at least one digit after `.`.
   - `float:fromString("1_000.0")` should be rejected since `_` is not part of `DecimalNumber`.
   - `float:fromString("Inf")`/`"inf"` should be rejected; the lang.float docs only mention `Infinity`/`NaN`.
   - `float:fromHexString("0x1.0")` seems valid since `DottedHexNumber` allows `HexDigit+ "." HexDigit+`, and the hex
 exponent is optional for dotted hex numbers.

   So I think the current Go behavior is aligned with the spec for these cases.

warunalakshitha (dismissed):
LGTM

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+	fractionDigits := args[1].(int64)
+	if x == 0 || math.IsNaN(x) || math.IsInf(x, 0) {
+		return x, nil
+	}
+	if fractionDigits > 308 {
+		return x, nil
+	}
+	if fractionDigits < -323 {
+		return 0.0, nil
+	}
```
heshanpadmasiri: I don't think this is meaningful given you don't control x. multiplication can still overflow/underflow even for values smaller than this (I assume that's what you are trying to avoid here). Either you should handle overflows properly or not at all, not maybe. It is not clear to me from reading the spec which is the expected behavior, I lean towards if spec don't say we explicitly handle them (like it it says for inf) we don't handle at all

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+	return parseFloat(s)
+}
+
+func floatFromHexString(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	s := args[0].(string)
+	if !hexadecimalFloatingPointStringPattern.MatchString(s) {
+		return values.NewErrorWithMessage("invalid hexadecimal floating point string: " + s), nil
+	}
+	return parseFloat(normalizeHexFloatString(s))
```
heshanpadmasiri: Do we need to validate these using regex? wouldn't parseFloat give us an error anyway?

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+	if strings.HasPrefix(exp, "-") {
+		sign = "-"
+		exp = exp[1:]
+	}
+	exp = strings.TrimLeft(exp, "0")
+	if exp == "" {
+		exp = "0"
+	}
+	return s[:idx+1] + sign + exp
+}
```
heshanpadmasiri: I don't fully understand the purpose of this "normalization" spec require us to comply with `HexFloatingPointLiteral` with some modifications. I ti snot clear to why you need to strip "p" for that?

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+
+func floatToBitsInt(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	return int64(math.Float64bits(args[0].(float64))), nil
+}
+
+func floatFromBitsInt(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	out := math.Float64frombits(uint64(args[0].(int64)))
+	if math.IsNaN(out) {
+		return math.NaN(), nil
+	}
```
heshanpadmasiri: Why?

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+		return values.FormatFloat(x), nil
+	}
+	if args[1] == nil {
+		return formatMinimalFixed(x), nil
+	}
+	digits := args[1].(int64)
+	if digits < 0 {
+		panic(values.NewErrorWithMessage("fractionDigits must be non-negative"))
+	}
+	return fmt.Sprintf("%.*f", int(digits), x), nil
```
heshanpadmasiri: It is not clear to me why you can't just use strconv.FormatFloat (if nil prec = -1 and in other cases just just pass in given prec). you can handle negative digits before conversion?

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+	if err != nil {
+		return s
+	}
+	sign := "+"
+	if exp < 0 {
+		sign = "-"
+		exp = -exp
+	}
+	return fmt.Sprintf("%se%s%02d", s[:idx], sign, exp)
+}
```
heshanpadmasiri: It is not clear to me what is the purpose of this as well?

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+	}
+	exp, err := strconv.Atoi(out[idx+1:])
+	if err != nil || math.Abs(x) >= 1e7 || math.Trunc(x) != x {
+		return out
+	}
+	zeros := exp
+	if zeros < 1 {
+		zeros = 1
+	}
+	return out[:idx] + "." + strings.Repeat("0", zeros) + out[idx:]
```
heshanpadmasiri: Why?

lib/langlibs/go/lang.float/float.go
```diff
@@ -29,31 +33,271 @@ const (
...
+		return values.FormatFloat(x), nil
+	}
+	if args[1] == nil {
+		return normalizeExp(formatMinimalExp(x)), nil
+	}
+	digits := args[1].(int64)
+	if digits < 0 {
+		panic(values.NewErrorWithMessage("fractionDigits must be non-negative"))
+	}
+	return normalizeExp(strconv.FormatFloat(x, 'e', int(digits), 64)), nil
```
heshanpadmasiri: Same a Fixed string I is not clear to me why we need all these special case handlings why can't we directly use FormatFloat? One concern I have is that spec says "any necessary rounding will use roundTiesToEven" I am fairly certain go don't do that that (same behaviour as `round`)

projects/module_context.go
```diff
@@ -452,7 +452,7 @@ func mergeCompilationUnitImports(imports []semantics.CompilationUnitImports) map
 // their langlib key, so they are usable without an import statement. No-op
 // until the lib has been compiled (e.g. while compiling the lib itself).
 func seedMigratedLangLibs(implicitImports map[string]model.ExportedSymbolSpace, publicSymbols map[semantics.PackageIdentifier]model.ExportedSymbolSpace) {
-	for _, name := range []string{"lang.int", "lang.boolean", "lang.decimal", "lang.error", "lang.string", "lang.value", "lang.xml", "lang.float", "lang.array", "lang.map"} {
+	for _, name := range []string{"lang.int", "lang.boolean", "lang.decimal", "lang.error", "lang.float", "lang.string", "lang.value", "lang.xml", "lang.array", "lang.map"} {
```
heshanpadmasiri: Why?

projects/package_resolution_test.go
```diff
@@ -344,8 +345,8 @@ func TestPackageResolution_TransitiveDependency(t *testing.T) {
 	// Step 5: Verify external packages were resolved and cached during compilation.
 	// middlepkg declares aaaleafpkg and leafpkg as direct deps; with the main
 	// project that's 4 packages, plus the always-compiled implicit lang libs
-	// (lang.int, lang.boolean, lang.decimal, lang.error, lang.string, lang.value,
-	// lang.xml, lang.float, lang.array, lang.map, lang.runtime), giving 15 packages total in the cache.
+	// (lang.int, lang.boolean, lang.decimal, lang.error, lang.string, lang.float, lang.value,
```
heshanpadmasiri: why?

heshanpadmasiri (commented):
Front end changes LGTM, I am bit confused by some to string stuff. Each to string method has a contract that it declares in the spec I don't think we should try to satisfy anything beyond that. 

Given ballerina spec is underspecified interms of how floats are represented in string either we should match exactly as jBallerina or just stick to what go gives us **as long as it satisfy contract we give in  the API doc**

heshanpadmasiri (dismissed):
LGTM

warunalakshitha:
Lets fix the build

snelusha:
> Lets fix the build

Fixed.

heshanpadmasiri (approved):
LGTM

heshanpadmasiri (dismissed):

warunalakshitha:
@snelusha shall we fix conflicts

warunalakshitha (approved):
LGTM

PR #590, opened 2026-06-30
test(corpus): skip update for CLI goldens

heshanpadmasiri:
## Summary
- prevent CLI corpus tests from rewriting golden files when --update is passed
- keep fragment-based CLI golden checks unchanged in normal test runs

## Tests
- go test ./corpus --update

## Summary by CodeRabbit

* **Tests**
  * Tightened command output checks so expected results are no longer rewritten automatically during test runs.
  * Fixture mismatches will now fail clearly, helping catch regressions in command output earlier.


heshanpadmasiri:
@azinneera I assume this is correct given these tests are not actually golden file tests (they only check if each line in the txtar file is in the output somewhere not output matches the stdout and stderr in txtar file) so --update flag is not meaningful for them.

azinneera (approved):

PR #612, opened 2026-07-08
Revamp stdlib agent skills with shared references and validation tooling

TharmiganK:
## Purpose

- add-stdlib-support: move manifest/source templates and shared patterns (native-state weak map, PAL hookup, parity risks) into templates/ and references/ files; add a step/gate overview table, run-jballerina-based parity verification, a validate-stdlib-contract gate in the final checklist, and recurring gotchas (langlib import wiring, shared JSON helpers, empty-stderr rule for -v.bal tests)
- fill-stdlib-gap: replace duplicated blocks with pointers to the shared reference files; add the same parity/contract verification steps
- validate-stdlib-contract: add cmd/extract-surface, a parser-based tool that dumps the public surface of a .bal tree for mechanical diffing against jBallerina; declare CONTRACT_VALIDATION.md ephemeral (gitignored)
- stdlib-readme-format: add scripts/check_readmes.py validating per-package READMEs and the aggregator (counts, percentages, verbatim bullet mirroring); permit ### grouping subsections for large modules; make the module-error-type rule conditional on how the type ships
- Fix violations surfaced by the new checkers: stale io row and Total footer in the aggregator, three drifted http behavioural-change bullets, and a backticked Feature/API cell in the http README


warunalakshitha (approved):
LGTM

PR #613, opened 2026-07-09
Fix missing parent directories creation on file write in ballerina/io

TharmiganK:
## Purpose

jBallerina's io module creates any missing parent directories before opening a file for writing or appending, but this Go implementation did not, causing fileWriteString/Lines/Bytes/Json/Xml to fail with ENOENT when the target path's parent directory doesn't exist yet.

Add parent-directory creation to every PAL FS.WriteFile/AppendFile implementation (native, and the two test harness PALs), update the existing error corpus test that had locked in the old (divergent) behavior, and add a corpus test covering auto-creation for all write/append functions.

## Summary by CodeRabbit

* **Bug Fixes**
  * File write and append operations now automatically create missing parent directories for nested paths, improving reliability for workflows that target deep file locations.
* **Tests**
  * Added integration coverage for mkdir/write/read/append behavior across multiple formats (including JSON and XML).
  * Updated fixtures and adjusted writer error assertions to reflect the new behavior when parent directories are missing.


corpus/bal/library/subset2/io-file-mkdir-v.bal
```diff
@@ -0,0 +1,48 @@
...
+    io:println(check io:fileReadJson(jsonPath)); // @output {"k":"v"}
+
+    string xmlPath = "/tmp/bal_io_mkdir_xyz/a/b/data.xml";
+    check io:fileWriteXml(xmlPath, xml `<a/>`);
+    io:println(check io:fileReadXml(xmlPath)); // @output <a/>
+
+    // Appending also creates missing parent directories.
+    string appendPath = "/tmp/bal_io_mkdir_xyz/c/d/append.txt";
+    check io:fileWriteString(appendPath, "First");
+    check io:fileWriteString(appendPath, "Second", io:APPEND);
```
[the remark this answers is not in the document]
TharmiganK: Fixed in [8e707b3](https://github.com/ballerina-nutcracker/ballerina/pull/613/commits/8e707b3c8e95c8c67f5b04de2878b1ff32e5cc82)

heshanpadmasiri (dismissed):
LGTM

test_util/test_util.go
```diff
@@ -333,10 +344,18 @@ func TestPal(stdout io.Writer, stderr io.Writer) pal.Platform {
 				return os.ReadFile(normalizePath(path))
 			},
 			WriteFile: func(path string, data []byte) error {
-				return os.WriteFile(normalizePath(path), data, 0o644)
+				path = normalizePath(path)
```
warunalakshitha: Is this code still used? I think we are now using testpal from test_harness.go. We can remove this duplicated code.

TharmiganK: Removed

warunalakshitha (approved):
LGTM

PR #546, opened 2026-06-18
Support distinct errors

heshanpadmasiri:
## Purpose
$subject
Resolves #540
Depends on #541

## Summary by CodeRabbit

* **New Features**
  * Added full support for `distinct error` types, including correct `is` behavior across modules and compilation units, plus proper handling of combined/intersection error types.

* **Bug Fixes**
  * Improved validation for `distinct` type targets and updated error messages for unsupported inline/anonymous distinct forms.
  * Refined typed error constructor analysis for named detail arguments and strengthened compatibility checks for detail types.

* **Tests**
  * Added and updated corpus and integration fixtures covering `distinct error` behavior, invalid `distinct` declarations, and cross-module/cross-unit `is` checks.

PR #614, opened 2026-07-09
Remove restrictions on special arguments in invocations

### LINKED ISSUE #336: Add support for named arguments for function values
### Description

Fallow on from #335 . To implement this properly we need to track the type descriptor of all function values. Also we need to think how this is going to work when we have a non-atomic function type. Spec don't specify a requirement on type being atomic so either we need to introduce a unimplemented restriction or think of some sort of projection on type descriptor to make this work. I think this is purely a semantic analysis thing, where as desugar should generally fallow the same design as #335 

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #622: [Bug]: Included-record parameter at new(...) call sites are unsupported
### Description

> $Subject

### Steps to Reproduce

Ballerina code:
```bal
public type Config record {|
    string path;
    boolean recursive = false;
|};

public isolated class Listener {
    public isolated function init(*Config config) returns error? {
        _ = config;
    }
}

    Listener _ = check new (path = "/tmp", recursive = true);
}
```

Error:
```console
error[SEMANTIC_ERROR]: failed to determine object type with fitting init function
  --> sample.bal:13:24
   |
13 |     Listener _ = check new (path = "/tmp", recursive = true);
   | 
```

### Affected Version(s)

Ballerina Nutcracker - 0.6.0

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #623: [Bug]: Named arguments at new(...) call sites are unsupported
### Description

 >$Subject

### Steps to Reproduce

Ballerina code:
```bal
public isolated class Listener {
    public isolated function init(string path, boolean recursive = false) returns error? {
        _ = path;
        _ = recursive;
    }
}

    Listener _ = check new (path = "/tmp", recursive = true);
    // Listener _ = check new ("/tmp", recursive = true);
}
```

Error:
```console
panic: unexpected expression type: *ast.BLangNamedArgsExpression
```

### Affected Version(s)

Ballerina Nutcracker - 0.6.0

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #641: [Bug]: `$default$N` name collision across compilation units in the same module
### Description

Every function/method parameter with a default value gets a compiler-generated "default value function" — a small function that computes the default expression — named `$default$0`, `$default$1`, `$default$2`, ... in declaration order. This name is assigned by a counter on `moduleSymbolResolver` (`semantics/symbol_resolver.go`).

A Ballerina module can span multiple `.bal` files, and each file gets its own per-compilation-unit resolver (`forCompilationUnit`). On main branch, `defaultCounter` is a plain `int` field that gets **copied by value** into each per-file resolver instead of being shared across them. The practical effect: every `.bal` file in a module restarts its `$default$N` numbering from 0, independently of every other file in that module.

The generated name isn't just cosmetic — it's the literal string key the BIR runtime registry uses to look up the default-value function at call time (`runtime/internal/modules/registry_impl.go`). So if two *different* files in the same module each declare a function whose first defaultable parameter becomes `$default$0`, both default functions register under the identical key, and whichever one is registered later (order depends on Go map iteration during module registration — nondeterministic per process run) silently overwrites the other in the registry.

At the call site, a call that omits a defaultable argument invokes whatever function currently sits at that key. If the two colliding default functions have different signatures (different parameter counts, as in the repro below), the wrong one gets invoked with too few/many arguments, producing a runtime crash. If their signatures happen to be compatible, the caller instead silently receives the **wrong stdlib's/file's default value** — no crash, no diagnostic, just a wrong answer.

Because the registration order depends on Go's randomized map iteration, this is **flaky**: the same binary run against the same source can pass or fail on different invocations.

### Impact

- Any multi-file module where two or more files each declare at least one defaultable-parameter function/method can be affected — this is core language behavior, not specific to any one stdlib.
- Symptom is either a runtime crash (`error: too many arguments` / `index out of range [N] with length M`, depending on which colliding function's frame layout wins) or, worse, a silently wrong default value with no error at all.
- Flaky: does not reproduce on every single run.

### Steps to Reproduce

Create a small package with two files, each declaring a defaultable-parameter function:

`Ballerina.toml`:
```toml
[package]
org = "testorg"
name = "multifiledefaults"
version = "0.1.0"
```

`main.bal`:
```ballerina
import ballerina/io;

function single(int x = 1) returns int {
    return x;
}

public function main() {
    io:println(single());
    io:println(sum(3, 4));
    io:println(tagged());
}
```

`util.bal`:
```ballerina
function sum(int a, int b, int c = a + b) returns int {
    return c;
}

function tagged(string tag = "t", int v = 9) returns int {
    return v + tag.length();
}
```

`single`'s default (`x = 1`, 0 preceding params) and `sum`'s default (`c = a + b`, 2 preceding params) both become `$default$0` in their respective files, because each file's resolver restarts the counter at 0.

Run it repeatedly against main branch (rebuild once, then invoke the same binary many times — the flakiness is per-process-launch, driven by Go's randomized map iteration, not per-compile):

```shell
git checkout main
go build -o /tmp/bal-upstream ./cli/cmd
for i in $(seq 1 20); do /tmp/bal-upstream run /path/to/multifiledefaults; done
```

Observed on main branch (19/20 runs in one sample):
```
1
error: too many arguments
        at testorg/multifiledefaults:main(main.bal:9)
```

i.e. the call to `sum(3, 4)` (line 9) invokes `single`'s registered default-value function instead of `sum`'s, which expects a different argument count. Occasionally (1/20 in the same sample) the crash instead surfaces at the `single()` call site (line 8) — which colliding function wins is whatever Go's map iteration picked that run.

Correct output (and what the fixed branch prints on every run) should be:
```
1
7
10
```

### Affected Version(s)

Ballerina Nutcracker - 0.6.0

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Summary
Remove restrictions on default, named and included record param arguments form invocations
Depends on #546

## Tests
- go test ./...

Closes ballerina-nutcracker/ballerina#336

## Summary by CodeRabbit

* **New Features**
  * Improved default arguments across lambdas, function values, invocations, object construction, and record fields.
  * Enhanced support for higher-order, typed, narrowed, and dependently typed function signatures.
  * Strengthened included-record and rest-parameter handling, including named-argument matching.

* **Bug Fixes**
  * Refined diagnostics for invalid named arguments, duplicate declarations, and included-record conflicts.
  * Improved function-pointer calls and compiled program reliability.

* **Tests**
  * Added and updated coverage for default arguments, function signatures, object initialization, and related compiler diagnostics.


corpus/bal/subset8/08-function/dependent-fn-3-e.bal
```diff
@@ -14,19 +14,6 @@
 // specific language governing permissions and limitations
 // under the License.
 
-public function main() {
```
heshanpadmasiri, line 17: Since we moving some of the validations to symbol resolution we can't do all these validation in single test

corpus/parser/subset9/09-function/inferred-lambda-args-v.json
warunalakshitha, line 1: Seems mistakenly committed json file

heshanpadmasiri: fixed

desugar/statement.go
```diff
@@ -244,24 +244,14 @@ func walkSimpleVariableDef(cx *functionContext, stmt *ast.BLangSimpleVariableDef
...
-				initStmts = append(initStmts, varDef)
+			for _, fn := range result.functions {
+				cx.pkgCtx.pkg.Functions = append(cx.pkgCtx.pkg.Functions, *fn)
 			}
+			for _, field := range result.recordFields {
+				initStmts = append(initStmts, desugarRecordFieldDefault(cx, field))
+			}
+		} else {
+			cx.pkgCtx.addDefaultClosureOwner(stmt.Var.Expr)
 		}
```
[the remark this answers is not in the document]
heshanpadmasiri: This comment targets an older implementation. walkSimpleVariableDef in desugar/statement.go no longer appends result.functions directly to pkg.Functions. It calls desugarLocalTypeDescDefaults, which lowers every provider through desugarLocalDefaultClosure -> walkExpression -> walkLambdaFunction/desugarNestedFunction and records the resulting local closure mapping. A focused local function-type default whose expression invokes another defaultable function ran successfully and printed 10; corpus/bal/subset9/09-function/inferred-lambda-args-v.bal and narrowed-nested-defaults-v.bal also exercise this path. Applying the suggested package-level registration would be incorrect for these local providers.

semantics/symbol_resolver.go
```diff
@@ -935,40 +964,92 @@ func isExternalFunctionBody(body ast.FunctionBodyNode) bool {
...
+			internalError(resolver, "default symbol allocator not found", pos)
+			return 0, false
+		}
+		return ensureFunctionTypeSignature(alloc, resolver.GetScope(), ty)
+	case *ast.BLangUserDefinedType:
+		return resolver.GetCtx().FunctionSignatureRef(ty.Symbol())
+	default:
+		return 0, false
+	}
+}
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

desugar/desugar.go
```diff
@@ -1092,19 +1175,122 @@ type desugaredRecordFieldResult struct {
...
+	varName, varSymRef := cx.addDesugardSymbol(fnType, model.SymbolKindVariable, false)
+	varIdent := &ast.BLangIdentifier{Value: varName}
+	varIdent.SetDeterminedType(semtypes.NEVER)
+	simpleVar := &ast.BLangSimpleVariable{Name: varIdent}
+	simpleVar.Expr = lambda
+	simpleVar.SetDeterminedType(fnType)
+	simpleVar.SetSymbol(varSymRef)
+	varDef := &ast.BLangSimpleVariableDef{Var: simpleVar}
+	setPositionIfMissing(varDef, fn.GetPosition())
+	return varDef
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

desugar/desugar.go
```diff
@@ -1129,57 +1331,139 @@ func desugarTopLevelTypeDescs(cx *packageContext, pkg *ast.BLangPackage) {
...
+func desugarFunctionParamDefaults(ctx desugarContext, fn *ast.BLangFunction) []*ast.BLangFunction {
+	results := desugarInvokableParamDefaults(ctx, fn.Symbol(), fn.RequiredParams, fn.Scope())
+	for i := range fn.RequiredParams {
+		param := &fn.RequiredParams[i]
+		result := desugarTypeDesc(ctx, param.TypeNode(), fn.Scope())
+		results = append(results, result.functions...)
+		for _, field := range result.recordFields {
+			results = append(results, field.fn)
+		}
 	}
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

context/internal/functionsignatures/store.go
```diff
@@ -0,0 +1,101 @@
...
+func (s *Store) UpdateIncludedRecords(ref model.FunctionSignatureRef, includedRecords []*model.IncludedRecordMetadata) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	index := signatureIndex(ref)
+	sig := s.signatures[index]
+	for i, metadata := range includedRecords {
+		if metadata == nil {
+			continue
+		}
+		sig.SetIncludedRecordMetadata(i, *metadata)
```
warunalakshitha, line 75: I did'nt get why we keep includedRecord meta data in function signature, Should that be aprt of record typedesc?

heshanpadmasiri: included record metadata is used to represent parameters of the form *RecType (https://ballerina.io/spec/lang/master/#included-record-params), at call side then you can pass individual fields as named arguments, or you can pass in the mapping value directly. 

In type resolution phase we lower all these to named arguments to a single mapping constructor at the correct position (basically lowering lowers all the special arguments such as named args, default args, etc to positional args), using the (untyped) function signature.

IMO included record parameters are part of the function signature not the record type itself (you could use the same record type in different ways in different functions for instance)

model/symbolpool/serializer.go
```diff
@@ -686,6 +705,84 @@ func (sw *symbolWriter) writeTypeOp(buf *bytes.Buffer, op model.TypeOp) error {
...
+				}
+			}
+		}
+		metadata := sig.IncludedRecordMetadata[i]
+		if err := write(buf, metadata != nil); err != nil {
+			return err
+		}
+		if metadata == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
heshanpadmasiri: The reported short-slice state is not constructible through the function-signature store. `context/internal/functionsignatures/store.go:Allocate` always calls `model.NewUntypedFunctionSignature`, which creates IncludedRecordMetadata with exactly len(ParamNames) entries; `Store.UpdateIncludedRecords` updates that existing slice through `UntypedFunctionSignature.SetIncludedRecordMetadata`, which explicitly rejects a length mismatch; and `model/symbolpool/deserializer.go:readUntypedFunctionSignatureParams` reconstructs signatures through the same allocator. Repository-wide search found no UntypedFunctionSignature composite literals. Guarding only this serializer access would hide a corrupted signature by dropping metadata, while other consumers such as `model.UntypedFunctionSignature.Index` intentionally rely on the same length invariant. No change is needed in `model/symbolpool/serializer.go:writeUntypedFunctionSignature`.

semantics/lock_analyzer.go
```diff
@@ -214,7 +214,12 @@ func isolatedParamLambdas(ctx *context.CompilerContext, n *ast.BLangInvocation)
 	if !ok {
 		return nil
 	}
-	paramNames := resolvedFn.ParamNames()
+	sig, ok := ctx.GetFunctionSignature(n.Symbol())
+	if !ok {
+		ctx.InternalError("function signature not found", n.GetPosition())
+		return nil
+	}
+	paramNames := sig.ParamNames
```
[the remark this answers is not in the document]
heshanpadmasiri: The concern is invalid for the current monomorphization paths. Both `semantics/type_resolver.go:resolveFunctionCallArgs` (dependently typed functions) and `storeMonomorphizedOpaqueFn` (opaque lang-lib functions) explicitly associate an untyped signature with the monomorphic invocation symbol, so `CompilerContext.GetFunctionSignature(n.Symbol()/fnRef)` does not fail. For opaque array:map, `opaqueFunctionParams` creates the concrete signature with the same `arr`/`func` parameter ordering used by the original OpaqueFunctionSymbol's isolated-parameter predicate; using the invocation symbol is therefore internally consistent and avoids unnecessarily coupling lookup to the polymorphic symbol. Focused execution of `corpus/bal/subset9/09-list/map-isolated-param-v.bal` succeeded, including named `array:map(arr = ..., func = ...)` and lock cases, while `map-isolated-param-e.bal` produced only the expected SEMANTIC_ERROR diagnostics and no INTERNAL_ERROR/function-signature failure. No changes are needed in `semantics/lock_analyzer.go:isolatedParamLambdas` or `isolatedInvocationViolationInner`.

semantics/type_resolver.go
```diff
@@ -3761,9 +3961,8 @@ func determineObjectType(t typeResolver, expr *ast.BLangNewExpression, argTys []
 	}
 	var candidates []candidate
 	for _, alt := range alts {
-		altArgTys, _ := padNewExprArgTypesForDefaults(t, alt.ObjectType, argTys, expr.GetPosition())
```
warunalakshitha, line 3764: type Foo object { function init(int a, int b = 10); }
type Bar object { function init(int a, string c); }
Foo|Bar x = new(1);

Will this work with this change?

heshanpadmasiri: Assuming this is of the form (you need class descriptors) will be fixed with https://github.com/ballerina-nutcracker/ballerina/issues/635

```ballerina
import ballerina/io;

class Foo {
    int a;
    int b;

    function init(int a, int b = 10) {
        self.a = a;
        self.b = b;
    }

    function value() returns int {
        return self.a + self.b;
    }
}

class Bar {
    int a;
    string c;

    function init(int a, string c) {
        self.a = a;
        self.c = c;
    }
}

public function main() {
    Foo|Bar positional = new (1);
    if positional is Foo {
        io:println(positional.value()); // @output 11
    }

    Foo|Bar named = new (a = 2);
    if named is Foo {
        io:println(named.value()); // @output 12
    }
}
```

warunalakshitha (approved):
LGTM

PR #639, opened 2026-07-14
feat(semantics): support isolatedParam

### LINKED ISSUE #604: [New Feature]: Add support for `@isolatedParam`
### Description

I assume we need to support this similar to how we deal with `@typeParam`

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Summary
- support `@isolatedParam` metadata when resolving opaque function calls
- infer arrow-function parameter types from their contextual function type
- add `lang.array:map` compilation and runtime support to exercise isolated callbacks
- add corpus coverage for valid and invalid isolated-parameter use cases

## Testing
- `go test ./...`

Closes #604

## Summary by CodeRabbit

* **New Features**
  * Added contextual type inference for anonymous and arrow functions.
  * Enhanced array mapping with function values, captured variables, inline lambdas, and isolated callbacks.
  * Added support for registering graceful-stop handlers through mapped handler collections.

* **Bug Fixes**
  * Improved validation and source locations for isolation-related compiler diagnostics.
  * Improved handling of expression-bodied functions and function-value invocation.

* **Documentation**
  * Added documentation describing supported language features and restrictions.


ast/node_builder.go
```diff
@@ -4321,8 +4321,49 @@ func (n *NodeBuilder) TransformImplicitAnonymousFunctionParameters(implicitAnony
...
+			paramValue := paramName.Text()
+			if paramValue == "_" || paramValue == "'_" {
+				n.cx.SyntaxError("'_' cannot be used as an identifier", paramPos)
+			}
+			ident = createIdentifier(paramPos, &paramValue, &paramValue)
+		}
+		fn.RequiredParams[i].Name = &ident
+		fn.RequiredParams[i].pos = getPosition(n.de(), param)
+		fn.RequiredParams[i].SetRequiredParam()
+	}
```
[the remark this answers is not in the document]
heshanpadmasiri: The concern is not reachable through the compiler pipeline. parser/tree always creates a SimpleNameReferenceNode with a token child; an incomplete inferred-lambda parameter receives a non-nil missing token, so SimpleNameReferenceNode.Name() is non-nil even when IsMissing() is true. In addition, ast.NodeBuilder.TransformModulePart reports syntax diagnostics and skips a top-level member containing the malformed lambda before TransformImplicitAnonymousFunctionExpression can run. A focused CLI check with `(first,) => first` produced the expected `missing identifier` syntax diagnostic and no panic. No change is needed in ast/node_builder.go or TransformImplicitAnonymousFunctionExpression.

lib/langlibs/go/lang.array/array.go
```diff
@@ -17,17 +17,50 @@
...
 )
 
 const (
 	orgName    = "ballerina"
 	moduleName = "lang.array"
 )
 
+func arrayMap(ctx *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	source := args[0].(*values.List)
+	callback := args[1].(*values.Function)
```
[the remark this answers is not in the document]
heshanpadmasiri: The assertions in lib/langlibs/go/lang.array.arrayMap rely on the typed internal extern-call ABI, as do many other langlib implementations (including lang.map, lang.xml, lang.float, lang.decimal, and lang.error). semantics/type_resolver.go:monomorphizeArrayMap constructs a signature whose first parameter is a list subtype and whose second parameter is a compatible function, and compilation stops on violations. A focused Ballerina check passing an int as either argument produced semantic diagnostics and never invoked arrayMap. Adding checks only for these two assertions would be inconsistent partial hardening (argument count and many sibling externs would still panic on malformed internal calls), so this is not worth changing in lib/langlibs/go/lang.array/array.go.

lib/langlibs/go/lang.array/array.go
```diff
@@ -17,17 +17,50 @@
...
+			return nil, err
+		}
+		items[i] = result
+	}
+
+	resultDef := semtypes.NewListDefinition()
+	resultTy := resultDef.DefineListTypeWrappedWithEnvSemType(ctx.Env.TypeEnv, resultMemberTy)
+	atomic := semtypes.ToListAtomicType(ctx.TypeCtx, resultTy)
+	filler, _ := values.FillerFactoryFor(ctx.TypeCtx, resultMemberTy)
+	return values.NewList(resultTy, atomic, false, filler, 0, items), nil
```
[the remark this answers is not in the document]
heshanpadmasiri: A nil filler is a supported values.List state, not a deferred construction failure. values.NewList accepts a nil FillerFactory, values.List.Append does not require a filler, and FillingGet/FillingSet intentionally raise the Ballerina `no filler value` panic only when indexed growth must create intermediate members. Rejecting array:map when values.FillerFactoryFor returns false would incorrectly reject valid arrays whose member type has no filler. A focused run mapped int values to `Choice[]` where `Choice` is the no-filler singleton union `"a"|"b"`; map returned `["a"]`, and push successfully produced `["a","b"]`. Existing no-filler corpus cases also expect failure only on gap-filling indexed growth. No change should be made to lib/langlibs/go/lang.array/array.go:arrayMap, values/list.go, or values/values.go.

semantics/type_resolver.go
```diff
@@ -1954,6 +1992,111 @@ func resolveLambdaFunctionExpr(t typeResolver, chain *binding, e *ast.BLangLambd
...
+		implicitImports:     make(map[string]ast.BLangImportPackage),
+		mappingAtomToBType:  make(map[*semtypes.MappingAtomicType]ast.BType),
+		monoCounters:        make(map[string]int),
+		scope:               e.Function.Scope(),
+		mappingAtomToSymRef: make(map[*semtypes.MappingAtomicType]model.SymbolRef),
+	}
+	boundaryChain := &binding{flags: bindingFlagFunctionBoundary, prev: chain}
+	prevCaptured := t.getCapturedVars()
+	ft.setCapturedVars(make(map[model.SymbolRef]bool))
+	body := e.Function.Body.(*ast.BLangExprFunctionBody)
```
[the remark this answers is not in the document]
heshanpadmasiri: The asserted body kind is guaranteed by the parser-to-AST contract. parser/parser.go:parseImplicitAnonFuncWithParams always parses an expression and creates an ImplicitAnonymousFunctionExpressionNode; ast/node_builder.go:TransformImplicitAnonymousFunctionExpression always wraps that expression in ast.BLangExprFunctionBody and is the only source of BLangLambdaFunction nodes with InferredParams set. Syntax recovery still produces an expression node inside that wrapper, while explicit block-bodied lambdas have InferredParams false and use resolveLambdaFunctionExpr instead. A future block-bodied form would require coordinated parser and AST changes and should update this invariant then. Adding defensive recovery for an AST shape that the real compiler pipeline cannot produce is not worth changing semantics/type_resolver.go:resolveInferredLambdaFunctionExpr.

semantics/type_resolver.go
```diff
@@ -7498,59 +7642,64 @@ func (m *monomorphicOpaqueFn) PolymorphicSymbol() model.SymbolRef { return m.pol
...
 	space := sym.SymbolSpace
 	idx := space.AppendSymbol(mono)
 	mono.name = fmt.Sprintf("%s$mono$%d", sym.Name(), idx)
 	ref := space.RefAt(idx)
 	if sym.Store != nil {
-		sym.Store(ref, containerTy)
+		sym.Store(ref, cacheKeys...)
 	}
 	return ref
 }
```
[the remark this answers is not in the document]
heshanpadmasiri: No partially initialized read is reachable in the current compiler pipeline. semantics/type_resolver.go:storeMonomorphizedOpaqueFn assigns mono.name before publishing the SymbolRef through OpaqueFunctionSymbol.Store; the cache mutex provides synchronization for a later Lookup. AppendSymbol does not publish the symbol through SymbolSpace's name lookup table, and no local-resolution goroutine iterates the symbol space; model.SymbolSpace.Symbols/serialization are used only after concurrent local-node resolution has completed. Preserving the index-based name before AppendSymbol would also require a new atomic append-and-initialize SymbolSpace API rather than the suggested reordering, since the final index is assigned inside AppendSymbol. With no current observer or demonstrated race, changes to semantics/type_resolver.go:storeMonomorphizedOpaqueFn and model/symbol.go:SymbolSpace are not worth making for this concern.

corpus/integration_test.go
```diff
@@ -80,11 +80,11 @@ var (
 		// once that's registered in DiagnosticEnv).
 		"project/missing-package-e",
 		"project/parse-error-e",
-		// Pre-existing -fp.bal test that does not currently surface a runtime
+		// Pre-existing future test that does not currently surface a runtime
 		// panic or a compile-time `fatal[...]` bailout, so it does not satisfy
 		// the future-test contract yet. Tracked separately.
 		"subset8/08-future/fieldlvalue1-fp.bal",
-		// https://github.com/ballerina-nutcracker/ballerina/issues/417
+		// https://github.com/ballerina-platform/ballerina-lang-go/issues/417
```
warunalakshitha: Shall we fix this

runtime/internal/exec/dispatch_test.go
```diff
@@ -0,0 +1,44 @@
...
+package exec
+
+import (
+	"testing"
+
+	"ballerina/runtime/extern"
+	"ballerina/values"
+)
+
+func TestInvokeRejectsTypedNilHandle(t *testing.T) {
```
warunalakshitha: Will dispatch logic covers with normal bal tests. Why do we need a separate unit test?

runtime/internal/exec/dispatch.go
```diff
@@ -248,7 +249,24 @@ func containsByte(s string, b byte) bool {
...
+	if h == (*values.Function)(nil) {
+		return nil, fmt.Errorf("nil function value")
+	}
+	switch h := h.(type) {
+	case *InvokableHandle:
+		return h.invoke(ctx, args)
+	case *values.Function:
+		handle, err := NewFunctionValueHandle(ctx.Env, h)
+		if err != nil {
+			return nil, err
```
warunalakshitha, line 264: can we have nil here?

warunalakshitha:
following sample end up with compilation error

```
isolated function square(int n) returns int {
    return n * n;
}

function demo(int[] values) returns int {
    lock {
        // OK: arrow lambda — isolated-ness inferred automatically.
        int a = values.map(n => square(n)).length();

        // ERROR: explicit lambda — same body, but not inferred isolated.
        // int b = values.map(function(int n) returns int {
        //     return square(n);
        // }).length();

        // OK: explicit lambda works once marked `isolated` by hand.
        int c = values.map(isolated function(int n) returns int {
            return square(n);
        }).length();

        return a + c;
    }
}

public function main() {
    int _ = demo([1, 2, 3]);
}``

heshanpadmasiri:
> following sample end up with compilation error
> 
> ```
> isolated function square(int n) returns int {
>     return n * n;
> }
> 
> function demo(int[] values) returns int {
>     lock {
>         // OK: arrow lambda — isolated-ness inferred automatically.
>         int a = values.map(n => square(n)).length();
> 
>         // ERROR: explicit lambda — same body, but not inferred isolated.
>         // int b = values.map(function(int n) returns int {
>         //     return square(n);
>         // }).length();
> 
>         // OK: explicit lambda works once marked `isolated` by hand.
>         int c = values.map(isolated function(int n) returns int {
>             return square(n);
>         }).length();
> 
>         return a + c;
>     }
> }
> 
> public function main() {
>     int _ = demo([1, 2, 3]);
> }``
> ```

We don't support isolation inference https://github.com/ballerina-nutcracker/ballerina/issues/363

runtime/internal/exec/dispatch_test.go
heshanpadmasiri: You can technically pass nil fn values in extern functions. Move to an extern test

runtime/internal/exec/dispatch.go
heshanpadmasiri: Assume this is about the handle h, err; yes. We allow extern functions to call function values as well and they will reach this function too. We should never code gen h == nil. That being said I am not entirely sure if we should just panic instead. (I would assume if that happen that's programming error)

warunalakshitha: means error non nil cases.

warunalakshitha (approved):
LGTM

PR #644, opened 2026-07-16
Add Level 3 standard library support - Phase 1

TharmiganK:
## Summary

Depends on this PR: https://github.com/ballerina-nutcracker/ballerina/pull/585 and the following issues:
- https://github.com/ballerina-nutcracker/ballerina/issues/640
- https://github.com/ballerina-nutcracker/ballerina/issues/641
- https://github.com/ballerina-nutcracker/ballerina/issues/642

This PR merges `stdlibs-l3`'s work on top of `http-svc`. It adds six new Go-native `ballerina/*` standard library packages, extends `ballerina/http` with multipart/form-data payload support built on the new `ballerina/mime` package, and adds the Platform Adaptation Layer (PAL) surface needed to back the filesystem- and socket-based packages in this round. Compiler/interpreter-core changes (bug fixes and new langlib functions) unrelated to this stdlib work are documented separately in `lang-changes.md` and are not repeated here, aside from the PAL additions below.

## New standard libraries

| Package | Supported | Partially Supported | Not Yet Supported | Support % |
|---|---|---|---|---|
| `file` | 21 | 0 | 0 | 100% |
| `ldap` | 15 | 2 | 0 | 83% |
| `mime` | 13 | 2 | 1 | 81% |
| `tcp` | 18 | 1 | 1 | 82% |
| `udp` | 15 | 2 | 2 | 79% |
| `uuid` | 19 | 1 | 0 | 95% |

(Support % = `round(Supported / Total * 100)`, Total = Supported + Partially Supported + Not Yet Supported + Cannot Support. Full per-package tables and behavioural-change notes live in each package's own README under `lib/stdlibs/ballerina/<name>/0.0.1/`.)

### `ballerina/mime`

MIME media type parsing, content disposition handling, entity header/body management, multipart bodies, and Base64 encode/decode, per RFC 2045/2046.

- Media type parsing (`getMediaType`) and content disposition parsing (`getContentDispositionObject`).
- Entity header management (`setHeader`/`getHeader`/`getHeaders`/`getHeaderNames`/`addHeader`/`removeHeader`/`removeAllHeaders`/`hasHeader`) and content metadata (type, ID, length, disposition).
- Entity body as text, JSON, byte array, or multipart parts (`setText`/`getText`, `setJson`/`getJson`, `setByteArray`/`getByteArray`, `setBodyParts`/`getBodyParts`, generic `setBody`).
- Base64 encode/decode for strings and byte arrays.
- Partially supported: multipart bodies — flat (single-level) only, a part whose own body is itself multipart is not recursively decoded; `distinct` error subtypes are plain `error` aliases.
- Not yet supported: entity XML body (blocked on XML type support in this interpreter).

### `ballerina/uuid`

RFC 4122 UUID generation (versions 1, 3, 4, 5), validation, version detection, and string/byte/record conversion.

- Generation as string or record for all four supported versions, plus a `createRandomUuid()` alias for type 4.
- Validation, version detection, and conversions between string/byte-array/record forms.
- Partially supported: `distinct` error subtype (`Error` is a plain `error` alias).
- Behavioural change: type 1 UUIDs use a random 6-byte node identifier instead of the host's MAC address (portability/privacy); still RFC-valid and passes `validate()`.

### `ballerina/file`

File, directory, and path manipulation, built on the new PAL filesystem interface (see below).

- Directory operations (cwd, create, remove — recursive and non-recursive), file operations (create, rename/move, copy with option flags), metadata retrieval, and directory listing.
- Temporary file/directory creation and path utilities (absolute resolution, basename, parent, normalize, split, join, relative path).
- Directory change listener/file watcher service (`onCreate`/`onModify`/`onDelete`), backed by the new PAL `Watch` function (`fsnotify`).
- Full package coverage: 21/21 rows Supported (100%) — no Not Yet Supported or Cannot Support rows remain.
- Behavioural changes: `distinct` error types are flattened to plain `file:Error` aliases; Windows detection relies on the `OS` environment variable being set at startup; `gracefulStop()` deterministically releases the underlying OS watch (jBallerina's is a no-op that leaves the watch thread running); `attach()` returns its validation error through the documented `error?` return instead of panicking.

### `ballerina/ldap`

Connect, authenticate, and interact with LDAP directory servers — search, add, modify, rename, compare, and delete entries.

- Plaintext or TLS (PEM certificate) client init with host-name verification and TLS version selection.
- Add/delete/rename/update an entry, attribute comparison, single-entry retrieval and subtree search (with attributes inferred from the target record when not given explicitly), connection lifecycle.
- Binary attribute decoding (`objectGUID`/`objectSid` to canonical string form; other binary values base64-encoded per RFC 2849).
- Cannot support: PKCS12 trust stores — use a PEM certificate file path instead.
- Partially supported: search result references/controls (the underlying Go LDAP client flattens referral URIs across references without preserving per-reference message IDs/controls, so this port synthesizes one `SearchReference` per URI with a placeholder message ID); `distinct` error subtypes are plain `error` aliases.
- Behavioural changes: two mismatched jBallerina result-code strings are corrected to match the declared `Status` enum; client-side-only failures map to `OTHER`; TLS version constraints are scoped per-connection rather than JVM-wide.

### `ballerina/tcp`

Low-level, connection-oriented TCP client and listener/service APIs for exchanging raw bytes with no built-in framing.

- Client connect (local-interface binding, independent read/write timeouts, TLS via PEM certificate) and listener (TLS via cert+key pair, cipher suite restriction).
- Attach/detach a service on a listener, start/gracefulStop/immediateStop, `onConnect`/`onBytes`/`onError`/`onClose` dispatch (including auto-reply), inline `service on new tcp:Listener(...)` declarations.
- Cannot support: PKCS12 trust store (client) or key store (listener) — use a PEM certificate/key file pair instead.
- Not yet supported: `Service`/`ConnectionService` as `distinct service object` types (declared as plain `service object`, since this interpreter can't yet attach an anonymous inline service body to a `distinct` target type, and the inline declaration style was prioritized).
- Partially supported: `distinct` error subtype (`Error` is a plain `error` alias).
- Behavioural changes (all deliberate, documented divergences — see the "known limitations" note below): plain default-valued config records instead of included-record spread parameters; a connection closes if `onConnect` returns an error (jBallerina leaves it open); `onClose` fires exactly once; `Caller` identity fields are computed once at accept time; `immediateStop()` actually force-closes instead of being a no-op; `detach()` validates the given service.

### `ballerina/udp`

Low-level connectionless and connection-oriented UDP client APIs, plus a listener/service/caller surface for datagram exchange.

- Connectionless `Client` and connection-oriented `ConnectClient` (with outbound auto-fragmentation across datagrams), listener with attach/detach/start/gracefulStop/immediateStop, `onBytes`/`onDatagram`/`onError` dispatch (including auto-reply), inline listener service declarations.
- Not yet supported: a listener bound to a single fixed remote peer; `Service` as a `distinct service object` type (same inline-attachment reason as `tcp`).
- Partially supported: dispatch only invokes the fixed two-parameter `onBytes`/`onDatagram` forms (jBallerina also accepts other parameter orders/subsets via reflection); `distinct` error subtype (`Error` is a plain `error` alias).
- Behavioural changes: same default-valued-config-record pattern as `tcp`; `detach()` validates the given service; `immediateStop()` force-closes immediately (UDP has no per-connection state to drain, so it behaves the same as `gracefulStop()`).

## `ballerina/http` enhancements

Built on top of the new `ballerina/mime` package:

- **Multipart request/response bodies**: `setBodyParts(mime:Entity[], string? contentType)` / `getBodyParts() returns mime:Entity[]|error` on both `Request` and `Response`. Flat (single-level) multipart only, matching `mime`'s own limitation.
- **Generic payload setter**: `setPayload(json|byte[]|mime:Entity[] payload, string? contentType)`, dispatching by runtime type to the existing `setTextPayload`/`setBinaryPayload`/`setJsonPayload`/`setBodyParts`. jBallerina's broader signature (streams, arbitrary `anydata`) isn't accepted, since this port has no corresponding stream/`toJson()` support yet.
- **Form-urlencoded parsing**: `getFormParams()` decodes an `application/x-www-form-urlencoded` body into a `map<string>`.
- `RequestMessage` (the type accepted by the client's `post`/`put`/`patch`/`execute`) is widened to include `mime:Entity[]`.

Net effect on `http`'s own coverage: 25 Supported / 6 Partially Supported / 45 Not Yet Supported (33%, unchanged as a percentage but several rows moved out of Not Yet Supported).

## Platform Adaptation Layer (PAL) additions

New capability surface needed to back `file`, `tcp`, `udp`, and `ldap`, added to the PAL rather than any individual stdlib (per this repo's rule that all platform interaction goes through PAL, not the OS/net APIs directly):

- **`platform/pal/platform.go`**: new file-operation surface (`Getwd`, `Mkdir`, `MkdirAll`, `Remove`, `RemoveAll`, `Rename`, `CreateFile`, `Stat`, `Lstat`, `ReadDir`, `Copy`, `CreateTemp`, `CreateTempDir`, `Readlink`, `Watch`) and `Time.Sleep`.
- **`platform/palnative/net.go`, `tls.go`, `watcher.go`** (new) plus `pal.go`/`http.go` (extended): real OS-backed implementations — filesystem watching via `fsnotify`, TLS config helpers for the socket-based packages, net helpers shared by `tcp`/`udp`.
- **`test_util/testharness/test_harness.go`**: matching test-PAL wiring so `corpus/extern` tests can exercise the new surface.

## Known limitations applying across this round

- **`distinct` types**: none of the new packages' error types or service marker types are declared `distinct` — they're plain aliases/plain `service object` types. This is consistent with every existing Go-native stdlib in this interpreter and is called out per-package above where it affects `is`-based type narrowing.
- **Included-record config parameters**: `tcp`, `udp`, and `ldap`'s `Client.init`/`Listener.init` take a plain (default-valued, for `tcp`/`udp`) configuration record instead of jBallerina's included-record spread parameter (`*Config`). Call sites pass a record literal (`check new("host", 80, {localHost: "x"})`) instead of named-argument spread (`check new("host", 80, localHost = "x")`). This is a deliberate public-API divergence working around a current interpreter limitation resolving included-record parameters that follow other positional parameters in a multi-import module — tracked separately, not part of this PR's scope.

## Testing

Each new package is covered by corpus tests under `corpus/bal/` (compile-and-run `-v.bal`/error `-e.bal` cases) and, for the socket/listener and multipart-over-the-wire behavior, `corpus/extern` integration tests that spin up real listeners/clients. `go build ./...` and `go test ./...` are clean on this branch.

## Out of scope

The packages requiring annotation support (not yet available in this interpreter) were not attempted in this round.


TharmiganK:
Closing this PR, since the changes will be sent via separate PRs for each std lib

PR #645, opened 2026-07-16
Fix semtype to string rendering not showing builtin types in some contexts

### LINKED ISSUE #642: [Bug]: Recursive-atom index collision across list/mapping kinds in semtype to-string
### Description

Recursive semtypes (used for self-referential types like `json`, `anydata`, or any user-defined recursive `record`/array type) are represented internally as "recursive atoms" (`recAtom`). List-recursion, mapping-recursion, function-recursion, xml-recursion, and distinct-type-recursion each draw their index from their **own** independent, 0-based counter on `env` (`recListAtom()`, `recMappingAtom()`, `recFunctionAtom()`, etc. in `semtypes/env.go`). Both the list-atom table and the mapping-atom table are pre-padded to the same reserved length (`predefinedTypeEnv.ReservedRecAtomCount()`), so in practice the *first* user-/stdlib-defined recursive list type and the *first* recursive mapping type end up assigned the **same numeric index**.

On main branch, an atom's "canonical key" — used both for BDD-node interning and, critically, for the cycle-detection `visited` set in the semtype-to-string printer (`semtypes/to_string.go`) — was just `{index, rec: true}`, with no indication of *which counter* (list vs. mapping vs. function vs. xml vs. distinct) the index came from. So a list atom and a mapping atom sharing the same index produce the *same* key.

The printer's cycle detector (`listAtomToString`/`mappingAtomToString` in `to_string.go`) marks an atom's key as `visited` before recursing into its members, to know when it's looped back to an already-open atom (and should print `...` instead of recursing forever). Because the key didn't carry a kind, entering a mapping atom whose index happened to equal an outer list atom's index falsely looks like "we're already printing this" — the printer stops early and prints `...` instead of expanding the mapping atom's actual structure, even though it's a completely different type it has never seen before.

This doesn't affect type-checking correctness (subtyping, actual program behavior) — `semtypes.ToString` is only used for producing human-readable type names in **compile-time semantic error messages** and in the debug BIR pretty-printer (`--dump-bir`). But those messages come out truncated/misleading, and the exact shape depends on allocation order across concurrently-desugared functions/classes/services (compiler stages 5–10 run per-module concurrently), so the same source can print differently between compiles.

### Impact

- Misleading/truncated type names in `SEMANTIC_ERROR` diagnostics and BIR debug dumps whenever a recursive type mixes list-recursion and mapping-recursion (e.g. `anydata`, `json`, or any user `record` type with a self-referencing array/table field) — no user code needs to reference any stdlib for this to trigger; it is core language behavior.
- No effect on program *execution* — this is purely a to-string/diagnostic-quality bug.

### Steps to Reproduce

```ballerina
type Config record {|
    anydata...;
|};

public function main() {
    Config c = {};
    int x = c;
}
```

(`anydata...` makes `Config`'s rest-field type an inline, unnamed `anydata` — an open record whose value type is the built-in recursive `anydata` union, which mixes list and mapping recursion. Using it inline, rather than through a named type alias, forces the printer to expand it structurally instead of just printing the name `anydata`.)

Run against main branch:
```shell
git checkout main
go run ./cli/cmd run /path/to/repro.bal
```

Output on main branch:
```
error[SEMANTIC_ERROR]: incompatible type: expected int, got {| nil|boolean|int|float|decimal|string|regexp|xml|[nil|boolean|int|float|decimal|string|regexp|xml|...|...|table...]|{| nil|boolean|int|float|decimal|string|regexp|xml|...|...|table... |}|table... |}
```

Note the `[nil|...|xml|...|...|table...]` segment: the list member's own mapping-typed element collapses straight to `...` without ever showing its structure — a mapping atom is being (incorrectly) treated as already-visited.

### Affected Version(s)

Ballerina Nutcracker - 0.6.0

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Purpose
$subject
Fixes https://github.com/ballerina-nutcracker/ballerina/issues/642

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved semantic type error messages with shorter, clearer representations for unions, arrays, `anydata`, functions, records, and optional fields.
  * Preserved accurate source locations and highlighting for reported type errors.
  * Standardized type formatting across compiler diagnostics for more consistent output.

### DIFF
--- a/corpus/bir/library/subset1/http-client-methods-v.txt
+++ b/corpus/bir/library/subset1/http-client-methods-v.txt
@@ -72,7 +72,7 @@ main() -> nil|error{
     $desugar$8 = %28;
     %30 = ConstantLoad key
     %31 = ConstantLoad val
-    %32 = newMap {| nil|boolean|int|float|decimal|string|...|...... |}{%30=%31}
+    %32 = newMap {| json... |}{%30=%31}
     $desugar$9 = %32;
     %34 = $default$22($desugar$8,$desugar$9) -> bb14;
   }
--- a/corpus/bir/library/subset1/http-client-post-v.txt
+++ b/corpus/bir/library/subset1/http-client-post-v.txt
@@ -100,7 +100,7 @@ main() -> nil|error{
     %44 = ConstantLoad test
     %45 = ConstantLoad count
     %46 = ConstantLoad 1
-    %47 = newMap {| nil|boolean|int|float|decimal|string|...|...... |}{%43=%44, %45=%46}
+    %47 = newMap {| json... |}{%43=%44, %45=%46}
     $desugar$10 = %47;
     %49 = $default$24($desugar$9,$desugar$10) -> bb18;
   }
--- a/corpus/bir/library/subset2/civil-functions-v.txt
+++ b/corpus/bir/library/subset2/civil-functions-v.txt
@@ -58,7 +58,7 @@ main() -> nil|error{
     %32 = ConstantLoad 6
     %33 = ConstantLoad day
     %34 = ConstantLoad 15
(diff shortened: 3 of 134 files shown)


TharmiganK (approved):
LGTM

PR #656, opened 2026-07-21
feat(langlib): add core lang lib support for stdlib L3

### LINKED ISSUE #640: Add support for lang libs required to onboard level 3 standard libraries
The following lang lib functional support is required to implement the initial support for level 3 standard libraries:

| Lang Lib | New functions |
|---|---|
| `lang.array` | `removeAll`, `toBase16`, `toBase64`, `fromBase16`, `fromBase64`,`indexOf` and `remove` |
| `lang.string` | `substring`, `equalsIgnoreCaseAscii`, `toLowerAscii`, `toUpperAscii`, `trim`, `toBytes`, `fromBytes` |
| `lang.int` | `fromString`, `fromHexString` |
| `lang.map` | `hasKey` |
| `lang.runtime` | `sleep(decimal)` |
| `lang.value` | `toString(anydata)` |

TharmiganK:
## Purpose

> Fixes: https://github.com/ballerina-nutcracker/ballerina/issues/640

- Add new langlib functions across several modules:
  - `lang.array`: `indexOf`, `remove`, `removeAll`, `toBase16`/`fromBase16`, `toBase64`/`fromBase64`
  - `lang.string`: `substring`, `toLowerAscii`/`toUpperAscii`, `equalsIgnoreCaseAscii`, `trim`, `toBytes`/`fromBytes`
  - `lang.int`: `fromString`, `fromHexString`
  - `lang.map`: `hasKey`
  - `lang.runtime`: `sleep` (new module)
  - `lang.value`: `toString` (new module)
- Wire `indexOf`/`remove` as opaque functions requiring real monomorphization in the type resolver, plus `List.RemoveAt`/`Clear` helpers backing them
- Add a `Time.Sleep` PAL primitive (native + WASM) to back `lang.runtime:sleep`
- Fix a `lifecycle.transition` panic when a stop signal arrives after the runtime already reached `Stopped` via another path (e.g. a `$start` failure cascading straight to `Stopped`) — now a no-op instead of a panic
- Add `.gitattributes` entries to exempt corpus golden directories from line-ending normalization, since some fixtures embed literal CR bytes compared byte-for-byte
- Regenerate BIR goldens affected by the atom-table shift from the new langlib opaque symbols

## Summary by CodeRabbit

* **New Features**
  * Added `lang.array` support for searching (`indexOf`), element removal (`remove`), and clearing (`removeAll`).
  * Added `lang.int` parsing from decimal/hex strings, plus `lang.map:hasKey`.
  * Added `lang.runtime:sleep`, `lang.value:toString`, and new `lang.string` utilities (substring, ASCII trim/case/compare).
* **Bug Fixes**
  * Improved out-of-range and invalid-input handling for arrays and strings.
* **Tests**
  * Added/updated corpus and integration fixtures for expected success and runtime error outputs.
  * Added list memory-release tests for `RemoveAt` and `Clear`.
* **Chores**
  * Ensured consistent byte-for-byte corpus handling by disabling text normalization for corpus subdirectories.


values/list.go
```diff
@@ -130,6 +130,32 @@ func (l *List) Append(tc semtypes.Context, vs ...BalValue) {
 	l.elems = append(l.elems, vs...)
 }
 
+// RemoveAt removes and returns the element at idx, shifting subsequent elements down.
+// Panics if the list's inherent type is fixed-length.
+func (l *List) RemoveAt(idx int) BalValue {
+	l.checkMutable()
+	l.checkNotFixedLength()
+	val := l.elems[idx]
+	l.elems = append(l.elems[:idx], l.elems[idx+1:]...)
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [3198eb6](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/3198eb63785467a0d5e620c43760afe27b54f1d8)

values/json.go
```diff
@@ -68,6 +71,49 @@ func balToGoJSON(v BalValue) any {
...
+		return p.listTy, p.mapTy
+	}
+	jsonTy := semtypes.CreateJSON(ctx)
+	listLd := semtypes.NewListDefinition()
+	mapMd := semtypes.NewMappingDefinition()
+	listTy := listLd.DefineListTypeWrappedWithEnvSemType(env, jsonTy)
+	mapTy := mapMd.DefineMappingTypeWrapped(env, nil, jsonTy)
+	p := jsonTypePair{listTy, mapTy}
+	jsonTypesByEnv.Store(key, p)
+	stdruntime.AddCleanup(env, cleanupJSONTypes, key)
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [3198eb6](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/3198eb63785467a0d5e620c43760afe27b54f1d8)

runtime/lifecycle_test.go
```diff
@@ -366,6 +366,63 @@ service on l {
...
+	if code != 1 {
+		t.Fatalf("expected start-failure exit code 1, got %d", code)
+	}
+	if got, want := pal.Stdout(), "graceful:one\n"; got != want {
+		t.Fatalf("unexpected stdout: got %q, want %q", got, want)
+	}
+
+	pal.Send(palSignalGracefulStop)
+	time.Sleep(100 * time.Millisecond)
+}
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [3198eb6](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/3198eb63785467a0d5e620c43760afe27b54f1d8)

lib/langlibs/go/lang.array/array.go
```diff
@@ -95,6 +96,50 @@ func initArrayModule(rt *runtime.Runtime) {
...
+		if len(args) > 2 && args[2] != nil {
+			if si, ok := args[2].(int64); ok {
+				startIndex = si
+			}
+		}
+		if startIndex < 0 {
+			panic(values.NewErrorWithMessage(fmt.Sprintf("invalid array index: %d", startIndex)))
+		}
+		for i := int(startIndex); i < list.Len(); i++ {
+			if values.DeepEquals(list.Get(i), val) {
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [3198eb6](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/3198eb63785467a0d5e620c43760afe27b54f1d8)

lib/langlibs/ballerina/lang.string/0.0.1/any/lang.string.bal
```diff
@@ -20,14 +20,50 @@
...
+# + str - the string to be converted
+# + return - `str` with any occurrences of an ASCII lowercase character replaced with the
+#   equivalent uppercase character
+public isolated function toUpperAscii(string str) returns string = external;
+
+# Removes ASCII white space characters from the start and end of a string.
+#
+# + str - the string to be trimmed
+# + return - `str` with leading or trailing ASCII white space characters removed
+public isolated function trim(string str) returns string = external;
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [3198eb6](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/3198eb63785467a0d5e620c43760afe27b54f1d8)

lib/langlibs/go/lang.array/array.go
```diff
@@ -95,6 +96,56 @@ func initArrayModule(rt *runtime.Runtime) {
 	runtime.RegisterExternFunction(rt, orgName, moduleName, "push", func(ctx *extern.Context, args []values.BalValue) (values.BalValue, error) {
 		return arrayPush(ctx, args)
 	})
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "indexOf", func(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
```
warunalakshitha: shall we refactor the code to seperate functions than inline like [lib/langlibs/go/lang.map/map.go‎](https://github.com/ballerina-nutcracker/ballerina/pull/656/changes?file-filters%5B%5D=.go#diff-748c7971ef857bfa2860dee380b593e0b4caa0f09b5d44c524ebfdeca4fa1d54)

.gitattributes
```diff
@@ -1,2 +1,13 @@
 * text=auto eol=lf
 corpus/** linguist-generated=true
+
+# Golden output captures are compared byte-for-byte by the corpus test
+# harness against the interpreter's raw output, which can embed literal CR
+# bytes (e.g. multipart-body test fixtures). Line-ending normalization on
+# checkout/commit would silently corrupt those bytes, so exempt these
+# directories from text normalization entirely.
+corpus/ast/** -text
```
warunalakshitha, line 4: wont corpus/** linguist-generated=true work?

warunalakshitha: do we need this comment

values/list_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+package values
+
+import "testing"
+
+// TestListRemoveAtReleasesVacatedSlot checks that RemoveAt nils out the slot
+// it vacates instead of just shrinking the visible length, which would leave
+// the backing array holding a strong reference to the removed value forever
+// (as long as the list itself is reachable).
+func TestListRemoveAtReleasesVacatedSlot(t *testing.T) {
```
warunalakshitha, line 29: Why we add these native tests?

lib/langlibs/go/lang.string/string.go
```diff
@@ -63,6 +65,86 @@ func initStringModule(rt *runtime.Runtime) {
 	runtime.RegisterExternFunction(rt, orgName, moduleName, "fromBytes", func(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
 		return stringFromBytes(args)
 	})
+
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "substring", func(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
```
warunalakshitha: Shall we refactor these codes as well.

runtime/lifecycle.go
```diff
@@ -126,6 +126,15 @@ func (rt *Runtime) transition(target State) {
 	action := transitionTable[from][target]
 	rt.mu.Unlock()
 	if action == nil {
+		if from == StateStopped && (target == StateGracefulStopping || target == StateImmediateStopping) {
```
warunalakshitha: @heshanpadmasiri can u review this logic

runtime/lifecycle_test.go
```diff
@@ -366,6 +367,113 @@ service on l {
 	}
 }
 
+const startFailureLifecycleSource = `
+import ballerina/io;
```
warunalakshitha: cant we write these as corpus/bal tests

lib/langlibs/go/lang.string/string.go
```diff
@@ -63,6 +65,86 @@ func initStringModule(rt *runtime.Runtime) {
...
+		return strings.Trim(s, " \t\n\v\f\r"), nil
+	})
+
+}
+
+func equalsIgnoreCaseASCII(s1, s2 string) bool {
+	r1 := []rune(s1)
+	r2 := []rune(s2)
+	if len(r1) != len(r2) {
+		return false
```
warunalakshitha, line 121: shall we write bal tests for cover these uncovered lines as well

runtime/lifecycle.go
heshanpadmasiri: AFAIK in the spec sending a stop signal to a stopped runtime is undefined behaviour. I originally didn't supported this because I can't think of legitimate case where this can happen (once we stop the runtime process should get killed so user can't send signals). I assume this getting triggered in tests because tests with services we naively send the stop signal without checking if init/listen initialisation failed. 

Personally I don't have a strong opinion against allowing this; if it require other changes to runtime in order to allow test harness not to send the stop signal. Otherwise I think it should be fixed in the test harness

TharmiganK: Agree, this might not needed in this PR, will remove this. I faced it while testing tcp and udp tests in WASM CI build, let me check the reason again in tcp and udp implementation. This PR does not require this change

runtime/lifecycle_test.go
TharmiganK: This is removed now

runtime/lifecycle.go
TharmiganK: Removed via [3370d07](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/3370d07f1d5a93e077dfb66447be9d3a8bdf93cb)

values/list_test.go
TharmiganK: Initial implementation of List.RemoveAt/List.Clear used to shrink the visible length without clearing the backing array. These tests are added to guard against such changes in the future. This makes sure the `RemoveAt` nils out the slot, it vacates instead of just shrinking the visible length

.gitattributes
TharmiganK: > wont corpus/** linguist-generated=true work?

No, it is a different property. Here Line-ending normalization affects the byte vice comparison. That's why skipping that for the directories where we use byte wise comparison

> do we need this comment

No need, will remove that

lib/langlibs/go/lang.string/string.go
TharmiganK: Added via [6ab1809](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/6ab18093d759131e84636ff237fa50f6a0399bf2)

lib/langlibs/go/lang.string/string.go
TharmiganK: Fixed via [6ab1809](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/6ab18093d759131e84636ff237fa50f6a0399bf2)

.gitattributes
TharmiganK: Removed via [55c4886](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/55c488623be5be28329d781b350fc96a41c9ef24)

lib/langlibs/go/lang.array/array.go
TharmiganK: Fixed via: [6ab1809](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/6ab18093d759131e84636ff237fa50f6a0399bf2)

heshanpadmasiri:
I think runtime.sleep needs #683 (context.yeild) to properly yield the current strand.

TharmiganK:
> I think runtime.sleep needs #683 (context.yeild) to properly yield the current strand.

Ack, we could wait until the yield support to be merged and overwrite the runtime.sleep properly

lib/langlibs/go/lang.array/array.go
```diff
@@ -103,30 +108,62 @@ func arrayMap(ctx *extern.Context, args []values.BalValue) (values.BalValue, err
 	return values.NewList(resultTy, atomic, false, filler, 0, items), nil
 }
 
+// arrayIndexOf bails out before the int64->int conversion below: on a
+// 32-bit int platform (e.g. wasm) a large startIndex would truncate,
+// possibly to a negative value, and list.Get is unchecked.
```
heshanpadmasiri: Do we have a test case for this? I don't think it can happen given go internally uses int for indices anyway. I would imagine you to run out of memory long before reaching that limit

lib/langlibs/go/lang.string/string.go
```diff
@@ -47,22 +51,95 @@ func stringFromBytes(args []values.BalValue) (values.BalValue, error) {
...
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "equalsIgnoreCaseAscii", stringEqualsIgnoreCaseAscii)
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "toLowerAscii", stringToLowerAscii)
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "toUpperAscii", stringToUpperAscii)
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "trim", stringTrim)
+}
 
-	runtime.RegisterExternFunction(rt, orgName, moduleName, "toBytes", func(ctx *extern.Context, args []values.BalValue) (values.BalValue, error) {
-		return stringToBytes(byteArrTy, ctx, args)
-	})
+func equalsIgnoreCaseASCII(s1, s2 string) bool {
```
heshanpadmasiri, line 119: I think we can do better such that is is 0 allocation. See how equal fold is implemented in go standard library (https://cs.opensource.google/go/go/+/master:src/strings/strings.go;l=1210) difference is we don't treat non ascii characters to be case insensitive (should simplify this further)

lib/langlibs/go/lang.string/string.go
```diff
@@ -47,22 +51,95 @@ func stringFromBytes(args []values.BalValue) (values.BalValue, error) {
...
+	endIndex := args[2].(int64)
+	runes := []rune(s)
+	length := int64(len(runes))
+	if startIndex < 0 || startIndex > length || endIndex < startIndex || endIndex > length {
+		panic(values.NewErrorWithMessage(fmt.Sprintf("string index out of range: startIndex=%d endIndex=%d length=%d", startIndex, endIndex, length)))
+	}
+	return string(runes[startIndex:endIndex]), nil
+}
+
+func stringEqualsIgnoreCaseAscii(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
```
heshanpadmasiri: NIT: go naming convention is ASCII

lib/langlibs/go/lang.runtime/runtime.go
```diff
@@ -0,0 +1,44 @@
...
+const (
+	orgName    = "ballerina"
+	moduleName = "lang.runtime"
+)
+
+func initRuntimeModule(rt *runtime.Runtime) {
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "sleep",
+		func(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+			seconds, _ := args[0].(*decimal.Decimal)
+			rt.Platform().Time.Sleep(time.Duration(seconds.Float64() * float64(time.Second)))
```
heshanpadmasiri: I don't think we can push how to "sleep for at least T secons" to PAL, because this have some gurantees in the spec such that this is a yield point, and this (not entirely sure if gurantee this) is not going to burn cycles. I do belive we need a way in the pal to get notified after a given time has passed (may be over a channel).

values/list.go
```diff
@@ -130,6 +130,36 @@ func (l *List) Append(tc semtypes.Context, vs ...BalValue) {
 	l.elems = append(l.elems, vs...)
 }
 
+// RemoveAt removes and returns the element at idx, shifting subsequent elements down.
+// Panics if the list's inherent type is fixed-length.
+func (l *List) RemoveAt(idx int) BalValue {
+	l.checkMutable()
+	l.checkNotFixedLength()
```
heshanpadmasiri: Is this correct consider `[int, int, int...]` we can remove say 10th element right?

heshanpadmasiri: I think what you need is a function to check the fixed length using the atomic type.

values/list.go
```diff
@@ -130,6 +130,36 @@ func (l *List) Append(tc semtypes.Context, vs ...BalValue) {
...
+	l.elems[last] = nil // release the vacated slot so the GC can reclaim it
+	l.elems = l.elems[:last]
+	return val
+}
+
+// Clear removes all elements from the list.
+// Panics if the list's inherent type is fixed-length.
+func (l *List) Clear() {
+	l.checkMutable()
+	l.checkNotFixedLength()
```
heshanpadmasiri: Here you can check fixed length is 0

lib/langlibs/go/lang.string/string.go
TharmiganK: Fixed via [c50be74](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/c50be7411a2865d4e1883b5b4a1c7fb031c35ecb)

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: Introduce a channel based PAL interface like `Time.After` via [c50be74](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/c50be7411a2865d4e1883b5b4a1c7fb031c35ecb)

lib/langlibs/go/lang.string/string.go
TharmiganK: Fixed via [c50be74](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/c50be7411a2865d4e1883b5b4a1c7fb031c35ecb)

lib/langlibs/go/lang.array/array.go
TharmiganK: Removed via [c50be74](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/c50be7411a2865d4e1883b5b4a1c7fb031c35ecb)

values/list.go
TharmiganK: Fixed via [62cf5ca](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/62cf5caaedb3e589382ffbe823d3618d24fe346d).

values/list.go
TharmiganK: Properly fixed via [62cf5ca](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/62cf5caaedb3e589382ffbe823d3618d24fe346d)

Two checks done similar to jBallerina:
- Check whether we can shrink
- Check the type matches after removal

```bal
import ballerina/io;

    [int, int, int...] arr1 = [1, 2, 3];

    // Remove an element from rest values
    int|error val1 = trap arr1.remove(2);
    io:println(val1 is int); //@output: true

    // Remove an element from the fixed length
    // part of the array
    val1 = trap arr1.remove(1);
    io:println(val1 is error); //@output: true

    arr1.push(4);
    // Remove an element from the fixed type 
    // part of the array
    // Shifting does not violate the type
    val1 = trap arr1.remove(1);
    io:println(val1 is int); //@output: true

    [string, int, int...] arr2 = ["a", 1, 2];

    // Remove an element from the fixed type 
    // part of the array
    // Shifting violates the type
    string|int|error val2 = trap arr2.remove(0);
    io:println(val2 is error); //@output: true
}
```

values/bytes.go
```diff
@@ -27,11 +27,16 @@ func (l *List) ToByteSlice() []byte {
 	return b
 }
 
+// byteFiller is the filler value for byte[]. Supplying it keeps a converted
+// array growable, so a filling index assignment past its length works the same
+// as it does on any other byte[].
+func byteFiller() BalValue { return int64(0) }
```
warunalakshitha: why separate function. can we directly use int64(0)

lib/langlibs/go/lang.runtime/runtime.go
```diff
@@ -0,0 +1,44 @@
...
+const (
+	orgName    = "ballerina"
+	moduleName = "lang.runtime"
+)
+
+func initRuntimeModule(rt *runtime.Runtime) {
+	runtime.RegisterExternFunction(rt, orgName, moduleName, "sleep",
+		func(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+			seconds, _ := args[0].(*decimal.Decimal)
+			<-rt.Platform().Time.After(time.Duration(seconds.Float64() * float64(time.Second)))
```
warunalakshitha, line 37: Should we wait on https://github.com/ballerina-nutcracker/ballerina/pull/683? or may be we can add todo issue and proceed.

TharmiganK: Better to wait for #683

values/bytes.go
TharmiganK: Fixed

semantics/internal/types/type_resolver.go
```diff
@@ -7843,6 +7889,84 @@ func monomorphizeArrayPush(t typeResolver, sym *model.OpaqueFunctionSymbol, poly
...
+	if hasStartIndex {
+		arityKey = semtypes.Int
+	}
+	if sym.Lookup != nil {
+		if ref, ok := sym.Lookup(containerTy, arityKey); ok {
+			return ref, chain, true
+		}
+	}
+	cx := t.typeContext()
+	if !semtypes.IsSubtype(cx, containerTy, semtypes.List) {
```
warunalakshitha: We are checking any List.should that be anydata array. I think ,

error[] errs = [error("a")];
_ = errs.indexOf(error("a"));
class Person { }
Person p = new;
_ = [p].indexOf(p);

will work too

warunalakshitha: Check other places as well

TharmiganK: Fixed via [86a594e](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/86a594e2af8477d46e298cff51499415950d3995)

TharmiganK:
Making this as draft and refactor runtime sleep implementation based on #683

corpus/ast/subset10/10-langlibs/array-base-encode-v.txt
```diff
@@ -0,0 +1,38 @@
+(compilation-unit
```
heshanpadmasiri, line 1: Let's move new tests to subset 10

model/opaque.go
```diff
@@ -39,6 +39,11 @@ type OpaqueFunctionSymbol struct {
 	name        string
 	ID          int          // per-package opaque id; serialization handle and (with the package) selects the monomorphizer
 	SymbolSpace *SymbolSpace // space the monomorphized function is added to
+	// Params are the declared parameters, in order, so named arguments resolve
+	// the same way they do in jBallerina. They belong to the symbol rather than
+	// to a lookup keyed by function name, because the same name means different
+	// things in different packages (array:remove vs map:remove).
```
heshanpadmasiri: comment here is not entierly clear to me. I don't belive the stated issue (name collision in different packages) can happen and if that happens it is not clear to me how this can solve it.

semantics/internal/types/type_resolver.go
```diff
@@ -8298,6 +8344,88 @@ func monomorphizeArrayPush(t typeResolver, sym *model.OpaqueFunctionSymbol, poly
...
+		if ref, ok := sym.Lookup(containerTy); ok {
+			return ref, chain, true
+		}
+	}
+	cx := t.typeContext()
+	if !semtypes.IsSubtype(cx, containerTy, semtypes.List) {
+		t.semanticError("expect first argument to be a subtype of (any|error)[]", pos)
+		return model.SymbolRef{}, chain, false
+	}
+	elementType := semtypes.ListProj(cx, containerTy, semtypes.Int)
```
heshanpadmasiri, line 8414: I think you can also check if if the container type is readonly as well, and give compile time errors. You will need to have -p tests where you assign readonly lists to normal list and try to remove

values/bytes.go
```diff
@@ -27,11 +27,14 @@ func (l *List) ToByteSlice() []byte {
...
+// filler keeps a converted array growable, so a filling index assignment
+// past its length works the same as it does on any other byte[].
 func ByteSliceToList(byteArrTy semtypes.SemType, env semtypes.Env, data []byte) *List {
 	items := make([]BalValue, len(data))
 	for i, b := range data {
 		items[i] = int64(b)
 	}
-	return NewList(byteArrTy, semtypes.ToListAtomicType(env, byteArrTy), false, nil, 0, items)
+	filler := func() BalValue { return int64(0) }
+	return NewList(byteArrTy, semtypes.ToListAtomicType(env, byteArrTy), false, filler, 0, items)
```
heshanpadmasiri, line 39: NIT: I think we don't have any tests that try to grow the returned array to valdate this filler is working

lib/langlibs/go/lang.array/array.go
```diff
@@ -104,30 +109,61 @@ func arrayMap(ctx *extern.Context, args []values.BalValue) (values.BalValue, err
...
 }
 
+// arrayIndexOf returns nil (no match) once startIndex reaches the list's
+// length, before the loop below ever indexes into it.
+func arrayIndexOf(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	list := args[0].(*values.List)
+	val := args[1]
+	startIndex := int64(0)
+	if len(args) > 2 && args[2] != nil {
+		if si, ok := args[2].(int64); ok {
```
heshanpadmasiri: If !ok just panic

lib/langlibs/go/lang.runtime/runtime.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/ballerina-nutcracker/ballerina/values"
+)
+
+const (
+	orgName    = "ballerina"
+	moduleName = "lang.runtime"
+)
+
+func runtimeSleep(ctx *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	seconds, _ := args[0].(*decimal.Decimal)
```
heshanpadmasiri: if failed panic instead of ignoring

lib/langlibs/go/lang.runtime/runtime.go
```diff
@@ -0,0 +1,54 @@
...
+	dur := time.Duration(seconds.Float64() * float64(time.Second))
+	if dur <= 0 {
+		return nil, nil
+	}
+	// Yield before the actual wait so strands sharing this one's cooperative
+	// thread run while this strand sleeps, then wait for our turn to come
+	// back around once the timer fires.
+	continuation := ctx.Yield()
+	<-ctx.Env.Platform.Time.After(dur)
```
heshanpadmasiri: I am not super sure this is the correct way to do this. I would imagine something like

```
for delta < duration {
   <- ctx.Yield()
   update delta
}
```

Given you are not blocing on the channel that is returned by the runtime immediately that means effective two strands belonging to the same thread is continuing concurrently.

heshanpadmasiri (changes requested):

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: #683 merged, we're rebased on it now. `runtime:sleep` uses `ctx.Yield()` — see 6e84d289b.

corpus/ast/subset10/10-langlibs/array-base-encode-v.txt
TharmiganK: Moved, thanks.

model/opaque.go
TharmiganK: Fair, reworded in 0760fcbda. Concrete reason it's worth keeping on the symbol: `langlib-named-args-v.bal` calls `b.remove(index = 0)`(array) and `m.remove(k = "a")`(map) in the same file (lines 28 and 33) — that only resolves correctly because each symbol carries its own params. A name-keyed lookup would need the same package scoping duplicated to get that right.

semantics/internal/types/type_resolver.go
TharmiganK: Added in 5e80c1505, checked against `bal build` — jBallerina rejects `arr.remove(i)` at compile time when `arr`'s type is readonly. One thing: if a readonly value gets assigned to a non-readonly-typed variable first, that's a legal widening and can't be caught statically, so it still panics at runtime like before. Added a `-p` test for that case and an `-e` test for the compile-time one. Heads up — `push`, `removeAll`, and `map:remove`/`get` have the same gap in jBallerina (verified).

values/bytes.go
TharmiganK: Added a growth check to `string-bytes-v.bal`, checked against `bal run`.

lib/langlibs/go/lang.array/array.go
TharmiganK: Done, both spots (2efd24a35, 6e84d289b).

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: Done, both spots (2efd24a35, 6e84d289b).

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: Went with your suggestion — single `<-ctx.Yield()` per round, never held across other work. Found two things doing it though. A bare `for { <-ctx.Yield() }` spins hard when nothing else is queued, ~110% CPU for a 1s sleep. And under `go test` it actually starved a sibling strand from running for the whole sleep, flipping output order in my new test — reproduced that consistently, with and without `-race`. So now it waits 10ms then does one yield each round — CPU's basically nothing now and the ordering test's reliable (checked 5/5 plus a full `-race` run of the package). Also hit a bug in the test harness chasing this down: `MonotonicNow` was computing time since year 1, which overflows int64 and freezes at a constant. It's a pre-existing line, not something this PR introduced — but nothing had ever looped on it before, so it stayed latent until this did. Fixed separately in c3e0b621b.

model/opaque.go
heshanpadmasiri: I belive we don't support named args, and defaultable params with opaque function symbols. The correct way to handle this is to associate untyped function signatures (and typed one I don't recall whether it is needed or not) with opaque symbols such that it works just like other functions and methods. I haven't fully fleshed out how to do this either way it is orthogonal to the specific library being implemented. Please move the relevant tests to skiplist or -fv/-fe. I still don't understand this same package scoping story. Can you show the code sample. Either there is a bug in symbol resolution if so this is not fix (symbols are unambiguously identified by the import prefix  in case of function or expression type in case methods) or this is slop

lib/langlibs/go/lang.runtime/runtime.go
heshanpadmasiri: Of course a bare `for { <-ctx.Yield()}` is going to just spin the CPU doing nothing, that is by definition. Example in the comment is not a bare for loop. 

This implementation is wrong consider you have 3 strands on the same thread (per spec), I call runtime.sleep for for 50 seconds, this strand yeild passing execution to next strand which runs for 10 seconds yeild next one another 10 seconds yeilds but still `ctx.Env.Platform.Time.After(dur)` havent' completed (only 20 seconds), runtime can't pass execution to the other strands other strands given this haven't yield either so for next 30 seconds non of the strands in same thread can't progress this violates the contract in the api. You must repeatedly yield control back see how this is implemented in the runtime multiple wait

heshanpadmasiri (changes requested):

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: Fixed via [1ca51a8](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/1ca51a82cba5b10209f96885c05ec3aa525093eb)

model/opaque.go
TharmiganK: Reverted this via [ebaae38](https://github.com/ballerina-nutcracker/ballerina/pull/656/commits/ebaae38f028b8aec401947f20e8966a406d1bad4)

platform/pal/platform.go
```diff
@@ -82,6 +82,11 @@ type (
 	Time struct {
 		Now          func() time.Time
 		MonotonicNow func() time.Duration
+		// After returns a channel that receives once d has elapsed. It is a raw
+		// platform timer primitive, not a blocking sleep: callers decide how to
+		// wait on it (e.g. combined with a strand yield point) to honor the
+		// language spec's guarantees for runtime:sleep.
+		After func(d time.Duration) <-chan time.Time
```
heshanpadmasiri: NIT: I think this is no longer needed.

values/list_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"testing"
+
+	"github.com/ballerina-nutcracker/ballerina/semtypes"
+)
+
+// TestListRemoveAtReleasesVacatedSlot checks that RemoveAt nils out the slot
+// it vacates instead of just shrinking the visible length, which would leave
+// the backing array holding a strong reference to the removed value forever
+// (as long as the list itself is reachable).
+func TestListRemoveAtReleasesVacatedSlot(t *testing.T) {
```
warunalakshitha, line 29: Shall we make these tests parallel

semantics/internal/types/type_resolver.go
```diff
@@ -8252,6 +8257,11 @@ func storeMonomorphizedOpaqueFn(t typeResolver, sym *model.OpaqueFunctionSymbol,
 	return ref, true
 }
 
+// opaqueFunctionParams returns parameter names for an opaque function. Opaque
+// symbols carry no function signature of their own, so named arguments and
+// defaultable parameters are not supported for them; the names here exist only
+// for the ones that already had them. Attaching real (untyped) signatures to
+// opaque symbols is the proper fix and is tracked separately.
```
warunalakshitha, line 8283: why this new comment?

semantics/internal/symbols/symbol_resolver.go
```diff
@@ -907,6 +907,11 @@ func fillinOpaqueSymbol(sym model.Symbol, space *model.SymbolSpace) {
 	fn.Lookup, fn.Store = newMonomorphizationCache()
 }
 
+// newMonomorphizationCache builds a cache keyed by one or more semtypes (e.g.
+// a container type, plus any extra keys a monomorphizer needs to disambiguate
+// call sites that share a container type but resolve differently, such as by
+// arity). Each key is interned independently and looked up through a trie so
+// call sites can share prefixes of keys.
```
warunalakshitha: do we need this comment?

semantics/internal/types/type_resolver.go
warunalakshitha: opaqueFunctionParams("remove") uses map names m/k for array:remove. So shall we split cases by package/function: array remove → arr/index; keep map remove/get as m/k; add indexOf names too

test_util/testharness/test_harness.go
```diff
@@ -333,7 +338,8 @@ func (p *testPal) Platform() pal.Platform {
 		},
 		Time: pal.Time{
 			Now:          time.Now,
-			MonotonicNow: func() time.Duration { return time.Since(time.Time{}) },
+			MonotonicNow: func() time.Duration { return time.Since(testProcessStart) },
+			After:        time.After,
```
warunalakshitha: We can remove "After" here as well

lib/langlibs/ballerina/lang.array/0.0.1/any/lang.array.bal
```diff
@@ -20,18 +20,22 @@
...
 public isolated function length((any|error)[] arr) returns int = external;
 
-# Returns a base64 representation of a byte array.
+# Removes all members of an array.
 #
-# + arr - the byte array
-# + return - the base64 encoding of `arr`
-public isolated function toBase64(byte[] arr) returns string = external;
+# + arr - the array
+public isolated function removeAll((any|error)[] arr) = external;
```
warunalakshitha: remove is opaque (checked in the monomorphizer) while removeAll is a plain extern. Shall we fix it the same way remove does: make removeAll an opaque langlib function so the type resolver can see the real receiver type and reject readonly.

platform/pal/platform.go
TharmiganK: Removed in 1a843049d — dropped it from `pal.Time` and both constructors. `runtime:sleep` was the only consumer and it now polls `MonotonicNow` between yields.

test_util/testharness/test_harness.go
TharmiganK: Gone in the same commit (1a843049d) — removed the hook from `pal.Time` entirely, so both this and palnative's wiring went with it.

values/list_test.go
TharmiganK: Done in 86c30c7fe.

semantics/internal/types/type_resolver.go
TharmiganK: You're right — removed in da97f2c2e. Both functions are upstream's; I'd added those while resolving a rebase conflict and they just restate the signatures.

semantics/internal/symbols/symbol_resolver.go
TharmiganK: Same here — removed in da97f2c2e. Added while resolving a rebase conflict; the signature already says it.

semantics/internal/types/type_resolver.go
TharmiganK: Good catch, fixed in a64553a57 — but scoped by package only, not adding the real names.

`array:remove` was matching the case written for `map:remove`/`get` and inheriting `(m, k)`, so `b.remove(m = 0)` failed with a confusing "repeated values for parameter m"; it now correctly says "no such parameter m".

I left `array:remove`/`indexOf`/`removeAll` with unnamed params rather than giving them `arr`/`index`, because supplying real names makes `b.remove(index = 0)` compile — and the thread on `model/opaque.go` asked to descope named args for opaque symbols until they carry real signatures. `map` keeps `m`/`k` and still resolves. Happy to add the real names in the follow-up that attaches signatures.

lib/langlibs/ballerina/lang.array/0.0.1/any/lang.array.bal
TharmiganK: Done in 0534209d1 — `removeAll` is now an opaque langlib function with its own monomorphizer, so the type resolver sees the receiver type and rejects readonly at compile time like `remove` does. Verified against `bal build`, and added `array-removeall-readonly-e.bal`. Dropping the `.bal` declaration shifts the base BIR goldens by one atom, which is in the same commit.

warunalakshitha:
LGTM

lib/langlibs/go/lang.runtime/runtime.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	orgName    = "ballerina"
+	moduleName = "lang.runtime"
+)
+
+func runtimeSleep(ctx *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	seconds, ok := args[0].(*decimal.Decimal)
+	if !ok {
+		panic(fmt.Sprintf("internal error: unexpected seconds type %T", args[0]))
```
warunalakshitha: this will never happen right?

heshanpadmasiri (commented):
frontend changes LGTM

heshanpadmasiri (dismissed):

lib/langlibs/go/lang.array/array.go
```diff
@@ -104,30 +109,63 @@ func arrayMap(ctx *extern.Context, args []values.BalValue) (values.BalValue, err
...
+// arrayIndexOf returns nil (no match) once startIndex reaches the list's
+// length, before the loop below ever indexes into it.
+func arrayIndexOf(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	list := args[0].(*values.List)
+	val := args[1]
+	startIndex := int64(0)
+	if len(args) > 2 && args[2] != nil {
+		si, ok := args[2].(int64)
+		if !ok {
+			panic(fmt.Sprintf("internal error: unexpected startIndex type %T", args[2]))
```
warunalakshitha: can these be checked at compile time

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: Correct, it can't — `runtime:sleep("x")` is a semantic error, so the assertion always holds. Removed in 0d320ddbd; it's now an unchecked assertion like `args[0]` in the array externs, which still panics if the invariant is ever broken but doesn't carry a branch that can't run.

Sorry for the churn — I'd added the explicit check for your earlier comment on this line before checking whether the compiler already ruled it out.

lib/langlibs/go/lang.array/array.go
TharmiganK: Yes, already are — the monomorphizer types `startIndex` as `int`, so `a.indexOf(2, "x")` fails at compile time and the `!ok` branch was unreachable. Dropped it in 0d320ddbd in favour of an unchecked assertion.

The `startIndex < 0` panic just below is a different case though: it's a genuine runtime value, not a type, so it can't be compile-checked in general (`a.indexOf(2, -1)` with a computed index reaches the runtime). jBallerina panics there too, and `array-indexof-negative-start-p.bal` covers it — leaving that one as is.

warunalakshitha (approved):
LGTM

PR #664, opened 2026-07-23
feat(pal): shared net/tls PAL for non-HTTP socket protocols

TharmiganK:
Depends on: https://github.com/ballerina-nutcracker/ballerina/pull/656

## Summary

Adds a `pal.Net` surface (`Dial`/`Listen`/`DialPacket`/`ListenPacket`) to the Platform Adaptation Layer for raw TCP/UDP(+TLS) socket access, alongside a `tls.go` that extracts the TLS-config assembly previously inlined in `NewHTTPClient` so it's shared between the HTTP client and the new raw socket path.

- `pal.Net.Dial` — dial a TCP connection, optionally bound to a local address, optionally upgraded to TLS before returning
- `pal.Net.Listen` — bind a TCP listener, optionally wrapped in TLS; caller owns the accept loop
- `pal.Net.DialPacket` / `pal.Net.ListenPacket` — connected/unconnected datagram (UDP) sockets
- `buildTLSConfig` extracted out of `NewHTTPClient` into `tls.go` so both HTTP and the new socket path share one TLS-config assembly path
- `tlsVerifyConnectionWithCNFallback` now takes an explicit `expectedServerName` instead of reading `tls.ConnectionState.ServerName`, since Go never sends SNI for IP-literal dials — hostname verification would otherwise always fail against a bare IP address

## Why

This is the base branch for the `ballerina/tcp`, `ballerina/udp`, and `ballerina/ldap` stdlibs — all non-HTTP wire protocols that need raw socket access rather than `pal.HTTP`. Landing the PAL surface on its own keeps those stdlib PRs focused on protocol logic instead of platform plumbing, and lets them share one TLS-config implementation instead of each reinventing it.

## Note on test coverage

This PR only introduces the `pal.Net` interface and its native-CLI implementation — there are no `tcp`/`udp`/`ldap` stdlib callers yet, so there's nothing to drive it through corpus tests. Coverage is intentionally deferred to the `ballerina/tcp`, `ballerina/udp`, and `ballerina/ldap` implementation PRs, which will exercise `Dial`/`Listen`/`DialPacket`/`ListenPacket` (and the shared TLS path) end-to-end via their own corpus tests.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...`

## Summary by CodeRabbit

- **New Features**
  - Added native networking support for TCP, TLS, and datagram connections, including dialing and listening.
  - Added support for configuring TLS connections with custom certificates, client authentication, protocol versions, and cipher suites.

- **Bug Fixes**
  - Improved TLS hostname verification, including safer handling when certificates lack subject alternative names.
  - Invalid TLS certificate configuration now reports an error instead of silently falling back.
  - Prevented invalid or excessively large sleep durations from causing runtime errors or unexpected behavior.

### DIFF
--- a/lib/langlibs/go/lang.runtime/runtime.go
+++ b/lib/langlibs/go/lang.runtime/runtime.go
@@ -17,6 +17,7 @@
 package langruntime
 
 import (
+	"math"
 	"time"
 
 	"github.com/ballerina-nutcracker/ballerina/decimal"
@@ -32,7 +33,7 @@ const (
 
 func runtimeSleep(ctx *extern.Context, args []values.BalValue) (values.BalValue, error) {
 	seconds := args[0].(*decimal.Decimal)
-	dur := time.Duration(seconds.Float64() * float64(time.Second))
+	dur := secondsToSleepDuration(seconds.Float64())
 	deadline := ctx.Env.Platform.Time.MonotonicNow() + dur
 	// Hand the thread back on every pass, the same way the wait actions poll
 	// (see waitAllFutures). Blocking on a timer between yields would hold this
@@ -46,6 +47,23 @@ func runtimeSleep(ctx *extern.Context, args []values.BalValue) (values.BalValue,
 
 func initRuntimeModule(rt *runtime.Runtime) {
 	runtime.RegisterExternFunction(rt, orgName, moduleName, "sleep", runtimeSleep)
+}
+
+// secondsToSleepDuration converts a duration given in seconds to a
+// time.Duration, clamping non-finite and out-of-int64-range results
+// (reachable from decimal's much wider value range) to the max
+// representable duration instead of relying on the platform-specific
+// float-to-int64 overflow behavior. Non-positive and NaN durations pass
+// through as a no-op, same as time.Sleep's own documented behavior.
+func secondsToSleepDuration(seconds float64) time.Duration {
+	nanos := seconds * float64(time.Second)
+	if math.IsNaN(nanos) || nanos <= 0 {
+		return 0
+	}
+	if nanos >= math.MaxInt64 {
+		return math.MaxInt64
+	}
+	return time.Duration(nanos)
 }
 
 func init() {
--- a/platform/pal/platform.go
+++ b/platform/pal/platform.go
@@ -26,6 +26,7 @@ package pal
 import (
 	"context"
 	"io"
+	"net"
 	"net/http"
 	"time"
 )
@@ -54,6 +55,7 @@ type (
 		OS      OS
 		Time    Time
 		HTTP    HTTP
+		Net     Net
 		Signals SignalSource
 	}
 	IO struct {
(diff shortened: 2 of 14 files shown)


lib/langlibs/go/lang.runtime/runtime.go
```diff
@@ -48,6 +49,23 @@ func initRuntimeModule(rt *runtime.Runtime) {
...
+// representable duration instead of relying on the platform-specific
+// float-to-int64 overflow behavior. Non-positive and NaN durations pass
+// through as a no-op, same as time.Sleep's own documented behavior.
+func secondsToSleepDuration(seconds float64) time.Duration {
+	nanos := seconds * float64(time.Second)
+	if math.IsNaN(nanos) || nanos <= 0 {
+		return 0
+	}
+	if nanos >= math.MaxInt64 {
+		return math.MaxInt64
```
warunalakshitha, line 64: shall we add a test for this

platform/pal/platform.go
```diff
@@ -83,6 +85,27 @@ type (
 		Now          func() time.Time
 		MonotonicNow func() time.Duration
 	}
+	// Net abstracts raw TCP(+TLS) socket dialing and listening for non-HTTP
+	// wire protocols (e.g. ldap, tcp). HTTP-based stdlibs must keep using
+	// pal.HTTP instead.
+	Net struct {
```
warunalakshitha, line 91: Lets get stdlib team feedback on API as well, naming etc.

warunalakshitha:
Shall we add some tests for pal native impl

platform/pal/platform.go
TharmiganK: @daneshk Can you please review?

TharmiganK:
> Shall we add some tests for pal native impl

The tests for net PAL will be added with the tcp, udp, ldap library implementation

lib/langlibs/go/lang.runtime/runtime.go
TharmiganK: Added `TestSecondsToSleepDuration` in `lib/langlibs/go/lang.runtime/runtime_test.go` (commit 35908ff56). It's a plain Go unit test rather than a corpus test since the overflow/clamp branch would need the sleep to actually run to completion (~292 years) to observe end-to-end through a corpus test. Covers typical values, zero/negative/NaN no-ops, and the overflow paths (+Inf, a decimal-range value far outside float64) that motivated the fix.

platform/pal/platform.go
```diff
@@ -83,9 +85,33 @@ type (
 		Now          func() time.Time
 		MonotonicNow func() time.Duration
 	}
+	// Net abstracts raw TCP(+TLS) socket dialing and listening for non-HTTP
+	// wire protocols (e.g. ldap, tcp). HTTP-based stdlibs must keep using
+	// pal.HTTP instead.
```
heshanpadmasiri, line 90: Out of curiosity (I vaguely remember that we had this conversation at the beginning of http but I can't recall the conclusion) what is the rational behind this, if I am not mistaken what http has is an abstraction on top of this not something orthogonal to this.

TharmiganK: Yes, that is correct, HTTP abstraction is on top of this Net abstraction, and it automatically handles socket connection and lifecycle, SSL handshake, HTTP versions and request building. So the HTTP PAL interface is just build on top of the HTTP request, not plain sockets and data frames

platform/palnative/http.go
```diff
@@ -323,15 +294,15 @@ func tlsVerifyConnectionWithCNFallback(rootCAs *x509.CertPool) func(tls.Connecti
...
 		}
-		// cs.ServerName is the SNI hostname (no port). Try SAN-based verification first.
-		// Only fall back to CN matching for certs that genuinely have no SANs — when SANs
-		// are present but don't match, that is a real mismatch and must not be bypassed.
+		// Try SAN-based verification first. Only fall back to CN matching for
+		// certs that genuinely have no SANs — when SANs are present but don't
+		// match, that is a real mismatch and must not be bypassed.
 		leaf := cs.PeerCertificates[0]
-		if err := leaf.VerifyHostname(cs.ServerName); err != nil {
+		if err := leaf.VerifyHostname(expectedServerName); err != nil {
```
warunalakshitha, line 301: This works for pal.Net.Dial since IP literals don't send SNI, but for HTTP we usually don't set secureSocket.serverName — transport fills it from the host. Looks like new ("https://example.com", {secureSocket: {cert: "ca.pem"}}) will now verify against empty name and fail. Shall we fall back to cs.ServerName when expectedServerName is empty? Also shall we add a NewHTTPClient test for custom CA without serverName? tcp/udp/ldap tests won't cover this HTTP path.

platform/palnative/tls.go
```diff
@@ -0,0 +1,78 @@
...
+		if !pool.AppendCertsFromPEM(cfg.CACertPEM) {
+			return nil, fmt.Errorf("failed to parse CA certificate PEM: no valid certificates found")
+		}
+		tlsConfig.RootCAs = pool
+		if !cfg.InsecureSkipVerify {
+			// Go 1.15+ requires SANs for hostname verification; many self-signed and
+			// Java-issued certs only set the CN field. When a custom CA is provided
+			// we do our own verification so CN-only certs are accepted as a fallback.
+			tlsConfig.InsecureSkipVerify = true //nolint:gosec
+			tlsConfig.VerifyConnection = tlsVerifyConnectionWithCNFallback(pool, cfg.ServerName)
```
warunalakshitha, line 51: Same. tlsVerifyConnectionWithCNFallback(pool, cfg.ServerName) captures the PAL config's ServerName, which is empty for a typical http:Client. Dial already fills ServerName from the address before calling this, HTTP does not. The fallback to cs.ServerName belongs here (or in the callback).

platform/palnative/net.go
```diff
@@ -0,0 +1,118 @@
...
+			effectiveTLSCfg.ServerName = address
+		}
+	}
+	tlsConfig, err := buildTLSConfig(effectiveTLSCfg)
+	if err != nil {
+		_ = conn.Close()
+		return nil, err
+	}
+	tlsConn := tls.Client(conn, tlsConfig)
+	if err := tlsConn.HandshakeContext(ctx); err != nil {
```
warunalakshitha, line 70: Dial ignores handshake timeout from the shared TLS config.

platform/palnative/net.go
```diff
@@ -0,0 +1,118 @@
...
+		_ = conn.Close()
```
warunalakshitha, line 71: Handshake error path does not close the tls.Conn. Shall we close

PR #670, opened 2026-07-24
perf: cut per-request allocations on the HTTP request/response path

### LINKED ISSUE #671: Cut per-request allocations across the HTTP request/response path
## Problem

Under core scarcity the interpreter's HTTP throughput trails a hand-written Go `net/http` server, and allocation profiling of the `hello` and `passthrough` benchmark services shows why: the gap is **CPU-per-request driven by rebuilding program-constant or otherwise-reusable state on every request** — not `net/http` vs Netty, and not lock contention (a mutex-profile lead there turned out to be a measurement artifact once verified with a release-build A/B).

## Allocation sources found

Profiling `hello` and `passthrough` (small and large payloads) attributes the bulk of per-request allocation to state that is identical across requests yet rebuilt each time:

- **`extern.Context`** — each dispatched invocation builds a fresh semtype context (8 memo maps + an interner) keyed on the program-constant `TypeEnv`.
- **Request/Response method-key maps** — a constant `{method: lookup-key}` map rebuilt while constructing each inbound `Request` / native `Response`.
- **Object method & resource tables** — `execNewObject` rebuilds a class's method-key map and deep-copies its resource table on every `new`, though both derive purely from the class definition and are never mutated after construction (the #1 allocator on `hello`, ~26% of bytes).
- **Inbound request headers** — materialized eagerly into a Ballerina `Map`/`List` graph even when the resource never reads a header.
- **Constant `MAPPING`/`LIST` atomic types** — `ToMappingAtomicType`/`ToListAtomicType` allocate a fresh copy per header map/list, though the type is program-constant.
- **Response-streaming copy buffer** — streaming a backend response to the client allocates a fresh 32 KB buffer per request (net/http's `ReadFrom` fallback, since kernel splice is unavailable for a userspace stream); dominates allocations at large payloads (~77–83% of bytes at 10–100 KB).

## Proposal

Precompute or pool each source, without changing observable behaviour: pool and reuse the context; build the method-key and class method/resource tables once and share them read-only; materialize request headers lazily; cache the constant atomic types; and pool the response-streaming copy buffer.

## Impact

Per-request allocation and GC pressure fall substantially across both scenarios — throughput rises under core contention (`hello`) and GC pressure / tail latency drop on core-constrained hosts (`passthrough`, especially at larger payloads). Measured numbers are in the PR.

TharmiganK:
Fixes: https://github.com/ballerina-nutcracker/ballerina/issues/671

## Purpose

Cut per-request allocations across the HTTP path so the interpreter's throughput sits closer to a hand-written Go server under core scarcity, and so GC pressure (and the latency tail) drop on core-constrained hosts. Profiling the `hello` and `passthrough` benchmark services showed the gap to plain Go / Spring Boot is CPU-per-request under contention — not `net/http` vs Netty, and not lock contention. Allocation profiles pinned the cost on program-constant or otherwise-reusable state that was rebuilt on every request.

### Context & call stack

1. **Pool and reuse the extern context per invocation.** Each request created a fresh `extern.Context`, reallocating a whole semtype context (8 memo maps + interner) even though those caches are keyed on the program-constant `TypeEnv`. The context is now pooled on the `Runtime` and reused: per-request state is reset (fresh strand id, cleared held-lock stack, truncated call stack) while the semtype caches stay warm. The http `Listener` releases the context at the end of `dispatchRequest` (response already written, resource ran synchronously on this strand).
2. **Clear the pooled call-stack backing array on reuse.** A pooled context outlives the request, so leftover `callStackEntry` slots kept request-local `*Frame` graphs reachable and unreclaimable until a deeper later request overwrote each slot. Zero the backing array before truncating.

### Object construction

3. **Cache the constant Request/Response method-key maps.** `requestMethodKeys`/`responseMethodKeys` rebuilt a constant `{method: lookup-key}` map on every call — allocated per request while building the inbound `Request` and native `Response`. The maps never change and `values.Object` only reads them, so they are built once at package load and shared read-only.
4. **Share immutable method/resource tables across object instances.** `execNewObject` rebuilt a class's method-key map and deep-copied its resource table on every `new`, though both derive purely from the class definition and are only ever read. Precompute them once per class at registration (a shared `ClassTemplate`) and reference them from every instance; object creation now allocates only the per-instance field map. This was the number one allocator on `hello`.

### Request path

5. **Build inbound request headers lazily.** The server converted every request header into a Ballerina `Map`/`List` value graph up front, even for resources that never read a header. Store the raw header map in a holder and materialize the Ballerina view only on first access (cached back under `$headers`); the client-forward and message-to-body paths read the raw map directly. Header-ignoring handlers now allocate nothing for headers.
6. **Cache the constant header `Map`/`List` atomic types.** `newMappingValue`/`newListValue` called `ToMappingAtomicType`/`ToListAtomicType(MAPPING|LIST)` per request; for those program-constant types the function ignores the type context and returns a freshly heap-allocated copy each call. Compute once and share the pointer (`Map`/`List` only read it).

### Response path

7. **Pool the copy buffer when streaming response bodies.** `writeStream` streamed the backend response to the client with `io.Copy`. Because the source is a userspace stream (no kernel splice), net/http's `response.ReadFrom` allocated a fresh 32 KB buffer on every request — the dominant allocation at large payloads. Copy through a pooled buffer instead, hiding the `ResponseWriter`'s `ReadFrom` (`onlyWriter`) so `io.CopyBuffer` actually uses it; output framing is unchanged.

## Performance

Measured across two campaigns (different hardware/conditions — noted per table).

**Context pooling + method-key cache (changes 1, 3) — local, `GOMAXPROCS=2`:**

| Scenario | Before | After | Δ |
|---|---|---|---|
| `hello` | ~88.5k req/s | ~95.3k req/s | **+8%** |
| `passthrough` | ~41.6k req/s | ~45.1k req/s | **+8.5%** |

This closes the `hello` gap to a hand-written Go `net/http` server from ~9% to ~1–2%.

**Object + request-path fixes (changes 4–6) — AWS EC2 m6a.xlarge (4 vCPU), `hello`, 100/200/500 users:**

| Metric | Before | After | Δ |
|---|---|---|---|
| Throughput | 56.7 / 57.4 / 55.7k req/s | 63.2 / 62.7 / 60.5k req/s | **+9–11%** |
| p99 latency | 21 / 31 / 62 ms | 17 / 28 / 52 ms | **−12 to −18%** |
| Total allocations (30-min run) | — | — | **−22%** |
| GC cycles | 1,622 | 1,245 | **−23%** |

Change 4 (`execNewObject`) alone dropped that call site from ~26% of allocated bytes to ~1%.

**Response copy-buffer pooling (change 7) — `passthrough`, allocation/GC:**

| Payload | Total allocations | GC CPU fraction |
|---|---|---|
| 10 KB | **−61%** | **−62%** |
| 100 KB | **−40%** | **−23%** |

Throughput is ≈flat here — at these payload sizes `passthrough` is bounded by moving the bytes through userspace (syscalls/`memmove`), which no runtime-side change escapes short of kernel splice; the win is GC pressure and the tail.

## Concurrency safety

- **Context pooling:** `sync.Pool` gives exclusive ownership, so a context and its caches are only ever touched by one strand at a time — no locking. Async work is safe because `StartMethod` snapshots frames by value and `runStrand` builds each started strand its own context, so a started strand never aliases the parent's; only the shared `Env` is retained, which every context already shares.
- **Method-key maps & class method/resource tables:** built once and shared read-only; `values.Object` never mutates them.
- **Atomic-type cache:** the atomic type is only ever read by `Map`/`List`, so a single shared pointer across instances is safe.
- **Copy-buffer pool:** each buffer is exclusively owned by one request between `Get`/`Put`.

## Testing

- http native unit tests, `corpus/extern` (service, client, forward, **graceful stop**), and the full `TestIntegration` corpus pass; the context-pooling change was additionally verified with `go test -race`.
- Request isolation verified under load: 300–500 concurrent unique-body `passthrough` requests → 0 body mismatches.
- Allocation profiles confirm each targeted allocator no longer appears (or drops sharply) in per-request allocations: `ContextFrom`/`CreateContext`, the method-key maps, `execNewObject`'s tables, the eager header `Map`/`List`, `ToMappingAtomicType`/`ToListAtomicType`, and the streaming copy buffer.

## Summary by CodeRabbit

* **Performance**
  * Improved HTTP request handling by deferring header materialization and reducing per-request overhead.
  * Improved streaming response throughput using pooled buffers.
  * Reduced repeated method lookup and class metadata work during dispatch.
* **Bug Fixes**
  * Improved consistency in type conversions across collections, bytes, XML, and crypto operations.
  * Updated HTTP request/forwarding paths so content-type extraction and forwarded headers match the same source.
* **Reliability**
  * Improved pooled runtime execution context reuse with safer reset behavior and added coverage for memo-cache clearing.


warunalakshitha:
Shall we create an issue for this? add M6

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go
```diff
@@ -619,7 +620,7 @@ func buildRequest(tc semtypes.Context, method, rawPath, httpVersion string, head
 			"rawPath":     rawPath,
 			"method":      method,
 			"httpVersion": httpVersion,
-			"$headers":    headersMap,
+			"$headers":    &lazyRequestHeaders{raw: headers},
```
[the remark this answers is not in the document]
TharmiganK: headers is always r.Header from the real *http.Request (never nil in practice), and even if it were, Go's len()/range over a nil map are safe no-ops — materializeHeaders and rawRequestHeaders (http_native.go:240-306) never panic on nil, and nothing mutates lazyRequestHeaders.raw in place

runtime/extern/extern.go
```diff
@@ -77,6 +77,16 @@ func CreateContext(env *Env) *Context {
 	return &ctx
 }
 
+// ResetForReuse clears a context's per-request mutable state — a fresh strand
+// ID and an empty held-lock stack — while retaining the constant Env and the
+// warm TypeCtx memo caches (keyed on the program-constant TypeEnv). It lets a
+// pooled context serve a fresh request without reallocating the semtype
+// context. Safe only under exclusive ownership (e.g. via sync.Pool).
+func (ctx *Context) ResetForReuse() {
```
heshanpadmasiri, line 109: You need to reset the semtype context as well. Otherwise it's caches will grow unboundedly

runtime/internal/modules/registry_impl.go
```diff
@@ -20,11 +20,24 @@ import (
...
 	"ballerina/runtime/extern"
+	"ballerina/values"
 )
 
+// ClassTemplate holds the per-class data that is identical for every instance
+// of a class and never mutated after construction: the method-key map and the
+// resource table. It is computed once when a class is registered and shared by
+// reference across all instances, so object creation only allocates the
+// per-instance field map. FieldCount pre-sizes that map.
+type ClassTemplate struct {
```
heshanpadmasiri, line 31: Can't we directly hold on to this instead of BIRClassDef?

heshanpadmasiri: Ideally registry should hold on to a single construct that can hold the class

heshanpadmasiri (changes requested):
So the idea is to push responsibility of handling context to services correct? For this to work they should always return the context back to pool.

On principle I am fine with that. I think we should have 2 separate function for creating context in that case *Pooled and normal. In general for cases where we create strands runtime/calling API wont know when to return context back to pool so always going through pool is just overhead for them. I think service resource/remote functions are a special case because you know when you are done executing one request. 

Also do we know for common cases if we really need the semtype context? (unlike jBallerina we don't trigger type checks for every function call so I think it is possible to write some programs potentially without needing to do any type check). In that case I would lazy initialize it in context (replace direct field access with a public function) and give anther API to directly get type env without having to go via type context.

runtime/extern/extern.go
TharmiganK: Fixed via [da8385e](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/da8385e191f81c1f19912e133d4eeee5c424b03a)

runtime/internal/modules/registry_impl.go
TharmiganK: Fixed via [edcade2](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/edcade27ebcd7d18f46e67a158339da081e455db)

TharmiganK:
> So the idea is to push responsibility of handling context to services correct? For this to work they should always return the context back to pool.
> 
> On principle I am fine with that. I think we should have 2 separate function for creating context in that case *Pooled and normal. In general for cases where we create strands runtime/calling API wont know when to return context back to pool so always going through pool is just overhead for them. I think service resource/remote functions are a special case because you know when you are done executing one request.
> 
> Also do we know for common cases if we really need the semtype context? (unlike jBallerina we don't trigger type checks for every function call so I think it is possible to write some programs potentially without needing to do any type check). In that case I would lazy initialize it in context (replace direct field access with a public function) and give anther API to directly get type env without having to go via type context.

Good points, both addressed.

**Pooled vs. normal:** split it into `Runtime.AcquirePooledContext/ReleasePooledContext` (must always be paired — that's now explicit in the name) versus the existing `extern.CreateContext/exec.CreateContext` for everything else. HTTP resource/remote dispatch is the only caller of the pooled path (`dispatchRequest`, released via `defer` once the response is written). `StartMethod`-spawned strands, `InvokeFunction`, and the lifecycle hooks all go through the unpooled path since none of them have a single clear release point — doc comments on both now spell out which to use when.

**Lazy semtype context:** agreed we don't always need it. `Context.TypeCtx` is now a private field built lazily behind a `TypeCtx()` method — only strands that actually call it pay for the memo caches. Added `TypeEnv()` too, for the couple of call sites that only needed the type environment and were paying for a full type-check context for nothing.

One thing this opened up: since a pooled context can go a long time without ever building a `TypeCtx`, but the ones that do build one keep it warm across reuses, I also added `Reset()` to clear its memo caches on release so they don't grow unbounded over the context's pooled lifetime — invocations that never type-check still pay nothing.

lib/langinternal/runtime/internal.go
```diff
@@ -117,7 +117,7 @@ type queryGroupState struct {
 type queryGroupIndex map[string]int
 
 func newQueryList(ctx *extern.Context) *values.List {
-	return values.NewList(semtypes.LIST, semtypes.ToListAtomicType(ctx.TypeCtx, semtypes.LIST), false, nil, 0, nil)
+	return values.NewList(semtypes.LIST, semtypes.ToListAtomicType(ctx.TypeCtx(), semtypes.LIST), false, nil, 0, nil)
```
heshanpadmasiri: I think semtypes.To*AtomicType functions should now take type Env instead of Context (thy internally just use context to get env any). Otherwise creating these values unnecessarily allocate a type context

lib/langlibs/go/lang.array/array.go
```diff
@@ -54,7 +54,7 @@ func arrayFromBase64(byteArrTy semtypes.SemType, ctx *extern.Context, args []val
 	if err != nil {
 		return values.NewErrorWithMessage("failed to decode base64 string"), nil
 	}
-	return values.ByteSliceToList(byteArrTy, ctx.TypeCtx, data), nil
+	return values.ByteSliceToList(byteArrTy, ctx.TypeCtx(), data), nil
```
heshanpadmasiri: I think after To*AtomicType change these can take the env directly as well

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
```diff
@@ -192,28 +192,123 @@ func eagerBufferResponse(respHeaders map[string][]string, bodyStream io.ReadClos
 }
 
 // newMappingValue builds a fresh open map<anydata|error> value.
+// The header/request maps and value lists built per request all use the
+// program-constant MAPPING / LIST inherent types, whose atomic type does not
+// depend on the type context. ToMappingAtomicType/ToListAtomicType allocate a
+// fresh copy on every call, so cache the results once and reuse the shared
+// pointers — Map/List only read the atomic type, never mutate it, so sharing is
+// safe.
+var (
```
heshanpadmasiri, line 202: Do we need to do this lazily per request. I think we can create this when registering with runtime and just reuse it without sync.Once

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
```diff
@@ -192,28 +192,123 @@ func eagerBufferResponse(respHeaders map[string][]string, bodyStream io.ReadClos
...
+	}
+	return headersMap
+}
+
+// requestHeadersMap returns the request's header Map and whether headers are
+// present, materializing a lazy inbound-header holder on first access and
+// caching the result back under "$headers". Objects that already hold a
+// *values.Map (outbound requests built via `new`) are returned unchanged.
+func requestHeadersMap(tc semtypes.Context, self *values.Object) (*values.Map, bool) {
+	hv, ok := self.Get("$headers")
```
heshanpadmasiri, line 258: I assume we can get rid of this with https://github.com/ballerina-nutcracker/ballerina/issues/638

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
heshanpadmasiri: With that some of the APIs should no longer need a type context as well

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go
```diff
@@ -526,7 +532,7 @@ func dispatchRequest(rt *runtime.Runtime, state *listenerState, w http.ResponseW
 			writeErrorJSON(rt, w, r, http.StatusInternalServerError, err.Error())
 			return
 		}
-		writeResult(rt, ctx.TypeCtx, w, r, result)
+		writeResult(rt, ctx.TypeCtx(), w, r, result)
```
heshanpadmasiri: I don't this writeResult needs a semtypeContext (currently it is `_`)

runtime/extern/extern.go
```diff
@@ -71,12 +71,49 @@ func (e *Env) AllocateStrandID() uint64 {
...
+
+// ResetForReuse clears a context's per-request mutable state — a fresh strand
+// ID, an empty held-lock stack, and (if one was ever built) a cleared TypeCtx
+// — while retaining the constant Env. Clearing TypeCtx rather than keeping its
+// memo caches warm bounds its memory across the pooled context's lifetime;
+// combined with TypeCtx's lazy construction, an invocation that never
+// type-checks pays nothing here either. Safe only under exclusive ownership
+// (e.g. via sync.Pool).
+func (ctx *Context) ResetForReuse() {
+	ctx.StrandID = ctx.Env.AllocateStrandID()
```
heshanpadmasiri, line 110: I think you can simply set new StrandID and set everything except CallStack and Env(Env don't change) to be nil. When we get from the pool we can set the CallStack appropriatly

semtypes/context.go
```diff
@@ -199,6 +199,35 @@ func (c *context) resetConjunctionStack(depth int32) {
 	c._conjunctions = c._conjunctions[:depth]
 }
 
+// Reset clears every memo cache and interner entry, leaving the context as if
```
heshanpadmasiri, line 184: I think it is simpler to just drop it rather than try to reset this

TharmiganK: Fixed via [b164ad9](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/b164ad9943ecee40b6b27af939fe5380aea7933a)

lib/langinternal/runtime/internal.go
TharmiganK: Fixed via [b164ad9](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/b164ad9943ecee40b6b27af939fe5380aea7933a)

lib/langlibs/go/lang.array/array.go
TharmiganK: Fixed via [b164ad9](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/b164ad9943ecee40b6b27af939fe5380aea7933a)

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
TharmiganK: Fixed via [24163be](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/24163be543bd0c9c2ce07c06fed72392f23fd307)

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go
TharmiganK: Fixed via [24163be](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/24163be543bd0c9c2ce07c06fed72392f23fd307)

runtime/extern/extern.go
TharmiganK: Fixed via [6bd3ac8](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/6bd3ac89040e7d5678273bfa5c9ec983c8a3cfaa)

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
TharmiganK: Yes, we should be able to

heshanpadmasiri (dismissed):
LGTM

semtypes/context.go
```diff
@@ -199,6 +199,35 @@ func (c *context) resetConjunctionStack(depth int32) {
...
+// freshly built from the same Env via ContextFrom. Intended for a pooled
+// context that is about to be reused by an unrelated invocation, so cached
+// entries don't accumulate without bound over the context's pooled lifetime.
+func (c *context) Reset() {
+	c._memoStack = c._memoStack[:0]
+	c._conjunctions = c._conjunctions[:0]
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [804a597](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/804a597bb5b570aad8a7222f80d7fbae1ee37631)

runtime/extern/extern.go
```diff
@@ -71,12 +71,49 @@ func (e *Env) AllocateStrandID() uint64 {
...
+	ctx.heldLocks = ctx.heldLocks[:0]
+	if ctx.typeCtx != nil {
+		ctx.typeCtx.Reset()
```
heshanpadmasiri, line 113: I think a more straightforward solution is to simply drop it.  With lazy initialization change for strands that don't do any type checks we shouldn't create type context anyway

TharmiganK: In the dispatching logic we do type checks right? IIRC I tried dropping this and there was a significant drop in the performance

TharmiganK: With this change where we simply dropping it:

## HTTP Benchmark — hello service

| Metric | `origin/perf-runtime-optimizations` (base) | `HEAD` (head) | Δ (head vs base) |
|---|---|---|---|
| Throughput (req/s) | 56,760 (±271) | 51,703 (±409) | -8.9% |
| Avg latency (ms) | 1.05 | 1.15 | +10.0% |
| p99 latency (ms) | 5.01 | 5.43 | +8.3% |
| Startup (s) | 0.051 | 0.051 | -0.1% |
| Peak RSS (MB) | 41.8 | 41.5 | -0.8% |

In AWS setup, it was around 10% drop in throughput

heshanpadmasiri: Okay then let's keep it

heshanpadmasiri (dismissed):
LGTM

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
```diff
@@ -192,32 +192,118 @@ func eagerBufferResponse(respHeaders map[string][]string, bodyStream io.ReadClos
...
+	if !ok {
+		return nil, false
+	}
+	switch h := hv.(type) {
+	case *values.Map:
+		return h, true
+	case *lazyRequestHeaders:
+		m := materializeHeaders(tc, h.raw)
+		self.Put("$headers", m)
```
warunalakshitha: shouldn't we use sync once here?

TharmiganK: Fixed via [fd1959f](https://github.com/ballerina-nutcracker/ballerina/pull/670/commits/fd1959f2fe0ace839da20ce3a5d9a6e9591a24a1)

warunalakshitha (approved):
LGTM

PR #680, opened 2026-07-27
Refactor repository into multiple Go modules

### LINKED ISSUE #675: [Task]: Convert the single module to multiple modules in a workspace
### Description

This should be straightforward. We'll map each top level package to it's own module

### Describe your task(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Summary
- split top-level packages into separate Go modules
- add a Go workspace and shared Makefile for repository-wide commands
- update imports, tests, and CI workflows for the multi-module layout

Closes #675


heshanpadmasiri:
@warunalakshitha, @nipunayf  most of the complexity here is to allow for independent releases for each part so that we can release parts separately (for example bal build depends on v1 of ast where as lsp depends on v2 of the ast).  In the future is this a realistic scenario? If not we can keep everything as a single module

interpsrc.go
```diff
@@ -31,17 +31,60 @@ import (
...
-//go:embed parser/*.go parser/nodes.json parser/common parser/tree
-var src embed.FS
+var rootSource embed.FS
+
+type sourceTree struct {
+	dir string
+	fs  fs.FS
+}
+
+var interpreterSources = []sourceTree{
```
warunalakshitha, line 68: Can we make this seperae go module(+ interpsrc_test.go) instead of keeping in root

warunalakshitha:
> @warunalakshitha, @nipunayf most of the complexity here is to allow for independent releases for each part so that we can release parts separately (for example bal build depends on v1 of ast where as lsp depends on v2 of the ast). In the future is this a realistic scenario? If not we can keep everything as a single module

I'd prefer to keep the release process simple. Releasing each part separately means maintaining separate versions per component, and I don't see a concrete requirement for independent releases right now. Even when the LS moves to its own repo, it can depend on a specific language version — so at most we'd need to publish the shared AST/contract as a versioned artifact then, not version every module independently today. I suggest we merge to a single module and split only when a real need appears.

ast/interpreter_source.go
warunalakshitha, line 1: can we rename this to ast_source.go (other module source files as well) and have one separate interptretr_source module to combine all and remove from root.

README.md
```diff
@@ -101,12 +101,17 @@ E.g.
 
 ### Testing
 
-To run the tests, use the following command:
+Use the repository Makefile to operate on every module in the Go workspace:
 
 ```bash
-go test ./...
+make build
```
warunalakshitha: These should be moved to DEVELOPING.md

ast/interpreter_source.go
warunalakshitha: saying interpreter_source.go inside ast module kind of wrong for me.

lib/go.sum
```diff
@@ -0,0 +1,16 @@
+github.com/cockroachdb/apd/v3 v3.2.3 h1:4Zx+I3R35bFXMnltzmjP79i2cravE4jTRL6ps9Aux80=
+github.com/cockroachdb/apd/v3 v3.2.3/go.mod h1:klXJcjp+FffLTHlhIG69tezTDvdP065naDsHzKhYSqc=
+github.com/lib/pq v1.10.7 h1:p7ZhMD+KsSRozJr34udlUrhboJwWAgCg34+/ZZNvZZw=
+github.com/lib/pq v1.10.7/go.mod h1:AlVN5x4E4T544tWzH6hKfbfQvm3HdbOxrmggDNAPY9o=
+github.com/santhosh-tekuri/jsonschema/v6 v6.0.2 h1:KRzFb2m7YtdldCEkzs6KqmJw4nqEVZGK7IN2kJkjTuQ=
+github.com/santhosh-tekuri/jsonschema/v6 v6.0.2/go.mod h1:JXeL+ps8p7/KNMjDQk3TCwPpBy0wYklyWTfbkIzdIFU=
+github.com/sergi/go-diff v1.4.0 h1:n/SP9D5ad1fORl+llWyN+D6qoUETXNZARKjyY2/KVCw=
+github.com/sergi/go-diff v1.4.0/go.mod h1:A0bzQcvG0E7Rwjx0REVgAGH58e96+X0MeOfepqsbeW4=
+golang.org/x/crypto v0.54.0 h1:YLIA59K4fiNzHzjnZt2tUJQjQtUWfWbeHBqKtk3eScw=
```
warunalakshitha, line 9: go mod has 	golang.org/x/crypto v0.52.0 but this golang.org/x/crypto v0.54.0 h1:YLIA59K4fiNzHzjnZt2tUJQjQtUWfWbeHBqKtk3eScw=. should we fix that. same thing on other module go.sum as well

heshanpadmasiri:
Closing in favor of #718, which is part of stack #724.

PR #683, opened 2026-07-28
feat: add support for start/wait actions

### LINKED ISSUE #283: Add support for future values
### Description

$subject

### Describe your task(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #296: Add frontend support for future values
For properly creating futures I think we need to support workers, but we can still allow for defining these values with `extern` functions

### LINKED ISSUE #297: Add backend support for future values
Fallow on from #296 . This is mostly focusing of the structure of the value it self rather than how to get it (which we can't without workers)

### LINKED ISSUE #427: Add support for start and wait actions
This requires #283 . Internally future must track the strand so we can wait on it

heshanpadmasiri:
## Purpose
$subject
Resolves https://github.com/ballerina-nutcracker/ballerina/issues/427

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous `start` actions and single, alternate, and multiple `wait` actions.
  - Added support for futures, including typed futures, result retrieval, error handling, panic propagation, and duplicate-wait detection.
  - Added asynchronous calls for functions, methods, closures, and resource actions.
  - Added cooperative scheduling and strand execution behavior.
- **Bug Fixes**
  - Improved diagnostics for invalid future usage, isolation violations, and starting actions while holding locks.
- **Documentation**
  - Documented concurrency, scheduling, and new call formats.
- **Tests**
  - Added comprehensive compiler, runtime, integration, and concurrency coverage.


ast/node_builder.go
```diff
@@ -4601,7 +4597,67 @@ func (n *NodeBuilder) TransformTrippleGTToken(trippleGTTokenNode *tree.TrippleGT
...
+		expr, ok := n.createActionOrExpression(futureExpr.(tree.ExpressionNode)).(BLangExpression)
+		if !ok {
+			n.cx.InternalError("single wait operand is not an expression", n.getPosition(futureExpr))
+			return nil
+		}
+		action := &BLangSingleWaitAction{FutureExpr: expr}
+		action.SetPosition(n.getPosition(waitActionNode))
+		return action
+	}
 }
```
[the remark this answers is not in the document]
heshanpadmasiri: Not a reachable compiler bug. In parser/tree/node_gen.go, tree.ExpressionNode is an alias for tree.NonTerminalNode, so after TransformWaitAction filters tree.Token children from an ExternalTreeNodeList, every remaining parser-produced child is already an ExpressionNode. For the default branch, parser/parser.go parseSingleOrAlternateWaitAction always supplies either an expression nonterminal, an ExternalTreeNodeList, or a WaitFieldsListNode; the latter two are handled by earlier switch cases. Focused CLI checks with malformed forms (`wait;`, `wait |;`, `wait 1 | ;`, `wait ||;`, and `wait { };`) all emitted syntax diagnostics without panicking, and `go test ./ast` passed. Adding guards in ast/node_builder.go TransformWaitAction would only defend against manually violating the syntax-tree construction invariant and would not improve parser-generated compilation.

corpus/integration/subset9/09-future/start7-e.txtar
```diff
@@ -0,0 +1,13 @@
...
+  --> start7-e.bal:23:16
+   |
+23 |         return trap wait result; // @error action is not an isolated expression
+   |                ^^^^^^^^^^^^^^^^
+
+error[SEMANTIC_ERROR]: access of mutable variable
+  --> start7-e.bal:23:16
+   |
+23 |         return trap wait result; // @error action is not an isolated expression
+   |                ^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

values/future.go
```diff
@@ -0,0 +1,106 @@
...
+	return &Future{Type: ty, ready: make(chan struct{})}
+}
+
+func (f *Future) Complete(result BalValue, panicValue any) {
+	f.outcome = futureOutcome{result: result, panic: panicValue}
+	f.mu.Lock()
+	f.state |= futureReady
+	f.ord = futureCompletionOrder.Add(1)
+	f.mu.Unlock()
+	close(f.ready)
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

values/future.go
```diff
@@ -0,0 +1,110 @@
...
+	ready   chan struct{}
+	mu      sync.Mutex
+	state   futureState
+	ord     uint64
+	outcome futureOutcome
+}
+
+func NewFuture(ty semtypes.SemType) *Future {
+	return &Future{Type: ty, ready: make(chan struct{})}
+}
```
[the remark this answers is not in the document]
heshanpadmasiri: Not a functional bug in this runtime architecture. The execution strand is represented by *extern.Context, not by values.Future. runtime/internal/exec/strand.go startFuture creates that context with contextForStartedStrand before launch; the started closure retains and executes on it, and non-isolated strands are queued on the shared runtime/internal/thread.Thread. runtime/internal/exec/terminators.go waitFuture, waitAnyFuture, and waitAllFutures yield the waiting context while the producer context runs, then consume Future's completion state. No wait path needs to retrieve the producer context from Future. Storing *extern.Context in values.Future would also introduce an import cycle because runtime/extern already imports values. Focused CLI runs of corpus/bal/subset9/09-future/start1-v.bal, start12-v.bal, and wait2-v.bal exercised single, cooperative non-isolated, alternate, panic, and repeated waits successfully; go test -race ./runtime/... ./values also passed. Keep values/future.go Future/NewFuture and runtime/internal/exec/strand.go startFuture unchanged.

runtime/internal/exec/terminators.go
```diff
@@ -17,13 +17,177 @@
...
+		}
+		for i, candidate := range futures {
+			if !completed[i] && candidate == future {
+				errors[i] = err
+				completed[i] = true
+				remaining--
+			}
+		}
+	}
+	return errors[len(errors)-1]
```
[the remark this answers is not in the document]
heshanpadmasiri: fixed

corpus/bal/subset9/09-future/start12-v.bal
```diff
@@ -0,0 +1,53 @@
...
+    io:println(completedResult); // @output 4
+                                 // @output error("multiple waits on the same future is not allowed")
+
+    future<int> failed = start crash();
+    _ = start printAndReturn(5);
+    int|error trapped = trap wait failed;
+    io:println(trapped is error); // @output 5
+                                  // @output true
+
+    _ = start printAndReturn(6);
```
warunalakshitha, line 51: Should we print 6 before exit( same strand group items should be finish before graceful stop)?

corpus/parser/subset10/10-lock/lock-trapped-action-transfer-1-e.json
warunalakshitha, line 1: we do not need this file

warunalakshitha:
I think it is better to update the architecture model how ballerina strand mapped, usage of goroutines, strand group , diff between isolated and non isolated start for future references. may be with diagram if needed

semantics/internal/analysis/lock_analyzer.go
```diff
@@ -175,6 +175,10 @@ func validateLockInvocations(a analyzer, body ast.BLangNode) bool {
 		switch n := inner.(type) {
 		case *ast.BLangLambdaFunction, *ast.BLangFunction:
 			return false
+		case *ast.BLangStartAction:
```
warunalakshitha, line 178: Should we handle wait cases as well

runtime/extern/extern.go
```diff
@@ -102,20 +104,43 @@ func (ctx *Context) TypeEnv() semtypes.Env {
...
+
+// Schedule queues this strand on its thread. Every call must be paired with
+// exactly one Complete when the scheduled strand finishes.
+func (ctx *Context) Schedule() <-chan struct{} {
+	return ctx.thread.Schedule()
+}
+
+// Yield queues this strand and selects the next strand on its thread.
+func (ctx *Context) Yield() <-chan struct{} {
+	return ctx.thread.Yield()
```
warunalakshitha, line 136: should we handle strand cancel cases like https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/scheduling/Strand.java#L95

warunalakshitha: Should we check inside a lock in yield cases too

semantics/internal/analysis/lock_analyzer.go
heshanpadmasiri: AFAIK there are no restrictions on waiting, spec only restricts starting strands (if lexically in we give a compile time error else panic)

warunalakshitha:
What happen if we have gracefulstop (main completed), are we complete same strand group executions and wait for exit?

corpus/bal/subset9/09-future/start12-v.bal
heshanpadmasiri: `printAndReturn` is not isolated so we create the strand but wait until a yield point to start execution. We don't have a yield point after this (If I recall jBallerina treats all extern functions as yield points so probably it start execution at `io:println`)

runtime/extern/extern.go
heshanpadmasiri: We don't support strand cancelation yet. It's a lang.future feature IMO; when we do I think some changes will be need to the thread yield implementation but users shouldn't be affected (I haven't thought this through fully but cancelling trigger an abnormal completion so I don't think it is something the extern function can and should handle)

heshanpadmasiri: AFAIK there are no restrictions on yielding inside lock as well

heshanpadmasiri:
> What happen if we have gracefulstop (main completed), are we complete same strand group executions and wait for exit?

I don't fully understand this. If the question is are we waiting for started strands to complete (in the same thread or different threads) answer is no. AFAIK spec don't require it. It requires listeners to wait for any strands it started before returning in graceful stop. But there is nothing in the language to enforce that at compile time. Eitherway that's an issue for libraries not runtime

runtime/internal/exec/terminators.go
```diff
@@ -17,13 +17,177 @@
...
+			completed[i] = true
+			remaining--
+		}
+	}
+	return results
+}
+
+func waitFuture(ctx *extern.Context, future *values.Future) values.BalValue {
+	for {
+		<-ctx.Yield()
```
warunalakshitha, line 142: If we are waiting on an isolated future (store that on the future at start), we don’t need to yield — call GetClaimed and park on the ready channel. The yield-poll loop is only needed for non-isolated (same-thread) waits so the queued producer can run.

warunalakshitha:
LGTM except some clarifications/NITs

runtime/internal/exec/terminators.go
heshanpadmasiri: According to spec (https://ballerina.io/spec/lang/master/#section_7.2) wait is a yield point, so it must always yield irrespective of the future it is waiting on.  I don't have strong opinion on wether we should claim when we get the value or before future is ready. Spec is not very clear on this. Either way assuming multiple waits that actually try to read the future that will only change which wait will get the error

heshanpadmasiri:
> I think it is better to update the architecture model how ballerina strand mapped, usage of goroutines, strand group , diff between isolated and non isolated start for future references. may be with diagram if needed

Fixed

corpus/parser/subset10/10-lock/lock-trapped-action-transfer-1-e.json
heshanpadmasiri: fixed

runtime/internal/exec/terminators.go
warunalakshitha: Agree — wait should always yield once as a yield point.

My only concern was the repeated yield-poll for isolated futures turning into a busy loop. Fine to keep as-is if that’s intentional; otherwise yield once then park on the ready channel. That gets awkward for alternate/multiple wait though, so I’m okay leaving the current loop.

Claim timing: no preference either.

warunalakshitha (dismissed):
LGTM

warunalakshitha:
@coderabbitai please review

warunalakshitha:
Codecov shows less than 80% even it was covered due to a bug in codecov script. I tested same pr with https://github.com/ballerina-nutcracker/ballerina/pull/843 and it was showing 87%

warunalakshitha:
@coderabbitai please review

warunalakshitha:
There was a problem with collecting coverage packages with codecov and it shows the correct coverage with https://github.com/ballerina-nutcracker/ballerina/pull/843 87.63%

warunalakshitha:
Lets fix the coderabbitai comments.

warunalakshitha (approved):

PR #684, opened 2026-07-28
feat(query): implement query actions

### LINKED ISSUE #616: [New Feature]: Implement query actions
### Description

Refer to : https://ballerina.io/spec/lang/master/#section_7.10

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

gimantha:
## Purpose
Fix https://github.com/ballerina-nutcracker/ballerina/issues/616

## Automation tests
- Unit tests
  > Code coverage information
- Integration tests

## Security checks

 

## Summary by CodeRabbit

* **New Features**
  * Added query actions with `do` blocks, joins, filtering, ordering, grouping, limiting, nested queries, and control flow.
  * Query operations now support strings, XML, lists, maps, streams, and iterable objects.
  * Added Unicode string iteration and support for combining actions with expressions.
* **Bug Fixes**
  * Improved query scoping, error propagation, iterator completion, trapping, and final-variable validation.
* **Tests**
  * Added comprehensive coverage for query actions, failures, joins, ordering, scope, and control flow.


warunalakshitha:
Lets refactor PR subject and commit messages with Conventional commits format.

desugar/desugar.go
```diff
@@ -208,6 +209,28 @@ func (ctx *functionContext) currentLoopVar() ast.LExpr {
...
+func (ctx *functionContext) pushQueryActionControl(state *queryActionControlFlowState) {
+	ctx.queryActionControls = append(ctx.queryActionControls, state)
+}
+
+func (ctx *functionContext) popQueryActionControl() {
+	if len(ctx.queryActionControls) == 0 {
+		ctx.internalError("cannot pop from empty query action control stack")
+	}
+	ctx.queryActionControls = ctx.queryActionControls[:len(ctx.queryActionControls)-1]
+}
```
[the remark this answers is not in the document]
gimantha: @coderabbitai check if all the comments are addressed in commit https://github.com/ballerina-nutcracker/ballerina/pull/684/commits/dcfc3588b8d6f894560ef5c928581353112a7072 and resolve if fixed

ast/expressions.go
```diff
@@ -247,7 +258,7 @@ type (
 
 	BLangTrapExpr struct {
 		bLangExpressionBase
-		Expr BLangExpression
+		Expr BLangActionOrExpression
```
heshanpadmasiri, line 266: `trap` now accepts a query action, but the following valid program fails during desugaring with `INTERNAL_ERROR: Init statements will be hoisted outside of trap region`:

```ballerina
import ballerina/io;

public function main() {
    error? result = trap (from var value in [1, 2]
        do {
            if value == 2 {
                panic error("boom");
            }
        });
    io:println(result is error); // @output true
}
```

jBallerina prints `true`. Please keep the generated query statements inside the trap region and add this as a corpus test.

ast/node_builder.go
```diff
@@ -3358,6 +3366,9 @@ func (n *NodeBuilder) TransformLetVariableDeclaration(letVariableDeclarationNode
 	if annotations.Size() > 0 {
 		panic("annotations not yet supported")
 	}
+	if variableDef, ok := varDef.(*BLangSimpleVariableDef); ok && variableDef.Var != nil {
+		variableDef.Var.SetFinal()
+	}
```
heshanpadmasiri: `createBLangVarDef` returns a simple variable definition with a non-nil variable for every supported binding pattern; unsupported patterns panic. This condition cannot be false.

```suggestion
	variableDef := varDef.(*BLangSimpleVariableDef)
	variableDef.Var.SetFinal()
```

ast/node_builder.go
```diff
@@ -4348,10 +4359,13 @@ func (n *NodeBuilder) TransformQueryConstructType(queryConstructTypeNode *tree.Q
...
 	fromClause := &BLangFromClause{}
 	fromClause.pos = n.getPosition(fromClauseNode)
-	fromClause.SetCollection(n.createExpression(fromClauseNode.Expression()))
+	fromClause.SetCollection(n.createActionOrExpression(fromClauseNode.Expression()))
 	bindingPatternNode := fromClauseNode.TypedBindingPattern()
 	fromClause.SetVariableDefinitionNode(n.createBLangVarDef(n.getPosition(bindingPatternNode), bindingPatternNode,
 		nil, nil))
+	if fromClause.VariableDefinitionNode != nil && fromClause.VariableDefinitionNode.Var != nil {
+		fromClause.VariableDefinitionNode.Var.SetFinal()
+	}
```
heshanpadmasiri: The variable-definition invariant is already established by `createBLangVarDef`, so this guard is redundant.

```suggestion
	fromClause.VariableDefinitionNode.Var.SetFinal()
```

ast/node_builder.go
```diff
@@ -4378,11 +4392,14 @@ func (n *NodeBuilder) TransformLetClause(letClauseNode *tree.LetClauseNode) BLan
...
 	joinClause.pos = n.getPosition(joinClauseNode)
-	joinClause.SetCollection(n.createExpression(joinClauseNode.Expression()))
+	joinClause.SetCollection(n.createActionOrExpression(joinClauseNode.Expression()))
 	bindingPatternNode := joinClauseNode.TypedBindingPattern()
 	joinClause.SetVariableDefinitionNode(
 		n.createBLangVarDef(n.getPosition(bindingPatternNode), bindingPatternNode, nil, nil),
 	)
+	if joinClause.VariableDefinitionNode != nil && joinClause.VariableDefinitionNode.Var != nil {
+		joinClause.VariableDefinitionNode.Var.SetFinal()
+	}
```
heshanpadmasiri: The same invariant applies to join bindings.

```suggestion
	joinClause.VariableDefinitionNode.Var.SetFinal()
```

ast/node_builder.go
```diff
@@ -4417,10 +4434,30 @@ func (n *NodeBuilder) TransformQueryPipeline(queryPipelineNode *tree.QueryPipeli
...
+
+	intermediateClauses := queryPipeline.IntermediateClauses()
+	for i := 0; i < intermediateClauses.Size(); i++ {
+		clause := intermediateClauses.Get(i)
+		switch clause.Kind() {
+		case common.FROM_CLAUSE, common.JOIN_CLAUSE, common.LET_CLAUSE, common.WHERE_CLAUSE,
+			common.GROUP_BY_CLAUSE, common.LIMIT_CLAUSE, common.ORDER_BY_CLAUSE:
+			queryClauseAdder.AddQueryClause(n.TransformSyntaxNode(clause))
+		default:
+			n.cx.Unimplemented("only from + join + let + where + group by + order by + limit query clauses are supported for now", n.getPosition(clause))
```
heshanpadmasiri: NIT: These clause names should be separated by commas rather than `+`.

```suggestion
			n.cx.Unimplemented("only from, join, let, where, group by, order by, and limit query clauses are supported for now", n.getPosition(clause))
```

desugar/query_expression.go
```diff
@@ -46,7 +46,11 @@ func walkQueryExpr(cx *functionContext, expr *ast.BLangQueryExpr) desugaredNode[
 	} else {
 		collectClause = expr.QueryClauseList[finalClauseIndex].(*ast.BLangCollectClause)
 	}
-	if collectClause != nil || queryExprNeedsRowPipeline(expr, 1, finalClauseIndex) {
+	fromCollectionTy := fromClause.Collection.GetDeterminedType()
+	if collectClause != nil ||
+		semtypes.IsSubtype(cx.typeCtx(), fromCollectionTy, semtypes.STRING) ||
+		semtypes.IsSubtype(cx.typeCtx(), fromCollectionTy, semtypes.XML) ||
```
heshanpadmasiri: This can directly test the union.

```suggestion
		semtypes.IsSubtype(cx.typeCtx(), fromCollectionTy, semtypes.Union(semtypes.STRING, semtypes.XML)) ||
```

desugar/query_expression.go
```diff
@@ -228,6 +232,88 @@ func createQueryCollectionSource(
...
+	}
+
+	collectionResult := walkExpression(cx, collectionExpr)
+	*initStmts = append(*initStmts, collectionResult.initStmts...)
+	collectionValue := collectionResult.replacementNode
+	collectionTy = collectionValue.GetDeterminedType()
+	collectionVarDef, collectionRef := assignActionOrExpressionToLocal(cx, collectionValue, pos)
+	*initStmts = append(*initStmts, collectionVarDef)
+
+	if semtypes.IsSubtypeSimple(collectionTy, semtypes.STREAM) {
```
heshanpadmasiri: The list/map case is handled above; the remaining mutually exclusive cases can be made explicit with a switch like this:

```go
switch {
case semtypes.IsSubtype(tyCtx, collectionTy, semtypes.STREAM):
    return queryActionCollectionSource{
        nextReceiverRef: collectionRef,
        nextReceiverTy:  collectionTy,
    }, true
case semtypes.IsSubtype(tyCtx, collectionTy, semtypes.STRING),
    semtypes.IsSubtype(tyCtx, collectionTy, semtypes.XML),
    semtypes.IsSubtype(tyCtx, collectionTy, semtypes.OBJECT):
    iteratorInvocation := createIteratorInvocation(cx, collectionRef, collectionTy, pos)
    if iteratorInvocation == nil {
        return queryActionCollectionSource{}, false
    }
    iteratorVarDef, iteratorRef := assignToLocal(cx, iteratorInvocation, pos)
    *initStmts = append(*initStmts, iteratorVarDef)
    return queryActionCollectionSource{
        nextReceiverRef: iteratorRef,
        nextReceiverTy:  iteratorInvocation.GetDeterminedType(),
    }, true
default:
    cx.internalError("query action collection type should have been validated during type resolution")
    return queryActionCollectionSource{}, false
}
```

This also removes the duplicated iterator setup and the detached object guard.

lib/langlibs/ballerina/lang.string/0.0.1/any/lang.string.bal
```diff
@@ -20,6 +20,14 @@
 # + return - the number of characters (code points) in `str`
 public isolated function length(string str) returns int = external;
 
+# Returns an iterator over the Unicode characters in the string.
+#
+# + str - the string to iterate over
+# + return - a new character iterator
+public isolated function iterator(string str) returns object {
```
heshanpadmasiri, line 27: I think it is cleaner to simply implement this as a ballerina class with internal private handle. We can implement extern functions that accept the handle and return values. I would also keep next simple by returning just string and a separate hasNext function. 
Handle can point to a go struct which manages internal state. See https://github.com/ballerina-nutcracker/ballerina/issues/638

semantics/control_flow_analyzer.go
```diff
@@ -343,6 +344,47 @@ func (analyzer *functionControlFlowAnalyzer) analyzeStatement(curBB bbRef, stmt
 	}
 }
 
+func (analyzer *functionControlFlowAnalyzer) analyzeQueryActionsInStatement(curBB bbRef, stmt ast.StatementNode) {
+	collector := &queryActionCollector{root: stmt}
+	ast.Walk(collector, stmt.(ast.BLangNode))
+	for _, action := range collector.actions {
+		if action.DoClause == nil || action.DoClause.Body == nil {
```
heshanpadmasiri: I think it is simpler to just filter these in the visitor it self rather than doing it here

semantics/semantic_analyzer.go
```diff
@@ -1251,6 +1273,30 @@ func analyzeQueryExpr[A analyzer](a A, queryExpr *ast.BLangQueryExpr, expectedTy
...
+	if !analyzeQueryFromClause(a, fromClause) {
+		return false
+	}
+	if !analyzeQueryIntermediateClauses(a, action.QueryClauseList, len(action.QueryClauseList)) {
+		return false
+	}
+	if action.DoClause == nil || action.DoClause.Body == nil {
+		a.internalErr("query action shape should have been validated during type resolution", action.GetPosition())
+		return false
+	}
```
heshanpadmasiri: Type resolution has already validated all three shape invariants before semantic analysis. These defensive checks hide the stage contract and are unreachable from source; rely on the resolved AST instead.

```suggestion
	fromClause := action.QueryClauseList[0].(*ast.BLangFromClause)
	if !analyzeQueryFromClause(a, fromClause) {
		return false
	}
	if !analyzeQueryIntermediateClauses(a, action.QueryClauseList, len(action.QueryClauseList)) {
		return false
	}
```

semantics/semantic_analyzer.go
```diff
@@ -2004,6 +2054,10 @@ func analyzeAssignment[A analyzer](a A, assignment assignmentNode) bool {
 			a.semanticErr("cannot assign to annotation", variable.GetPosition())
 			return false
 		}
+		if valueSym, ok := ctx.GetSymbol(symbol).(model.ValueSymbol); ok && valueSym.IsFinal() {
+			a.semanticErr("cannot assign to final variable", variable.GetPosition())
+			return false
+		}
```
heshanpadmasiri: I don't think this can happen

```suggestion
```

semantics/type_resolver.go
```diff
@@ -4144,25 +4113,171 @@ func resolveQueryExpr(
...
+		valueRecordTy := semtypes.Diff(nextReturnTy, semtypes.Union(semtypes.NIL, semtypes.ERROR))
+		if !semtypes.IsSubtype(ctx, valueRecordTy, semtypes.MAPPING) {
+			t.semanticError("query iterator next method must return a record with a value field", pos)
+			return semtypes.SemType{}, semtypes.SemType{}, false
+		}
+		elementTy := semtypes.MappingMemberTypeInnerVal(ctx, valueRecordTy, semtypes.StringConst("value"))
+		if semtypes.IsZero(elementTy) {
+			t.semanticError("query iterator next method must return a record with a value field", pos)
+			return semtypes.SemType{}, semtypes.SemType{}, false
+		}
```
heshanpadmasiri: Once `collectionTy` is proven to be a subtype of `object:Iterable`, its contract guarantees callable `iterator()`/`next()` members and a next-result record containing `value`. None of these error branches can be reached by valid source; malformed shapes fail the inclusion check above. 

```suggestion
		iteratorFnTy := semtypes.ObjectMemberType(ctx, semtypes.StringConst("iterator"), collectionTy)
		iteratorTy := semtypes.FunctionReturnType(ctx, iteratorFnTy, emptyListTy)
		nextFnTy := semtypes.ObjectMemberType(ctx, semtypes.StringConst("next"), iteratorTy)
		nextReturnTy := semtypes.FunctionReturnType(ctx, nextFnTy, emptyListTy)
		valueRecordTy := semtypes.Diff(nextReturnTy, semtypes.Union(semtypes.NIL, semtypes.ERROR))
		elementTy := semtypes.MappingMemberTypeInnerVal(ctx, valueRecordTy, semtypes.StringConst("value"))
```

heshanpadmasiri (changes requested):
Generally LGTM

desugar/query_expression.go
gimantha: fixed

lib/langlibs/ballerina/lang.string/0.0.1/any/lang.string.bal
gimantha: Fixed

semantics/control_flow_analyzer.go
gimantha: Fixed

ast/expressions.go
gimantha: Fixed and added the same testcase

bir/bir_gen.go
```diff
@@ -991,12 +991,22 @@ func handleActionOrExpression(ctx context, curBB *BIRBasicBlock, expr ast.BLangA
 		return newExpression(ctx, curBB, expr)
 	case *desugar.BLangServiceInit:
 		return serviceInitExpression(ctx, curBB, expr)
+	case *ast.BLangStatementExpression:
+		stmtEffect := handleStatement(ctx, curBB, expr.Stmt)
+		return handleActionOrExpression(ctx, stmtEffect.block, expr.Expr)
```
heshanpadmasiri, line 996: The statement block is popped before its result expression is evaluated, so `trap` silently loses a query action's stream-completion error; this prints `false` on this branch instead of jBallerina's `true`.

```ballerina
import ballerina/io;

isolated class FailingIterator {
    private int index = 0;

    public isolated function next() returns record {| int value; |}|error? {
        lock {
            if self.index == 0 {
                self.index = 1;
                return {value: 10};
            }
            return error("completion failed");
        }
    }
}

public function main() {
    stream<int, error?> values = new (new FailingIterator());
    error? result = trap (from var value in values
        do {
            _ = value;
        });
    io:println(result is error);
}
```

ast/clauses.go
```diff
@@ -79,7 +79,7 @@ type (
 	}
 	BLangSelectClause struct {
 		bLangNodeBase
-		Expression BLangExpression
+		Expression BLangActionOrExpression
```
heshanpadmasiri, line 82: Not entiely sure if this is intentional. Accroding to spec https://ballerina.io/spec/lang/master/#select-clause, select don't support actions. 

We did relaxed this for check/checkpanic given jBallerina already allow actions not just expression. If this is due something similar it is fine, but we should probably need a comment in that case

ast/clauses.go
```diff
@@ -141,11 +141,11 @@ var (
 	_ BLangNode = &BLangOnFailClause{}
 )
 
-func (b *BLangJoinClause) GetCollection() BLangExpression {
+func (b *BLangJoinClause) GetCollection() BLangActionOrExpression {
```
heshanpadmasiri, line 144: ditto

ast/expressions.go
```diff
@@ -252,7 +263,14 @@ type (
 
 	BLangTrapExpr struct {
 		bLangExpressionBase
```
heshanpadmasiri, line 265: I assume we are relaxing this inline with check (spec restricts this to be only expression; but we have already relaxed check and checkpanic to match jBallerina)

ast/expressions.go
```diff
@@ -252,7 +263,14 @@ type (
...
-		Expr BLangExpression
+		Expr BLangActionOrExpression
+	}
+	// BLangStatementExpression is introduced during desugaring when statements
+	// must remain within an enclosing expression's evaluation region.
+	BLangStatementExpression struct {
```
heshanpadmasiri, line 270: I am not entierly sure this is the correct approach. I think this has to do with  check expression (ran in to same for ternary and fixed in https://github.com/ballerina-nutcracker/ballerina/pull/746)? I think if we can't properly desugar something to valid ballerina AST, desugar shouldn't desguar at all and pass in the responsibility to bir gen, unless there is no way to represent it in ballerina source code (this was the rational behind service start)

corpus/ast/subset9/09-query-actions/query-action-abrupt-completion-v.txt
```diff
@@ -0,0 +1,275 @@
+(compilation-unit  regular-source
```
heshanpadmasiri, line 1: Let's move all the new test cases to subset 10

desugar/desugar.go
```diff
@@ -199,6 +199,7 @@ type functionContext struct {
 	scopeStack           []model.Scope
 	desugarSymbolCounter int
 	loopVarStack         []ast.LExpr // Stack to track loop variables (nil for while, varRef for desugared foreach)
+	queryActionControls  []*queryActionControlFlowState
```
heshanpadmasiri, line 202: I don't see why this needs to be an array of pointers

semantics/symbol_resolver.go
```diff
@@ -127,10 +127,11 @@ type (
...
 	blockSymbolResolver struct {
-		parent     symbolResolver
-		scope      model.BlockLevelScope
-		node       ast.BLangNode
-		varTracker *varTracker
+		parent        symbolResolver
+		scope         model.BlockLevelScope
+		node          ast.BLangNode
+		varTracker    *varTracker
+		hiddenSymbols map[model.SymbolRef]bool
```
heshanpadmasiri, line 134: It is not clear to me why this is needed. Ideally if you create the scopes correctly within the query visibility should be handled naturally

bir/bir_gen.go
gimantha: Fixed in c28c1be88. Removed the statement-expression block wrapper and reused the upstream expression-thunk lowering, which evaluates the setup statements and result inline without popping the result variable’s scope first. Added `subset10/10-query-actions/query-action-trap-completion-v.bal` with your reproducer; it prints `true` in both this interpreter and jBallerina. The query integration and BIR corpus checks pass.

ast/clauses.go
gimantha: Confirmed this compatibility behavior with jBallerina: direct remote actions are accepted in `select`, including the existing `query-actions-or-expressions-v.bal` cases. Added a comment on `BLangSelectClause.Expression` explaining why it accepts actions. The corpus covers direct, parenthesized, checked, resource-call, and nested query-action results.

ast/clauses.go
gimantha: Restored the expression-only contract for join construction and collection access. From clauses still accept actions. A private generic input-clause base shares the binding fields and methods while keeping the two operand types distinct. Added `subset10/10-query-expression/query-join-actions-e.bal`; both implementations reject the two forms at the marked lines.

ast/expressions.go
gimantha: Yes, this follows jBallerina’s action-operand compatibility. The merged upstream `BLangTrapExpr` already documents it, so that explanation is retained. Query-action panic trapping and stream-completion errors now have corpus coverage, including your completion-error reproducer.

ast/expressions.go
gimantha: Removed `BLangStatementExpression` and its AST walker/printer/BIR special case. Query lowering now uses the existing upstream desugar-owned `BLangExpressionThunk`; the BIR generator lowers its setup and result inline at the original evaluation point. This preserves the surrounding trap and control-flow context without adding a new source AST expression or a synthetic function. The query trap, abrupt-completion, and loop-control corpus cases pass.

corpus/ast/subset9/09-query-actions/query-action-abrupt-completion-v.txt
gimantha: Moved all 26 corpus sources introduced by this PR to `subset10/10-query-actions` and `subset10/10-query-expression`, together with their AST, CFG, desugared, BIR, and integration goldens. Updated embedded source paths and regenerated the goldens for the current pipeline. The two new review regression tests are also in subset10.

desugar/desugar.go
gimantha: Changed `queryActionControls` to `[]queryActionControlFlowState`. Push now stores a value, and the current-state accessor takes the address of the last slice element when needed. The query loop-control and nested-query corpus cases pass.

semantics/symbol_resolver.go
gimantha: Removed `hiddenSymbols`, `withHiddenSymbols`, and the special filtering in symbol lookup/shadowing checks. Join collections and limits now resolve in the enclosing scope; the left join key resolves in the current pipeline scope, and the right key resolves in a separate scope containing the right binding. The continuation scope shares that binding’s symbol space and chains to the preceding pipeline scope, preserving symbol identity without hiding symbols. Existing positive and negative query-scope corpus cases pass. The smaller positive scope example also matches jBallerina; jBallerina crashes on the existing combined action-scope example.

lib/langlibs/go/lang.string/string.go
```diff
@@ -150,3 +158,19 @@ func mapASCII(s string, f func(rune) rune) string {
...
+	iterator := args[0].(*stringIteratorHandle)
+	return iterator.offset < len(iterator.value), nil
+}
+
+func stringIteratorNext(_ *extern.Context, args []values.BalValue) (values.BalValue, error) {
+	iterator := args[0].(*stringIteratorHandle)
+	char, size := utf8.DecodeRuneInString(iterator.value[iterator.offset:])
+	iterator.offset += size
+	return string(char), nil
+}
```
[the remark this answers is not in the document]
gimantha: Fixed in c2e0698eb. `iteratorNext` now performs the exhaustion check and UTF-8 advance under one mutex, and the separate `iteratorHasNext` native call has been removed. Added a Ballerina corpus regression that calls `next()` concurrently and verifies the single character is returned exactly once.

semantics/internal/symbols/symbol_resolver.go
```diff
@@ -1453,6 +1460,38 @@ func visitInnerSymbolResolver[T symbolResolver](resolver T, node ast.BLangNode)
...
+		case *ast.BLangJoinClause:
+			ast.Walk(parent, clause.Collection.(ast.BLangNode))
+			ast.Walk(resolver, clause.OnClause.OnExpr.(ast.BLangNode))
+			right := newBlockSymbolResolverWithBlockScope(parent, clause)
+			name := clause.VariableDefinitionNode.Var.Name.GetValue()
+			if isShadowed(resolver, name) {
+				semanticError(resolver, "Variable already defined: "+name, clause.VariableDefinitionNode.GetPosition())
+			}
+			ast.Walk(right, clause.VariableDefinitionNode)
+			ast.Walk(right, clause.OnClause.EqualsExpr.(ast.BLangNode))
```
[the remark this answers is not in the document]
gimantha: Fixed in c2e0698eb. `resolveQuerySymbols` now guards the join collection, both on-clause expressions, the variable definition, `Var`, and `Name` before accessing or walking them, while preserving the left/right join scopes for complete clauses. The query scope corpus passes, and a malformed join now reports syntax diagnostics without a panic.

PR #687, opened 2026-07-28
fix(http): make Listener.gracefulStop block until in-flight requests drain

TharmiganK:
## What

Makes `http:Listener.gracefulStop` conform to its contract: it now blocks until all in-flight requests have drained and the server is closed, returning `()` on a clean drain or an `error` if the shutdown fails.

## Why

The previous implementation special-cased a resource calling `ep.gracefulStop()` on its own listener — it detected the serving strand and returned `nil` immediately while draining in the background. That reported success *before* the drain had actually happened, violating the documented contract ("waiting for in-flight requests to complete").

## How

- Removed the special-casing and its supporting state (`servingStrands`, `shutdownOnce`, `shutdownDone`, `shutdownErr`).
- `gracefulStop` now blocks unconditionally on `server.Shutdown`. Calling it from within a resource on its own listener self-deadlocks by design — that is caller error, not a case to work around.
- As a side effect this removes two exclusive, listener-wide lock acquisitions per request from the dispatch hot path.
- Reworked the graceful-stop corpus test to drive the stop from a non-serving strand (`testMain`) and regenerated its golden.

## Summary by CodeRabbit

* **New Features**
  * HTTP listeners now support graceful shutdown that waits for active requests to complete before closing.
  * Requests made after shutdown fail as expected.
* **Bug Fixes**
  * Improved HTTP request handling and streaming response efficiency.
* **Documentation**
  * Clarified graceful shutdown timing, timeout behavior, and errors when invoked from a resource on its own listener.
* **Tests**
  * Expanded coverage for successful responses, graceful shutdown confirmation, and post-shutdown connection failures.


TharmiganK:
@coderabbitai Please review

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go
```diff
@@ -213,58 +207,31 @@ func registerListenerExterns(rt *runtime.Runtime) {
...
-						return values.NewErrorWithMessage("Listener.gracefulStop: " + state.shutdownErr.Error()), nil
-					}
-				default:
-				}
-				return nil, nil
-			}
-			<-state.shutdownDone
-			if state.shutdownErr != nil {
-				return values.NewErrorWithMessage("Listener.gracefulStop: " + state.shutdownErr.Error()), nil
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), timeout)
```
warunalakshitha, line 231: Should we through an error if inside resource function?

TharmiganK: The Ballerina language spec and HTTP library spec does not specifically mention regarding this use case. In jBallerina case, calling graceful stop inside the resource function worked because the graceful stop of HTTP is same as immediate stop. So we may not return an error in that case. For returning error we need to know whether graceful is called from resource function, so we need to track the requests similar to the previous fix. I think for now we can document this and leave it as it is. If there is actually a requirement like this, we could check with the spec.

Documented via [652ae14](https://github.com/ballerina-nutcracker/ballerina/pull/687/commits/652ae14fad86654ef628d3eaeb4ce57f508e0608)

warunalakshitha (dismissed):
LGTM

lib/stdlibs/ballerina/http/0.0.1/go1.26/README.md
```diff
@@ -184,3 +184,4 @@ Support Levels:
...
 - **`responseLimits.maxStatusLineLength` is not enforced.** The value is accepted and validated (must be ≥ 0) but has no runtime effect. Go's HTTP transport does not expose a configurable maximum status line length (unlike jBallerina's Netty `HttpClientCodec`).
 - **Proxy DNS resolution is lazy, not eager.** In jBallerina, `ProxyConfig.host` is DNS-resolved at client creation time, and an unknown hostname causes an `error` from `new http:Client(...)`. In the Go runtime, DNS resolution is deferred to the first request that uses the proxy. A bad proxy hostname does not fail at init time.
+- **`gracefulStop` waits for in-flight requests to drain.** In jBallerina, `gracefulStop` effectively behaves like an immediate stop — it returns promptly without waiting for active requests, so calling it from within a resource on its own listener succeeds. The Go-native version implements the `http:Listener` contract literally and blocks until in-flight requests complete or the graceful-stop timeout (default 60s) elapses. A resource that calls `gracefulStop` on the listener serving it therefore self-deadlocks until the timeout elapses and then returns an error, rather than succeeding.
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [9473ce4](https://github.com/ballerina-nutcracker/ballerina/pull/687/commits/9473ce4b0dfa422e5f85385d2ea5b9c826c607c7)

warunalakshitha (approved):

PR #682, opened 2026-07-28
ci: add HTTP throughput regression check for PRs

TharmiganK:
## What

Adds an HTTP throughput regression check that runs on PRs, reports the base-vs-head delta as a comment, and fails the check on a significant regression. It's the HTTP counterpart to the existing CPU-time and memory PR benchmarks.

## Why

Changes in the language runtime or the `ballerina/http` stdlib can move HTTP performance significantly — recently a single commit cost ~11% throughput. This catches that class of regression automatically.

## How

`compiler-tools/benchmark-http/` is a self-contained Go tool (own module, standard library only, modeled on `compiler-tools/benchmark/`) — it lives alongside the other dev tools rather than under `.github/`, and its coverage is tracked like theirs. Per ref it checks out a worktree (role-qualified path, so base/head never collide even when they sanitize identically) and builds `bal` (`go build -o bal ./cli/cmd`), runs an embedded hello service (`GET /hello`) via `bal run` while timing startup, and drives load with `wrk` (bounded by a context deadline so a hung `wrk` can't stall the job). It interleaves repeated runs of base and head on the same runner so environmental noise is shared, then compares medians and emits a markdown report plus a `result.json` verdict.

Two workflows mirror the existing time/memory prepare/comment pair:

- `benchmark-pr-http-prepare.yml` (`pull_request`): installs wrk, runs the tool comparing `origin/<base>` vs `HEAD`, uploads the artifact, then a gate step fails the check on regression.
- `benchmark-pr-http-comment.yml` (`workflow_run`): runs on both success and failure so a regression yields both a red check and a comment; upserts a marker-tagged PR comment with the diff table. PR number and head SHA are read from the trusted `workflow_run` event context rather than the uploaded artifact — the artifact is written by the prepare job while running PR-head code, so trusting it for identity would let a malicious PR redirect this write-permission job's comment to an unrelated PR.

## Config

- 2 interleaved repeats × (30s warmup + 330s measured) × 2 refs ≈ 24 min of load.
- 10% throughput-drop threshold. The gate is on throughput only; latency / startup / peak-RSS are shown for context.
- `codecov.yml`: no longer excludes the tool from coverage — like `compiler-tools/benchmark`, it's usable as a standalone tool, not just CI glue, so its coverage counts.

## Validation

[...]

PR #689, opened 2026-07-28
fix(benchmark): look up the PR via the API instead of workflow_run.pull_requests

TharmiganK:
## Summary
- `workflow_run.pull_requests` is documented but unreliable in practice — [run 30346997330](https://github.com/ballerina-nutcracker/ballerina/actions/runs/30346997330) confirms it comes back empty even for a normal same-repo PR (`perf-runtime-optimizations`), not just forks. Trusting it made `benchmark-pr-http-comment.yml`'s comment step silently skip on every run since #682 merged.
- Replace it with `pulls.list({ head: "<owner>:<branch>" })`, keyed off the trusted `head_repository`/`head_branch` fields from the `workflow_run` event context — never from the uploaded artifact, preserving the trust boundary from #682.
- This also fixes a case that was never actually exercised: fork-originated PRs, where `pull_requests` is documented to always be empty.

## Test plan
- [x] `node --check` on the updated script
- [x] Mock harness exercising: normal single match, zero matches, stale `head_sha`, fork-PR (cross-owner `head` filter), missing `head_repository` — all produce the expected create/update/skip behavior
- [x] Confirm the next real `workflow_run` for an HTTP benchmark PR successfully posts a comment

### DIFF
--- a/.github/scripts/post-http-benchmark-comment.js
+++ b/.github/scripts/post-http-benchmark-comment.js
@@ -7,21 +7,32 @@ module.exports = async ({ github, context, core }) => {
   // never from benchmark-meta.json — that file is written by the prepare
   // job while running PR-head code, so a malicious PR could forge it to
   // make this privileged (pull-requests: write) job comment on an
-  // unrelated PR. pull_requests is only populated for same-repo PRs and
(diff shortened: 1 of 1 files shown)


warunalakshitha (approved):
LGTM

PR #698, opened 2026-08-03
fix(desugar): fix range variable not being in the scope of the loop body

### LINKED ISSUE #697: Capturing a foreach index variable reuses the binding across iterations
## Description

A closure created inside a `foreach` loop does not capture the iteration-specific value of the foreach index variable. All retained closures observe the final value of the shared index binding.

This also makes `corpus/bench/09-bench/single-wait-fanout-isolated-v.bal` nondeterministic: isolated started strands race with updates to the captured foreach index.

## Reproduction

```ballerina
import ballerina/io;

public function main() {
    (function () returns int)[] fns = [];
    foreach int i in 0 ... 100 {
        fns.push(function () returns int { return i; });
    }

    int sum = 0;
    foreach function () returns int fn in fns {
        sum += fn();
    }
    io:println(sum);
}
```

jBallerina prints `5050`; the interpreter prints `10201`, indicating that every closure observes the final `i == 101`.

A local variable declared in the loop body works correctly:

```ballerina
foreach int i in 0 ... 100 {
    int local = i;
    fns.push(function () returns int { return local; });
}
```

Both jBallerina and the interpreter print a sum of `5050` for this version.

## Likely cause

The foreach index variable is most likely allocated outside the frame created for the loop body and then mutated for loop progression. Closures therefore capture one shared binding. Loop-local variables work because the body gets a fresh frame on each iteration.

The lowering should use a hidden mutable loop-control variable and initialize a fresh source-level foreach variable in the per-iteration body frame.

heshanpadmasiri:
$subject
Resolves https://github.com/ballerina-nutcracker/ballerina/issues/697

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed variable capture in `foreach` loops across ranges, collections, iterators, and XML values.
  - Fixed variable capture in `while` loops when creating and invoking closures.
  - Preserved correct loop behavior for bounds, increments, and `continue` statements.

- **Tests**
  - Added coverage for captured values and expected output across supported loop types.
  - Added validation for XML elements, text, comments, and closure calculations.


warunalakshitha (approved):
LGTM

PR #705, opened 2026-08-04
feat(http): add data-binding support in HTTP client calls

### LINKED ISSUE #652: Add data-binding support in HTTP client calls
### Description

> $Subject

### Describe your problem(s)

Current HTTP client calls can only get the response as a `http:Response` object. There is no basic data-binding support to get the payload directly as an appropriate typed value

### Describe your solution(s)

Add data-binding support in HTTP client calls

### Related area

Other Area

https://github.com/ballerina-nutcracker/ballerina/issues/502

_No response_

_No response_

TharmiganK:
Fixes: https://github.com/ballerina-nutcracker/ballerina/issues/652

## Summary

Adds client-side response data binding to the `ballerina/http` standard library. Client remote methods can now deserialize the response payload directly to the contextually expected type instead of only returning an `http:Response`, promoting the *Client-side response data binding* row in the module README from **Not Yet Supported** to **Partially Supported**.

- `TargetType` — new public type, `typedesc<Response|anydata>`
- Every remote method except `head` gains a trailing `TargetType targetType = <>` parameter and returns `targetType|error`; `head` keeps `Response|error`, matching jBallerina
- Builder selected from the response `Content-Type`, mirroring jBallerina's media-type patterns: `application/json` (plus `text/json` and `+json`/`.json`/`-json` suffixes), `text/plain`, `application/x-www-form-urlencoded`, `application/octet-stream`, with a target-type-driven fallback when the header is absent or unrecognised
- Targets strictly narrower than the type their builder produces — an enum or singleton against `string`, a closed all-string record against `map<string>`, a tuple or fixed-length array against `byte[]` — are converted to that target, so a body outside the narrow type is an `error` rather than a value outside its declared type
- 4xx and 5xx responses map to an `error` whose message is the reason phrase and whose detail carries `statusCode`, `headers`, and `body`; an `http:Response` target (or any union containing one) bypasses this and yields the raw response
- Empty-body nilification: an absent payload binds to `()` for a nilable target
- `Response.getTextPayload()` fixed from `returns string` to `returns string|error`, matching the jBallerina signature it was already violating — the implementation could already fail (e.g. exceeding `responseLimits.maxEntityBodySize`) through a signature that promised `string`

## Example

```ballerina
import ballerina/http;
import ballerina/io;

type Person record {|
    string name;
    int age;
|};

    http:Client c = check new ("https://api.example.com");

    // JSON body deserialised straight to the record
    Person person = check c->get("/person");
    io:println(person.name);

    // text/plain body bound to a string
    string greeting = check c->get("/greeting");

    // An http:Response target still yields the raw response, status code and all
    http:Response raw = check c->get("/person");
    io:println(raw.statusCode);

    // A 4xx or 5xx becomes an error carrying the status in its detail
    Person|error missing = c->get("/missing");
    if missing is error {
        io:println(missing.message()); // Not Found
    }
}
```

`var` provides no contextually expected type, so the target is passed explicitly in that position:

```ballerina
var person = check c->get("/person", targetType = Person);
```

## Notes

- **Implemented natively, not in Ballerina.** jBallerina implements the builder ladder in `.bal`, reached from an external Java method that trampolines back into a private `processGet`/`processPost`. That shape cannot be transcribed here: `values.SemTypeForValue` maps every `*values.TypeDesc` to the top `TYPEDESC` semtype, so a `targetType is typedesc<X>` test always evaluates false at runtime and the whole ladder falls through to its `else` branch. Dispatch is therefore done in Go over `semtypes.IsSubtype`/`Intersect`/`Diff`. Porting the `.bal` shape later needs a native→Ballerina trampoline plus a `typeSubtypeOf(typedesc, typedesc)` predicate extern; the payoff would be real `distinct` error types, since `.bal`-constructed errors carry the correct semtype.
- **Errors are plain `error` values.** `http:ClientError`, `http:ClientRequestError`, and `http:RemoteServerError` are not declared, so the 4xx and 5xx cases cannot be distinguished with an `is` check and `error:detail()` is not available to read the detail. This follows the module's existing package-wide decision rather than introducing a new divergence. The binding code already carries `statusCode`/`headers`/`body` in the detail and tags the error type name, so only the construction site changes once distinct errors become constructible.
- **`TargetType` narrows jBallerina's union**, dropping the `stream<SseEvent, error?>` member. Server-Sent Events are not implemented at all (no `SseEvent` type, no `getSseEventStream`), so no compilable customer code can name the dropped member.
- **Two deliberate divergences from jBallerina**, both documented under *Notable Behavioural Changes*: a `()` target returns `()` and discards the body, where jBallerina routes it through the string builder and hands back a non-empty body as a `string`; and the status-error message uses the status code's registered phrase, because the PAL transport contract surfaces only the status code, not the phrase sent on the wire. A code outside the IANA registry (499, which nginx sends for a client-closed request) has no registered phrase, so the message becomes `status code <code>` rather than being left empty.
- **A status error with an absent body keeps its reason phrase.** jBallerina extracts the error body with the builder its `Content-Type` selects, and an extraction failure replaces the reason phrase. A 4xx or 5xx sent with `Content-Type: application/json` and no body at all trips that path, since a JSON decoder rejects an empty document — common for auth failures. This implementation treats an absent body as having no payload, so the message stays the reason phrase and the detail's `body` is `()`.
- **`xml` is out of scope** — the runtime has no `xml` type, so both `xml` targets and `application/xml` responses return an explanatory `error`. Also not covered: status code response records (`http:StatusCodeClient`, `getStatusCodeRecord()`), the `validation`/`laxDataBinding` client configuration flags, and jBallerina's `data.jsondata` projection options.
- **Existing call sites needed updating.** `var r = check c->get(...)` is now a compile error where no contextual type exists — the same as jBallerina, which requires an explicit `targetType` there. 26 existing corpus and extern test files were updated: 25 call sites retyped to `http:Response`, and 32 `getTextPayload()` sites gained a `check` for the widened return type.
- Public contract validated against jBallerina's `module-ballerina-http` (`2.16.x`): **PASS with notes**. All nine remote methods carry the jBallerina parameter list verbatim, including parameter names, defaults, and the inferred-typedesc shape. The two divergences on the touched surface are the `TargetType` narrowing and the `error` unions above.
- Documented in `doc/library/subset3.md` (new) and the module README's support matrix.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `gofmt` clean; stdlib README format checker passes
- [x] New `corpus/bal/library/subset3/` compiler-pipeline tests (3 files, goldens across all 5 stages) covering `http:Response` passthrough, scalar and record targets, per-method binding, `head` staying unchanged, and the `var`-without-context error
- [x] New `corpus/extern` integration tests against a real `httptest` server (28 routes) covering every builder and error path: JSON to record/array/map/scalar, `text/plain`, form-urlencoded, octet-stream, absent and unrecognised `Content-Type`, narrow targets (enum, closed record, tuple) in both nilable and non-nilable form, 4xx/5xx mapping, unregistered status codes, empty-body nilification per builder, malformed JSON and form bodies, and incompatible target/media-type pairs
- [x] 93.5% statement coverage on `client_binding.go`; remaining uncovered lines are body-read-error branches reachable only through `responseLimits` and defensive nil guards

## Summary by CodeRabbit

* **New Features**
  * HTTP client methods can automatically bind response payloads to expected types, including JSON, text, forms, binary data, collections, and nilable values.
  * Added explicit target-type support with improved status-error, empty-body, and conversion handling.
* **Bug Fixes**
  * Response text and compressed payload reads now consistently report retrieval errors.
  * Improved handling of absent response bodies and status reason phrases.
* **Documentation**
  * Added guidance and examples for response binding, supported targets, and error behavior.
* **Tests**
  * Expanded coverage for successful binding, conversion failures, redirects, and truncated responses.


lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
```diff
@@ -0,0 +1,388 @@
...
+// Mirrors jBallerina's createResponseError. ClientRequestError and RemoteServerError are not
+// declared here, so the distinction survives only in the error's type name.
+func statusCodeError(ctx *extern.Context, types *httpTypes, resp *values.Object, statusCode int) *values.Error {
+	body, extractErr := statusErrorPayload(ctx, types, resp)
+	if extractErr != nil {
+		return values.NewError(semtypes.ERROR, "http:ApplicationResponseError creation failed: "+
+			strconv.Itoa(statusCode)+" response payload extraction failed", extractErr,
+			"PayloadBindingClientError", nil)
+	}
+	tc := ctx.TypeCtx()
```
heshanpadmasiri, line 90: Would it be useful to have a PutUnsafe function for mapping values. Idea is to allow http to insert values to maps by passing type checks, under the assumption type is always valid.

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// The prefix and type name match jBallerina's PayloadBindingClientError, though the distinct
+// type itself is not declared here.
+func payloadBindingError(message string, cause *values.Error) *values.Error {
+	return values.NewError(semtypes.ERROR, "Payload binding failed: "+message, cause,
+		"PayloadBindingClientError", nil)
+}
+
+// The runtime has no xml type, so neither an xml target nor an xml body can be bound.
```
heshanpadmasiri: I don't fully understand this, but runtime do support xml values

TharmiganK: This has to be corrected. The XML full support in HTTP will be addressed separately.

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
TharmiganK: I think that capability already exists — `values.NewMap(ty, atomic, isReadonly, entries)` is documented as constructing a map "without applying any inherent type or readonly checks on the seed entries" , and http already uses the List equivalent in `newListValue`. So I don't think we need another PutUnsafe.

Btw we ideally need to replace all these generic type creation with proper runtime APIs to create typed values and then insert map field values. Issue: https://github.com/ballerina-nutcracker/ballerina/issues/510.

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
TharmiganK: Comments are corrected via [7822121](https://github.com/ballerina-nutcracker/ballerina/pull/705/commits/78221211c7443dc184dbbea533ac3fdde7ff862e) and XML support is tracked via https://github.com/ballerina-nutcracker/ballerina/issues/708

doc/library/subset3.md
```diff
@@ -0,0 +1,92 @@
...
+explicitly in that position:
+
+```ballerina
+var p = check c->get("/person", targetType = Person);
+```
+
+| Feature | Notes |
+|---|---|
+| `http:Response` target | Returned untouched, including for 4xx and 5xx responses. Any union containing `http:Response` behaves the same way |
+| Status code mapping | With any other target, a 4xx or 5xx response returns an `error` whose message is the status code's reason phrase, or `status code <code>` when the code has no registered phrase; 1xx, 2xx, and 3xx responses are bound normally |
```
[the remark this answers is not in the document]
TharmiganK: Error details are not yet visible from Ballerina runtime. We do not need that level of documentation here

corpus/extern/http_client_databind_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"io"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// databindServer serves one canned response per path, each with a Content-Type that selects
+// a particular payload builder.
+func databindServer() *httptest.Server {
```
warunalakshitha: are these also covered with bal tests?

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
```diff
@@ -0,0 +1,412 @@
...
+	tc := ctx.TypeCtx()
+	target := targetArg.(*values.TypeDesc).Type
+	if admitsResponse(tc, target) {
+		return resp
+	}
+	statusCode := responseStatusCode(resp)
+	if statusCode >= 400 && statusCode <= 599 {
+		return statusCodeError(ctx, types, resp, statusCode)
+	}
+	return performDataBinding(ctx, types, resp, target)
```
warunalakshitha: Should we handle the error and read the body and skip for avoid streams being never closed and connections can leak

result := performDataBinding(ctx, types, resp, target)
	if _, isErr := result.(*values.Error); isErr {
		_, _ = responseBody(resp)
	}

corpus/extern/http_client_databind_test.go
TharmiganK: Yes, these are covered

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
TharmiganK: Fixed via [9b1608e](https://github.com/ballerina-nutcracker/ballerina/pull/705/commits/9b1608ed20912e6f489e10ea7e8d34e503cd68b7)

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
```diff
@@ -0,0 +1,412 @@
...
+		return textPayloadBuilder(ctx, types, resp, target, contentType)
+	case urlEncodedContentType.MatchString(contentType):
+		return formPayloadBuilder(ctx, types, resp, target, contentType)
+	case octetStreamContentType.MatchString(contentType):
+		return blobPayloadBuilder(ctx, types, resp, target, contentType)
+	case jsonContentType.MatchString(contentType):
+		return jsonPayloadBuilder(ctx, types, resp, target)
+	default:
+		return builderFromType(ctx, types, resp, target)
+	}
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [b063a1d](https://github.com/ballerina-nutcracker/ballerina/pull/705/commits/b063a1d023c8acd045f1115d7a3e5e6f5f34d4f1)

corpus/extern/http_client_databind_test.go
warunalakshitha: So do we need these go tests?

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+// Picks a builder from the Content-Type as jBallerina does, falling back to the target type
+// when the media type is absent or unknown.
+func performDataBinding(ctx *extern.Context, types *httpTypes, resp *values.Object, target semtypes.SemType) values.BalValue {
+	// jBallerina hands back a non-empty body here even though () was asked for; returning ()
+	// keeps the value within the requested type. The body is still read so a read failure
+	// (for example exceeding responseLimits.maxEntityBodySize) surfaces as an error instead
+	// of being silently discarded, and so the underlying stream is drained and closed.
+	if semtypes.IsSubtype(ctx.TypeCtx(), target, semtypes.NIL) {
```
warunalakshitha: will this work
```
import ballerina/http;
import ballerina/io;
    // Server: 200, Content-Type: text/plain, body: ""
    http:Client c = check new ("http://localhost:8080");
    int? n = check c->get("/empty");
    io:println(n is ()); // expected: true
}
```

lib/stdlibs/ballerina/http/0.0.1/go1.26/http.bal
```diff
@@ -614,12 +619,18 @@ public isolated class Listener {
...
-# `getBinaryPayload()` to extract the response body.
+# **Return type:** Every method except `head` binds the response to the contextually
+# expected type through the `targetType` parameter. `http:Response` (or any union
+# containing it) yields the raw response; any other `anydata` type is deserialised from
+# the response body according to the `Content-Type` header. `xml` targets are not
+# supported. When there is no contextually expected type (for example when assigning to
+# `var`), pass the target explicitly: `c->get("/path", targetType = http:Response)`.
 #
 # **Error types:** Errors are plain `error` values. The upstream distinct error types
-# (`http:ClientError`, `http:HeaderNotFoundError`, etc.) are not declared.
```
warunalakshitha, line 622: I hope we can do this now since distinct error types are supported. not with this PR. may be seperately

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
warunalakshitha, line 1118: can we use decodeJSONBody in getJsonPlayload.

warunalakshitha: do we need to check trailing JSON like client binding logic?

lib/stdlibs/ballerina/http/0.0.1/go1.26/http.bal
TharmiganK: Yes, we can, ideally we need runtime APIs to create distinct errors by specifying the type name

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
TharmiganK: Not working at the moment, but it should ideally work. Will fix this

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
TharmiganK: Yes, we can use it, will fix this

TharmiganK: We are going to reuse decodeJSONBody, so yes, it will check trailing JSON

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/client_binding.go
TharmiganK: Fixed via [152f24e](https://github.com/ballerina-nutcracker/ballerina/pull/705/commits/152f24ebd39c0e24cf1087b540491c33c4ae84e0)

lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_native.go
TharmiganK: Fixed via [0dadc3e](https://github.com/ballerina-nutcracker/ballerina/pull/705/commits/0dadc3e62e14da9e22343571875980576a062258)

TharmiganK: Reusing the same decoder for JSON via [0dadc3e](https://github.com/ballerina-nutcracker/ballerina/pull/705/commits/0dadc3e62e14da9e22343571875980576a062258)

warunalakshitha (approved):
LGTM

PR #712, opened 2026-08-05
feat(stdlib): add `ballerina/mime` standard library

TharmiganK:
Part of: https://github.com/ballerina-nutcracker/ballerina/issues/492

## Summary

Ports the `ballerina/mime` standard library to the Go-native interpreter — MIME (RFC 2045/2046) media types, content disposition, entity headers/metadata, text/JSON/XML/byte-array/file-backed bodies with lazy cross-conversion, multipart body parts, and Base64 encoding/decoding.

- Parse and construct `MediaType` (primary type, sub-type, suffix, parameters) and `ContentDisposition` (filename, name, parameters)
- Manage `Entity` headers (`setHeader`/`getHeader`/`getHeaders`/`addHeader`/`removeHeader`/`hasHeader`) and content metadata (content type, ID, length, disposition)
- Set/get an `Entity` body as text, JSON, XML, or byte array — every accessor cross-converts from whatever the body was actually set as, matching jBallerina's data-source model
- Set an `Entity` body directly from a file (`setFileAsEntityBody`) as a lazy byte-channel data source, read (and cached) on demand — implemented in `.bal` the same way jBallerina does it, calling into `io:openReadableFile`
- Set/get multipart body parts as `Entity[]` (`setBodyParts`/`getBodyParts`); a part whose own body is itself multipart decodes correctly via an explicit nested `getBodyParts()` call, matching jBallerina's own non-auto-recursive design
- Base64 encode/decode for strings, byte arrays, and `io:ReadableByteChannel`

## Example

```ballerina
import ballerina/io;
import ballerina/mime;

    mime:MediaType mt = check mime:getMediaType("application/json; charset=UTF-8");
    io:println(mt.primaryType);           // application
    io:println(mt.getBaseType());         // application/json
    io:println(mt.parameters["charset"]); // UTF-8

    mime:Entity entity = new ();
    entity.setHeader("Content-Type", mime:APPLICATION_JSON);
    io:println(entity.hasHeader("content-type")); // true

    entity.setText("Hello");
    string|mime:ParserError text = entity.getText();
    if text is string {
        io:println(text); // Hello
    }

    string|byte[]|mime:EncodeError encoded = mime:base64Encode("Hello");
    if encoded is string {
        io:println(encoded); // SGVsbG8=
    }
}
```

## Notes

- Public contract validated against jBallerina's `module-ballerina-mime` — no breaking changes. `Entity.setBody`'s union is narrower by the already-established `stream` omission elsewhere in this port; everything else matches jBallerina's declared signatures exactly.
- `EntityBodyKind`/`EntityBody` are exported from the native package so sibling stdlibs that construct or read `mime:Entity` values natively (e.g. `http`, for multipart request/response bodies) can reuse this representation instead of duplicating it.
- Notable behavioural change: a missing multipart boundary is a `ParserError` rather than jBallerina's silent empty array — this port lets a manually-constructed `Entity` decode multipart content at all, which jBallerina's public API can't exercise (it only decodes from a byte-channel data source the HTTP transport sets up internally). The same debuggable-over-silent-empty choice applies to `message/*` bodies, which jBallerina also never actually decodes (verified against a live `bal run` HTTP round trip) but returns silently empty for; this port returns an explicit `ParserError` instead.
- `mime:Error` and its subtypes are plain `error` aliases rather than `distinct error` (interpreter limitation), same divergence already documented for other stdlibs.
- See `lib/stdlibs/ballerina/mime/0.0.1/go1.26/README.md` for the full support matrix.

## Test plan

- [x] `make build`
- [x] `make vet`
- [x] Corpus integration tests pass (`corpus/lib/subset3/`): `mime-media-type1-v`, `mime-content-disposition1-v`, `mime-entity-headers1-v`, `mime-entity-metadata1-v`, `mime-entity-body1-v`, `mime-entity-xml1-v`, `mime-entity-file-missing-p`, `mime-base64-1-v`, `mime-constants1-v`, `mime-multipart-v`, `mime-multipart-error-v`
- [x] Native coverage on touched `native/` lines confirmed via `go tool cover`

## Summary by CodeRabbit

* **New Features**
  * Added the `ballerina/mime` standard library with MIME constants, media-type and content-disposition parsing, entity headers and metadata, and body accessors.
  * Added XML and file-backed entity bodies, lazy channel-based body handling, and multipart form-data support.
  * Added Base64 encoding and decoding for strings, byte arrays, and readable channels, including charset support.
  * Added clear errors for invalid content types, malformed data, missing files, and unsupported message bodies.

* **Documentation**
  * Documented MIME capabilities, supported behaviors, and usage examples.

* **Tests**
  * Added comprehensive coverage for MIME entities, XML, multipart data, headers, metadata, Base64, and error handling.


lib/stdlibs/ballerina/mime/0.0.1/go1.26/native/mime.go
```diff
@@ -0,0 +1,638 @@
...
+
+const (
+	orgName    = "ballerina"
+	moduleName = "mime"
+)
+
+// EntityBodyKind identifies the type of body content stored on an Entity.
+// Exported so sibling stdlibs (e.g. http) that construct or read mime:Entity
+// values natively can do so without duplicating this representation.
+type EntityBodyKind int
```
ThisaruGuruge: Do we need to export this? Asking because the jBallerina does not have a similar public type.

lib/stdlibs/ballerina/mime/0.0.1/go1.26/native/multipart.go
```diff
@@ -0,0 +1,338 @@
...
+		"setBody", "setJson", "getJson", "setXml", "getXml", "setText", "getText", "setByteArray", "getByteArray",
+		"getHeader", "getHeaders", "getHeaderNames", "addHeader", "setHeader",
+		"removeHeader", "removeAllHeaders", "hasHeader", "setBodyParts", "getBodyParts",
+	}
+	keys := make(map[string]string, len(names))
+	for _, n := range names {
+		keys[n] = "ballerina/mime:Entity." + n
+	}
+	return keys
+}
```
ThisaruGuruge, line 52: Do we need this as a function? Seems like a static map, which can be defined as a const or at least a package-level map. WDYT?

lib/stdlibs/ballerina/mime/0.0.1/go1.26/native/multipart.go
```diff
@@ -0,0 +1,338 @@
...
+// as multipart at all — only an inbound request/response body is eligible there).
+func DecodeMultipart(ctx *extern.Context, data []byte, boundary string) ([]*values.Object, error) {
+	if boundary == "" {
+		return nil, fmt.Errorf("no boundary parameter found in Content-Type")
+	}
+	reader := multipart.NewReader(bytes.NewReader(data), boundary)
+	var parts []*values.Object
+	for {
+		part, err := reader.NextPart()
+		if err == io.EOF {
```
ThisaruGuruge, line 183: Isn't the best practice to use this?


```suggestion
		if errors.Is(err, io.EOF) {
```

PR #718, opened 2026-08-06
refactor: break project in to multiple go modules

### LINKED ISSUE #675: [Task]: Convert the single module to multiple modules in a workspace

heshanpadmasiri:
## Summary
- split top-level packages into separate Go modules
- add a Go workspace and shared Makefile for repository-wide commands
- update imports, tests, and CI workflows for the multi-module layout

This is the stacked replacement for #680; the original PR remains open during the transition.

Closes #675


ast/interpreter_source.go
```diff
@@ -0,0 +1,31 @@
...
+
+package ast
+
+import (
+	"embed"
+	"io/fs"
+)
+
+//go:embed all:*
+var interpreterSource embed.FS
```
warunalakshitha: is that correct to user word "interpreterSource" here. can we rename to astSource and ast_source.go as file name. Similarly other files as well.

interpsrc/interpsrc.go
warunalakshitha, line 1: Can we move this to separate go module.

lib/go.sum
```diff
@@ -0,0 +1,16 @@
+github.com/cockroachdb/apd/v3 v3.2.3 h1:4Zx+I3R35bFXMnltzmjP79i2cravE4jTRL6ps9Aux80=
+github.com/cockroachdb/apd/v3 v3.2.3/go.mod h1:klXJcjp+FffLTHlhIG69tezTDvdP065naDsHzKhYSqc=
+github.com/lib/pq v1.10.7 h1:p7ZhMD+KsSRozJr34udlUrhboJwWAgCg34+/ZZNvZZw=
+github.com/lib/pq v1.10.7/go.mod h1:AlVN5x4E4T544tWzH6hKfbfQvm3HdbOxrmggDNAPY9o=
+github.com/santhosh-tekuri/jsonschema/v6 v6.0.2 h1:KRzFb2m7YtdldCEkzs6KqmJw4nqEVZGK7IN2kJkjTuQ=
+github.com/santhosh-tekuri/jsonschema/v6 v6.0.2/go.mod h1:JXeL+ps8p7/KNMjDQk3TCwPpBy0wYklyWTfbkIzdIFU=
+github.com/sergi/go-diff v1.4.0 h1:n/SP9D5ad1fORl+llWyN+D6qoUETXNZARKjyY2/KVCw=
+github.com/sergi/go-diff v1.4.0/go.mod h1:A0bzQcvG0E7Rwjx0REVgAGH58e96+X0MeOfepqsbeW4=
+golang.org/x/crypto v0.54.0 h1:YLIA59K4fiNzHzjnZt2tUJQjQtUWfWbeHBqKtk3eScw=
```
warunalakshitha: go mod has golang.org/x/crypto v0.52.0 but this golang.org/x/crypto v0.54.0 h1:YLIA59K4fiNzHzjnZt2tUJQjQtUWfWbeHBqKtk3eScw=. should we fix that. same thing on other module go.sum as well

interpsrc.go
```diff
@@ -119,38 +160,100 @@ func extractDev() (string, error) {
...
+	./projects
+	./runtime
+	./semantics
+	./semtypes
+	./tools
+	./values
+)
+
+replace (
+	ballerina v0.6.0 => .
```
warunalakshitha: should we chnage these to github.com/ballerina-nutcracker/ballerina v0.6.0 => .

lib/go.sum
heshanpadmasiri: fixed

ast/interpreter_source.go
heshanpadmasiri: I am more in favour of it being the same name. Every module will have exactly the same file, so IMO it is better for it to be same name as well.

interpsrc/interpsrc.go
heshanpadmasiri: fixed

interpsrc.go
heshanpadmasiri: fixed

warunalakshitha (dismissed):
LGTM

ast/interpreter_source.go
heshanpadmasiri: Actually I think we can avoid the embedded every thing except driver now that every thing else is a separate module driver depends on. They'll get pulled in just like driver currently pull in third party dependencies.

heshanpadmasiri: fixed in 0473fb48

warunalakshitha (approved):
LGTM

PR #715, opened 2026-08-06
perf: add start wait benchmarks

heshanpadmasiri:
## Purpose
$subject
Depends on #683 

## Summary by CodeRabbit

* **New Features**
  * Added benchmarks covering chained, fan-out, alternating, and multiple future waits.
  * Added isolated and non-isolated concurrency scenarios with validated checksum output.

* **Tests**
  * Added expected-output checks for successful standard output and empty standard error across all benchmark scenarios.
  * Added coverage for large future chains and concurrent task collections.

PR #726, opened 2026-08-06
fix(runtime): race in concurrent access to isolated variables

### LINKED ISSUE #700: [Bug] Concurrent access to module-level globals races on the underlying Go map
## Problem

`BIRModule.Globals` (`runtime/internal/modules/module.go`) stores every module-level variable's value in a plain `map[string]values.BalValue`, and it is read/written with no synchronization at all from `runtime/internal/exec/frame.go`:

```go
func getOperandValue(ctx *extern.Context, op *bir.BIROperand, currentFrame *Frame) values.BalValue {
	if gv, ok := op.VariableDcl.(*bir.BIRGlobalVariableDcl); ok {
		module := getModule(ctx, gv.PkgId)
		return module.Globals[gv.GlobalVarLookupKey] // unsynchronized read
	}
	return Load(currentFrame, op.Address)
}

func setOperandValue(ctx *extern.Context, op *bir.BIROperand, currentFrame *Frame, value values.BalValue) {
	if gv, ok := op.VariableDcl.(*bir.BIRGlobalVariableDcl); ok {
		module := getModule(ctx, gv.PkgId)
		module.Globals[gv.GlobalVarLookupKey] = value // unsynchronized write
	} else {
		Store(currentFrame, op.Address, value)
	}
}
```

An `isolated` module-level variable's `lock { }` block does compile to a real mutex acquire, keyed by that variable's name, so concurrent access *to that one variable* is correctly serialized at the language level. But that says nothing about the underlying Go map: two *different* `isolated` variables, each perfectly protected by its own lock, still write to the same shared Go map concurrently when accessed from two different strands — and Go maps are not safe for concurrent read/write across different keys, let alone a concurrent read and write to the same key racing during a rehash. In principle this can manifest as silent corruption or a hard `fatal error: concurrent map read and map write` crash, not just a race-detector warning.

This surfaces any time two strands run concurrently against the same module and touch more than one module-level variable between them — e.g. two concurrent inbound requests to an HTTP service with two `isolated` globals, one written per resource. There's already a nightly CI job (`.github/workflows/nightly-race.yml`) that runs the full suite under `-race` and auto-files an issue on failure, so this is a live gate, not a theoretical concern.

## Reproduction

Confirmed on `main` @ `44134e8c1`, using only the already-merged `ballerina/http` module — no other dependencies.

`race.bal`:
```ballerina
import ballerina/http;

isolated int counterA = 0;
isolated int counterB = 0;

service /race on new http:Listener(9099) {
    resource function get a() returns int {
        lock {
            counterA += 1;
            return counterA;
        }
    }
    resource function get b() returns int {
        lock {
            counterB += 1;
            return counterB;
        }
    }
}
```

```bash
go build -race -o bal-race ./cli/cmd
./bal-race run race.bal &

# fire concurrent requests at two different resources
for i in $(seq 1 300); do
  curl -s "http://localhost:9099/race/a" >/dev/null &
  curl -s "http://localhost:9099/race/b" >/dev/null &
done
wait
```

This reliably trips `WARNING: DATA RACE` within the first ~100 requests (under 2 seconds locally). The report:

```
WARNING: DATA RACE
Read at 0x00c000ae55f0 by goroutine 842:
  runtime.mapdelete_fast64()
  ballerina/runtime/internal/exec.getOperandValue()
      runtime/internal/exec/frame.go:54 +0x1f8
  ...
  ballerina/lib/stdlibs/ballerina/http/0.0.1/go1.26/native.dispatchRequest()
      lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go:530 +0x3cc
  ...

Previous write at 0x00c000ae55f0 by goroutine 843:
  runtime.mapaccess2_faststr()
  ballerina/runtime/internal/exec.setOperandValue()
      runtime/internal/exec/frame.go:62 +0x200
  ...
  ballerina/lib/stdlibs/ballerina/http/0.0.1/go1.26/native.dispatchRequest()
      lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go:530 +0x3cc
  ...
```

Both goroutines are ordinary concurrent HTTP requests (one per resource), each on its own `net/http` connection goroutine.

## Proposal

Guard `BIRModule.Globals` with a `sync.RWMutex` (read lock in `getOperandValue`, write lock in `setOperandValue`), independent of and in addition to the existing per-variable `lock { }` semantics, which remain necessary for the logical atomicity of a variable's own read-modify-write sequence. A single mutex per module is coarse but correct and simple; if it becomes a contention bottleneck it can be refined later (e.g. pre-resolving each global to a fixed slice index at module-load time so distinct variables no longer share any mutable Go data structure).

## Impact

This is a correctness bug, not a benign race-detector artifact: any concurrently-dispatched service (HTTP or otherwise) touching more than one module-level variable can hit undefined behavior up to and including a process crash. It will also fail the nightly race-detector CI job the first time a test exercises this concurrency shape.

### LINKED ISSUE #916: [Bug]: Concurrent access to object fields races on the underlying Go map
### Description

Two strands mutating **different** fields of the same `isolated` object race on the Go map backing that object's fields, even though the Ballerina program is fully correct and every access is inside a `lock { }` block.

This is not something a user can fix. `lock` on an isolated object's field is keyed **per field**, so `self.counterA` and `self.counterB` are protected by two different locks — but both values live in one shared `map[string]BalValue` inside `values.Object`. Two strands under two different locks still read and write that one Go map concurrently, which is undefined behavior in Go: silent corruption or a hard `fatal error: concurrent map read and map write`, not merely a race-detector warning.

This is the object-field counterpart of #700, which covers the same defect for module-level globals.

### Steps to Reproduce

`counters.bal`:

```ballerina
import ballerina/io;

isolated class Counters {
    private int counterA = 0;
    private int counterB = 0;

    isolated function incrementA() returns int {
        lock {
            while self.counterA < 10000 {
                self.counterA += 1;
            }
            return self.counterA;
        }
    }

    isolated function incrementB() returns int {
        lock {
            while self.counterB < 10000 {
                self.counterB += 1;
            }
            return self.counterB;
        }
    }
}

public function main() {
    Counters counters = new ();
    future<int> a = start counters.incrementA();
    future<int> b = start counters.incrementB();
    int|error aResult = wait a;
    int|error bResult = wait b;
    io:println(aResult); // expected: 10000
    io:println(bResult); // expected: 10000
}
```

Run the runtime under the race detector. Observed:

```
WARNING: DATA RACE
Read at 0x00c000c27860 by goroutine 1108:
  runtime.mapaccess1_faststr()
  values.(*Object).Get()          values/object.go:105
  exec.execObjectLoad()           runtime/internal/exec/non_terminators.go:180

Write at 0x00c000c27860 by goroutine 1109:
  runtime.mapaccess2_faststr()
  values.(*Object).Put()          values/object.go:101
  exec.execObjectStore()          runtime/internal/exec/non_terminators.go:174
```

### Possible cause

`Object.fields` is a plain `map[string]BalValue` read and written by `Object.Get`/`Object.Put` with no synchronization. Because the language-level lock is per field rather than per object, no user-level locking serializes access to that map. Replacing it with a `sync.Map` (or another concurrency-safe representation) removes the race; this was verified by reverting `values/object.go` alone, which reproduces the trace above, and restoring it, which passes clean under `-race`.

### Related area

Runtime

### Related issue(s)

#700 — the same defect for module-level globals. Fixed by #726.

heshanpadmasiri:
## Purpose

Module-level globals and object fields are each stored in a plain Go map that is
read and written with no synchronization. Because the language-level `lock` is
keyed per variable / per field rather than per map, two strands under two
different locks still touch the same Go map concurrently. That is undefined
behavior in Go, up to a hard `fatal error: concurrent map read and map write` --
and no amount of user-level locking can prevent it.

Both halves of this PR are the same defect in two storage locations, so they are
fixed together:

- `runtime/internal/modules/module.go` -- module globals move to a `sync.Map`
  behind `GetGlobal`/`SetGlobal`. Resolves #700.
- `values/object.go` -- object fields move to a `sync.Map` behind the existing
  `Get`/`Put`. Resolves #916.

Per-variable `lock { }` semantics are unchanged; they remain necessary for the
logical atomicity of a variable's own read-modify-write sequence. This PR only
makes the underlying storage safe to touch from more than one strand.

Both cases have corpus tests (`concurrent-module-globals-v`,
`concurrent-object-fields-v`) that fail with a `DATA RACE` under `-race` before
the change and pass after it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrency safety when accessing and updating module-level global values.
  * Improved concurrency safety for object field access and updates.
  * Improved annotation value resolution during execution.
  * Corrected runtime behavior for concurrently updated globals and object fields, ensuring expected results without errors.

* **Tests**
  * Added coverage for concurrent updates to module-level globals and object fields, including verification of expected output.


heshanpadmasiri:
[AI RESPONSE] On the pre-merge check warnings (re: https://github.com/ballerina-nutcracker/ballerina/pull/726#issuecomment-5202180810):

**Linked Issues check** — ignoring. The check reads `#700`'s Proposal as requiring a module-level `sync.RWMutex`. That section presents the single mutex as one option, not a requirement: *"A single mutex per module is coarse but correct and simple; if it becomes a contention bottleneck it can be refined later."* `sync.Map` is exactly such a refinement and satisfies the stated correctness requirement — the unsynchronized Go map accesses in `getOperandValue`/`setOperandValue` are gone, and the per-variable `lock { }` semantics that `#700` says "remain necessary" are untouched.

**Out of Scope Changes check** — addressed. The `values/object.go` change is the same defect in a second storage location, not a separate concern, and it now has its own linked issue: `#916`. Verified by reverting `values/object.go` alone, which reproduces a `DATA RACE` in `Object.Get`/`Object.Put` from `execObjectLoad`/`execObjectStore` for a fully-`lock`-guarded `isolated` class, and restoring it, which passes clean under `-race`. Because the language-level lock is keyed per field while all fields share one Go map, no user-level locking can serialize those map accesses. The PR description has been rewritten to state this and the stale `#715` reference dropped.

**Docstring Coverage** — fixed. Doc comments added for `NewBIRModule`, `GetGlobal`, `SetGlobal` and `SetGlobals`.

heshanpadmasiri:
[AI RESPONSE] failed to reproduce the remaining Out of Scope Changes warning (re: https://github.com/ballerina-nutcracker/ballerina/pull/726#issuecomment-5202180810).

The warning cites "the annotation behavior changes and new TypeAnnotations registration". Neither is in this PR:

- `git diff main...HEAD | grep -i TypeAnnotations` returns nothing.
- The only annotation-related change in the PR diff is a single mechanical call-site update in `runtime/internal/exec/annotations.go`, replacing `module.Globals[ref.GlobalLookupKey()]` with `module.GetGlobal(ref.GlobalLookupKey())` — required because the field became private.

The cause looks like the review range rather than the PR contents. The recent review reports it reviewed "between `a63608788` and `e4445c267`", and this branch was rebased onto `main` between those two commits. That range therefore spans 9 upstream `main` commits, including `3b2936467` ("fix(annotations): propagate field annotations through aliases") and `d279c2d9c` ("fix(annotations): return partial results and cover aliases in pool") from the already-merged `#783`. Those are the annotation-resolution and TypeAnnotations-registration changes the check is flagging, and they belong to `main`, not to this PR.

A review scoped to `main...HEAD` rather than to the incremental commit range should clear this.

warunalakshitha (approved):
LGTM

PR #730, opened 2026-08-07
fix(annotations): expose service and parameter metadata at runtime

### LINKED ISSUE #716: bug: language-level blockers for resource-parameter annotations (`@http:Payload`, `@http:Query`, `@http:Header`)
Five separate defects block annotation-driven data binding on HTTP service resources. They are filed together because they were found as one investigation and share reproducers, but they are independently fixable and can be split into separate issues if that suits ownership better. Blockers 1–3 must all land before `ballerina/http` can implement `@http:Payload` / `@http:Query` / `@http:Header` at all; 4 and 5 bound what the feature can cover once it exists.

Every reproduction below was run against `main` @ `c5ce6e641` with a stock `go build -o bal ./cli/cmd`. The quoted output is observed, not predicted.

**Summary**

| # | Defect | Blocks |
|---|---|---|
| 1 | A record-valued annotation is rejected on a parameter but accepted everywhere else | All three HTTP annotations — they do not compile |
| 2 | Annotations inside a service declaration body are never resolved or validated | Annotations in the idiomatic service form are silently ignored |
| 3 | Parameter annotation values are discarded after type resolution and are unreachable at runtime | Any stdlib reading a parameter annotation |
| 4 | The compiler panics with a Go stack trace on an annotated service declaration | `@http:ServiceConfig`, `@http:ResourceConfig`, `@http:Cache` |
| 5 | Defaultable parameters in resource methods are treated as required | Defaultable query parameters (http spec §2.3.4.3) |

---

### Problem

`semantics/constant_evaluator.go:113-137` `evaluateMappingConstructor` reads the mapping constructor's determined type and bails when it is not an atomic mapping type:

```go
ty := expr.GetDeterminedType()
atomic := semtypes.ToMappingAtomicType(e.resolver.typeContext(), ty)
if atomic == nil {
    return nil, fmt.Errorf("constant mapping type is not atomic")
}
```

It is reached from `semantics/type_resolver.go:1286` (`evaluateAnnotationValue`) inside `resolveAnnotationAttachments`. For a parameter attach point the determined type is not an atomic mapping type, so **every record-valued annotation on a parameter is a compile error**. The same annotation type with the same value on a `type` compiles and round-trips to the runtime.

The corpus misses this because the only two parameter-position annotations in the tree — `corpus/bal/subset9/09-annotation/3-v.bal:36` and `15-v.bal:43` — are bare marker annotations, which take the `prepareImplicitAnnotationValue` path (`type_resolver.go:1367-1370`) and never build a mapping constructor. Note that `15-v.bal:42` uses `@methodMeta {name: "add"}` on an *object function* and that works, so the failure is specific to the parameter position rather than to invokable nodes in general.

**Hypothesis for whoever picks this up** (not confirmed): parameters are `BLangSimpleVariable`s inside a function signature, and `resolveFunctionSignature` runs at `type_resolver.go:1023`, *before* `resolveTopLevelAnnotationAttachments` at `:1041`. If that earlier pass already stamps the annotation expression's determined type (for example `NEVER` via `setOtherNodesAsNever`), the later `resolveActionOrExpression(t, nil, ann.Expr, valueType)` at `:1276` would short-circuit and leave the wrong type behind. That would explain why object fields and object functions — whose annotations are not inside a pre-walked signature — are unaffected.

### Reproduction

`param-annot.bal`:
```ballerina
import ballerina/io;

type Meta record {|
    string name;
|};

annotation Meta onType on type;
annotation Meta onParam on parameter;

// Works: the identical record value on a type.
@onType {name: "type-ok"}
type Target int;

// Fails: the identical record value on a parameter.
function add(@onParam {name: "param-fails"} int left, int right) returns int {
    return left + right;
}

public function main() {
    io:println(add(1, 2));
}
```

```bash
go build -o bal ./cli/cmd
./bal run param-annot.bal
```

```
error[SEMANTIC_ERROR]: cannot evaluate annotation constant expression: constant mapping type is not atomic
  --> param-annot.bal:15:23
   |
15 | function add(@onParam {name: "param-fails"} int left, int right) returns int {
   |                       ^^^^^^^^^^^^^^^^^^^^^
```

Delete the `add` function and the file compiles and runs, proving the type-attached value is fine.

### Proposal

Trace why the mapping constructor's determined type differs between the parameter and type attach points, and make the parameter path resolve against the annotation's value type the same way. Add corpus coverage for a record-valued annotation on `parameter` and on `return` alongside the fix — `corpus/bal/subset9/09-annotation` currently has none, which is why this went unnoticed.

### Impact

This blocks the entire `ballerina/http` service data-binding surface: `@http:Payload {mediaType: ...}`, `@http:Header {name: "X-Api-Key"}` and `@http:Query {name: "..."}` are all record-valued parameter annotations and none of them compile. It equally blocks any user-defined parameter annotation that carries configuration, which is most of the reason parameter annotations exist.

---

### Problem

`resolveTopLevelAnnotationAttachments` (`semantics/type_resolver.go:1139-1198`) iterates `pkg.Annotations`, `pkg.TypeDefinitions`, `pkg.ClassDefinitions`, `pkg.Functions`, `pkg.Constants` and `pkg.GlobalVars` — but **not `pkg.Services`** (`ast.BLangPackage.Services []BLangService`, `ast/ast.go:177`).

A `service /x on new http:Listener(...)` declaration produces a `BLangService`, not a `BLangClassDefinition`, so annotations on its resource methods, regular methods, fields and parameters are never reached: not type-checked, not attach-point-validated, not evaluated. The equivalent `service class` **is** covered (`type_resolver.go:1160-1179`), so the two service forms disagree.

### Reproduction

Two files differing only in the service form.

`svc-class.bal` — control, correctly rejected:
```ballerina
import ballerina/io;

type OnTypeOnly record {|
    string name;
|};

annotation OnTypeOnly info on type;

service class Svc {
    @info {name: "nope"}
    resource function get foo() returns string {
        return "ok";
    }
}

public function main() {
    io:println("compiled");
}
```

`svc-decl.bal` — same misuse, silently accepted:
```ballerina
import ballerina/http;
import ballerina/io;

type OnTypeOnly record {|
    string name;
|};

annotation OnTypeOnly info on type;

service /svc on new http:Listener(9401) {
    @info {name: "nope"}
    resource function get foo() returns http:Response {
        http:Response r = new;
        r.setTextPayload("ok");
        return r;
    }
}

public function testMain() {
    io:println("compiled");
}
```

```bash
go build -o bal ./cli/cmd
./bal run svc-class.bal      # expected: diagnostic
./bal run svc-decl.bal       # expected: the same diagnostic
```

`svc-class.bal`:
```
error[SEMANTIC_ERROR]: annotation 'info' is not allowed on objectfunction
  --> svc-class.bal:11:5
   |
11 |     @info {name: "nope"}
   |     ^^^^^^^^^^^^^^^^^^^^
```

`svc-decl.bal` produces **no diagnostic at all** and proceeds to the listening phase.

This also currently masks defect 1 inside a service declaration — the value is never evaluated, so no error surfaces. Fixing this will expose defect 1 there.

### Proposal

Add a `pkg.Services` loop after the `pkg.ClassDefinitions` loop (after `:1179`) and before the `initialGlobalCount` reorder at `:1191`, covering members only: `svc.Fields` (`Point_OBJECT_FIELD`), `svc.InitFunction` and `svc.Methods` (`resolveFunctionAnnotationAttachments(..., true)`), and `svc.ResourceMethods` (`resolveInvokableAnnotationAttachments(..., Point_OBJECT_METHOD)`). `BLangService` embeds `classDefnBase` (`ast/ast.go:130-161`), so all four fields are already available.

Do **not** resolve `svc.AnnAttachments` (the declaration's own annotations) as part of this: those are rejected earlier by the panic in defect 4, and `BLangService` has no symbol field to key a value on — only `BLangClassDefinition` adds `symbol model.SymbolRef` (`ast/ast.go:149-153`).

Iterate `svc.Methods` through a sorted key slice. It is a Go map, and `$annotation$N` global names are allocated from a source-visit counter (`type_resolver.go:1410-1444`), so map order would make the generated names — and therefore `bal build` output — nondeterministic. The existing `classDef.Methods` loop at `:1173` has the same latent problem and is worth fixing at the same time.

### Impact

Any annotation written inside the idiomatic `service /x on <listener>` form is silently ignored, including invalid usage that should be a compile error. Once `ballerina/http` starts reading parameter annotations, a user's `@http:Payload` inside a service declaration would be dropped with no diagnostic.

---

### Problem

Three gaps compound; fixing any one alone changes nothing observable.

**a. Values are dropped in `semantics`.** `type_resolver.go:1303`:

```go
storedOnType := typeSymbol != (model.SymbolRef{}) && sym.IsRuntimeVisibleAt(pointKey)
```

Only type definitions (`defn.Symbol()`, `:1146`) and class definitions (`classDef.Symbol()`, `:1166`) pass a real symbol. `resolveInvokableAnnotationAttachments` passes `model.SymbolRef{}` for parameters (`:1219`, `:1222`), functions (`:1217`) and returns (`:1225`), so the evaluated value is computed and thrown away. It is written to `ann.AnnotationValue` at `:1301`, but that field has no readers anywhere in the tree.

**b. BIR has no carrier.** `bir/model.go` — `BIRParameter{BIRNodeBase; Name model.Name; Flags model.Flag}`. `BIRFunctionParameter` (`:127`, `:156-159`) looks like the intended slot but is dead code: never populated by `bir_gen.go`, never serialised by `bir/codec`.

**c. No runtime accessor.** The `ANNOT_ACCESS` instruction (`runtime/internal/exec/binary.go:248-274`) requires its operand to be a `*values.TypeDesc`; there is no analogue of jBallerina's `MethodType.getAnnotation("$param$.<name>")`. Native stdlib code cannot reach `bir.BIRFunction` either — `runtime/internal/modules` is internal, and `runtime/extern` exposes only an `extraArgs int` count from `LookupResourceMethodByPath` (`runtime/extern/handle.go:81`), with no parameter names, types or annotations.

### Reproduction

`param-annot-runtime.bal`:
```ballerina
import ballerina/io;

type Meta record {|
    string name;
|};

annotation Meta onType on type;
annotation onParam on parameter;   // marker only — a record value here hits defect 1

@onType {name: "on-the-type"}
type Target int;

function add(@onParam int left, int right) returns int {
    return left + right;
}

public function main() {
    Meta? m = Target.@onType;
    if m is Meta {
        io:println(m.name);        // prints: on-the-type
    }
    io:println(add(1, 2));         // prints: 3
    // There is no expression that can observe the annotation on `left`,
    // and no native API that can either.
}
```

```bash
go build -o bal ./cli/cmd
./bal run param-annot-runtime.bal
./bal run --dump-bir param-annot-runtime.bal   # `add` carries no annotation data
```

The type annotation round-trips; the parameter annotation is unobservable by construction, and the BIR dump for `add` shows parameter types only because `BIRParameter` has nowhere to record annotations.

### Proposal

In dependency order:

- **`semantics/type_resolver.go:1219,1222`** — pass `parameter.Symbol()` / `restParam.Symbol()`. The existing gate then routes the value through `setTypeAnnotationValue` → `CompilerEnvironment.symbolAnnotations` (`context/env.go:110`), which is already keyed by `model.SymbolRef`. Rename `typeSymbol`→`ownerSymbol`, `storedOnType`→`storeValue`, `setTypeAnnotationValue`→`setSymbolAnnotationValue` to match, and keep the zero-symbol guard so the function and return call sites are unaffected.

  Reusing this store rather than the AST field is required, not cosmetic: the `RuntimeAnnotationValueRef` global-lifting path (`:1313`, `:1410`) and the repeated-annotation grouping (`:1304-1340`) both write only to the store, and `sym.IsRuntimeVisibleAt` — the gate that keeps `source`-only annotations out of the runtime — lives inside the same predicate.

- **`bir/model.go`** — add `Annotations values.AnnotationValues` to `BIRParameter`; populate in `bir_gen.go:466-480` and `:1941-1955` (the path-param slots built at `:1928-1940` stay nil). Do not add a `Type` field: parameter types already live in `LocalVars` and round-trip through the codec. Add a `(*BIRFunction).ParamLocalVarOffset()` helper and route `executor.go:79`, `pretty_print.go:102` and `deserializer.go:432` through it — see the related `self` parameter-offset issue.

- **`bir/codec/{serializer,deserializer}.go`** — round-trip alongside `RequiredParams` (`serializer.go:179-186`, `deserializer.go:318-335,430-437`) using the existing `writeConstValue`/`readConstValue` machinery, which already covers the whole annotation-value domain including `RuntimeAnnotationValueRef`. **Write keys sorted**, as `symbolpool/annotation_values.go:44-50` and `serializer.go:620-628` already do, or `bal build` output stops being byte-reproducible.

- **`runtime/extern/handle.go`** — a `ParamDescriptor{Name, Type, Annotations}` / `ResourceSignature` pair, a `DispatchHandles.ResourceParams` hook and a `Context.ResourceSignature(MethodHandle)` method; implemented in `runtime/internal/exec/dispatch.go` off the `*values.ResourceEntry` captured by `newResourceHandle` (`runtime/internal/exec/handle.go:76`). Share the path-bound-prefix arithmetic with `resourceExtraArgCount` (`dispatch.go:112-135`) so the descriptor count and `extraArgs` cannot drift.

- Extract the `RuntimeAnnotationValueRef` dereference from `binary.go:262-273` into a shared helper returning `(value, ok)`. `execBinaryOpAnnotAccess` should keep its panic — that is correct inside the interpreter — but the dispatch accessor runs on a caller's Go goroutine outside any strand trap frame, where a panic kills the caller instead of raising a Ballerina error.

**Concurrency note.** `SetSymbolAnnotationValue` (`context/env.go:116-119`) mutates the inner map without a lock, and is safe only under the invariant documented at `env.go:106-109`: all writes happen single-threaded in stage 4, reads happen later. Stage 10 (`bir_gen`) runs concurrently per module and would now *read* this map for every parameter, so every write must stay inside `resolveTopLevelAnnotationAttachments` — no writes from `bir_gen`, no lazy population on first read.

### Impact

No stdlib can implement annotation-driven behaviour on parameters. Concretely this blocks `@http:Payload`, `@http:Query` and `@http:Header` in `ballerina/http`, and the same shape is needed by `graphql`, `grpc` and `websocket` when they are ported.

---

### Problem

`ast/node_builder.go:1796` in `TransformServiceDeclaration` calls `panic("TransformServiceDeclaration: annotations not yet supported")` when the service declaration's metadata carries any annotation. An unimplemented feature should produce a diagnostic, not a Go panic — the panic reaches the user as a raw stack trace with no source location. There are sibling panics for the same reason at `:1727` (listener declarations), `:3281` (type-cast params), `:3359` (`let` variables) and `:4247` (function-type params).

### Reproduction

`svc-annot.bal`:
```ballerina
import ballerina/http;

type Cfg record {|
    string host;
|};

annotation Cfg serviceCfg on service;

@serviceCfg {host: "localhost"}
service /svc on new http:Listener(9402) {
    resource function get foo() returns http:Response {
        return new;
    }
}
```

```bash
go build -o bal ./cli/cmd
./bal run svc-annot.bal
```

```
panic: TransformServiceDeclaration: annotations not yet supported

goroutine 1 [running]:
ballerina/ast.(*NodeBuilder).TransformServiceDeclaration(0x..., 0x...)
	ast/node_builder.go:1796 +0x748
ballerina/ast.(*NodeBuilder).TransformSyntaxNode(...)
	ast/node_builder.go:137 +0x2f5c
ballerina/ast.(*NodeBuilder).transformTopLevelInner(...)
	ast/node_builder.go:1618 +0x24
ballerina/ast.(*NodeBuilder).transformTopLevel(...)
	ast/node_builder.go:1607 +0x28
ballerina/ast.(*NodeBuilder).TransformModulePart(...)
	ast/node_builder.go:1430 +0x6f8
ballerina/ast.GetCompilationUnit(...)
	ast/ast.go:1543 +0x40
ballerina/projects.buildCompilationUnits(...)
	projects/module_context.go:445 +0x17c
ballerina/projects.resolveTypesAndSymbols(...)
	projects/module_context.go:255 +0x16c
ballerina/projects.(*PackageCompilation).compileModulesInternal(...)
	projects/package_compilation.go:128 +0x88c
```

### Proposal

Short term, replace the panic with a diagnostic carrying the annotation's position, so the user gets `annotation attachments on service declarations are not yet supported` pointing at the right line. Do the same for the four sibling panics.

Long term, supporting the attachment needs somewhere to key the value: `BLangService` embeds `classDefnBase` (`ast/ast.go:130-161`), which has no symbol field — only `BLangClassDefinition` adds one (`:149-153`). Giving `BLangService` a symbol, or desugaring service declarations into class definitions before annotation resolution, is the prerequisite.

### Impact

`@http:ServiceConfig`, `@http:ResourceConfig` and `@http:Cache` are unreachable, and with them everything they configure: `treatNilableAsOptional`, `consumes`/`produces` (the 415 path), CORS, and relaxed data binding. Since a service declaration is the idiomatic form, users hit a panic rather than a "not supported" message.

---

### Problem

`desugar/desugar.go:1147` `desugarFunctionParamDefaults` turns each defaultable parameter into a `$default$N` top-level function. It is invoked for `pkg.Functions` (`:1195`), init functions (`:1204`) and `class.Methods` (`:1209`) — but never for resource methods, neither `classDef.ResourceMethods` (`:1288`) nor `svc.ResourceMethods` (`:1305`). A defaultable parameter on a resource method is therefore accepted at declaration and then treated as required at every call site.

There is a second, latent half. `initLocalsForFunction` (`runtime/internal/exec/executor.go:83-86`) indexes `args[i+argOffset]` unguarded:

```go
requiredCount := len(birFunc.RequiredParams)
for i := range requiredCount {
    frame.SetLocal(i+1+argOffset, args[i+argOffset])
}
```

so any native caller that supplies fewer arguments than `RequiredParams` gets a Go index-out-of-range panic rather than a Ballerina error. Today the `ballerina/http` listener avoids it only by refusing resources with more than one non-path parameter (`lib/stdlibs/ballerina/http/0.0.1/go1.26/native/http_server.go:478-481`) — a guard that has to be removed for query/header/payload binding.

### Reproduction

`default-resource-param.bal`:
```ballerina
import ballerina/io;

client class Store {
    resource function get price(int id = 10) returns int {
        return id;
    }
}

public function main() {
    Store s = new;
    int v = s->/price.get();   // should use the default and print 10
    io:println(v);
}
```

```bash
go build -o bal ./cli/cmd
./bal run default-resource-param.bal
```

```
error[SEMANTIC_ERROR]: missing required parameter 'id'
  --> default-resource-param.bal:11:13
   |
11 |     int v = s->/price.get();
   |             ^^^^^^^^^^^^^^^
```

The same declaration on a regular method (`function price(int id = 10)`, called as `s.price()`) compiles and prints `10`.

### Proposal

Run `desugarFunctionParamDefaults` over `classDef.ResourceMethods` and `svc.ResourceMethods` in `desugar/desugar.go:1288` / `:1305`, so resource methods get the same `$default$N` treatment as regular methods. Independently, add a bounds check to `initLocalsForFunction` that raises a Ballerina error on a short argument list — cheap hardening that turns a process-level panic into a catchable error, and worth doing even if the desugar fix lands first.

### Impact

`resource function get price(int id = 10)` — an explicitly specified `ballerina/http` behaviour (spec §2.3.4.3: "If the query parameter is defined with a defaultable value in the resource signature, in the absence of particular query parameter, the default value will be assigned to the variable") — cannot be supported. The unguarded index is a latent crash on the HTTP listener goroutine.

gimantha:
Fixes #716

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for preserving and accessing annotations on classes, services, functions, methods, parameters, and rest parameters.
  * Externally implemented code can inspect callable signatures, parameter metadata, and object annotations at runtime.
  * Added runtime resolution for annotation values and references.
  * Added dependent default parameters for resource methods.

* **Bug Fixes**
  * Improved resource argument handling and annotation retention.
  * Standard-library object creation now preserves annotation metadata.

* **Tests**
  * Added comprehensive annotation, service, resource-method, and runtime metadata coverage.


semantics/type_resolver.go
TharmiganK: Should the constant annotation to be readonly as well?

The following fails:
```bal
import ballerina/io;

public type HttpHeader record {|
    string name?;
|};

public const annotation HttpHeader Header on parameter, record field;

function greet(@Header {name: "X-Api-Key"} string key) returns string {
    return key;
}

public function main() {
    io:println(greet("k"));
}
```

```console
error[SEMANTIC_ERROR]: const annotation type must be readonly
 --> test.bal:7:25
  |
7 | public const annotation HttpHeader Header on parameter, record field;
  |                         ^^^^^^^^^^
```

runtime/extern/handle.go
```diff
@@ -41,7 +59,9 @@ type DispatchHandles struct {
 	LookupRemote         func(*Context, *values.Object, string) (any, bool)
 	LookupResource       func(*Context, *values.Object, string, []values.BalValue) (any, bool) // resourceMethodName, path
 	LookupResourceByPath func(*Context, *values.Object, string, []string) (any, int, bool)     // accessor, segments
-	LookupFunction       func(*Context, string, string, string) (any, bool)                    // org, module, name
+	ResourceParams       func(*Context, any) (ResourceSignature, bool)
```
heshanpadmasiri: NIT: IMO these don't have anything to do with dispatch. I think they should be a different set of Handles

runtime/extern/handle.go
```diff
@@ -83,6 +103,19 @@ func (c *Context) LookupResourceMethodByPath(obj *values.Object, accessor string
 	return MethodHandle{impl: impl}, extraArgs, ok
 }
 
+// ResourceSignature returns the non-path parameter metadata associated with a
+// resolved resource handle. It returns false for non-resource handles or when
+// the backing BIR function is unavailable.
```
heshanpadmasiri: I am not entierly sure about the when backing BIR function is unavailable part. If someone have a MethodHandle I would assume method to exist

runtime/extern/handle.go
```diff
@@ -32,6 +35,21 @@ type FunctionHandle struct {
 	Fn any
 }
 
+// ParamDescriptor describes a non-path resource parameter for native
+// dispatchers that bind protocol values to a resource invocation.
+type ParamDescriptor struct {
+	Name        string
+	Type        semtypes.SemType
```
heshanpadmasiri: Do we actually need Name and Type as well, or is Annotation only (I am happy with Type not so much with Name). Either way I would like to keep Metadata seperate from Actualy type. Perhaps there should be one funcion that return "FunctionSignature" that return the param Name, Type + Return type. Another that takes the handle and return the Metadata (both structs will ideally have similar shapes).

runtime/extern/handle.go
heshanpadmasiri: Also even for extern functions/methods we do have signature so I persume we need to support all handles uniformly

test_util/testharness/test_harness.go
```diff
@@ -440,6 +454,24 @@ func Run(t testing.TB, tc test_util.TestCase, pal TestPal, externs []ExternRegis
 	if len(birPkgs) == 0 {
 		t.Fatalf("compilation succeeded but produced no BIR packages for %s", tc.Name)
 	}
+	if roundtripBIR {
```
heshanpadmasiri: Not entierly sure why we need special handling here. I would have assumed if serialization is working correctly by simply adding project tests this to just work

semantics/type_resolver.go
```diff
@@ -2241,27 +2252,36 @@ func resolveTypeData(t typeResolver, typeData *ast.TypeData) bool {
 	return true
 }
 
-type neverVisitor struct{}
+type neverVisitor struct {
```
heshanpadmasiri: It is not entierly clear to me why this change is needed. This code is there so that after setting the types of every node that has non never type we run this to fill them in. If annotations have type (i assume a mapping type) we should set it before filling the reset

semantics/symbol_resolver.go
```diff
@@ -1940,6 +1940,14 @@ func resolveServiceDefinition(ms *compilationUnitSymbolResolver, svc *ast.BLangS
 	allocateServiceResourceMethodSymbols(svcResolver, svc.ResourceMethods)
 
 	finishResolveClassDefinition(ms, svcResolver, svc.Fields, svc.Methods, svc.ResourceMethods, svc.InitFunction, nil, svcResolver.scope, serviceMethodSymbolName, resourceMethodsAreNetworkClass)
+
+	serviceSymbol := model.NewTypeSymbol("$service", false, svc.GetPosition())
```
heshanpadmasiri: NIT: I think you need a incrementing prefix (within the module scope) here. Otherwise we could have two type symbols with the same name if there are 2 services (probably will not cause an issue currently given there is no way to refer this symbol but still I would like to keep names seperate to avoid suprises)

runtime/internal/exec/handle.go
```diff
@@ -28,7 +28,8 @@ import (
 // InvokableHandle is provides a unified representation that can be used to execute any function/method
 // in runtime
 type InvokableHandle struct {
-	invoke func(ctx *extern.Context, args []values.BalValue) (values.BalValue, error)
+	invoke        func(ctx *extern.Context, args []values.BalValue) (values.BalValue, error)
+	resourceEntry *values.ResourceEntry
```
heshanpadmasiri: I would imagine the closure that is set as invoke to capture the any additional information it needs to dispatch to target instead

heshanpadmasiri: I am okay with there being additional closures for returning signatures and metadata

corpus/parser/subset9/09-annotation/27-v.json
warunalakshitha, line 1: We do not need this file in git

gimantha: Removed

runtime/extern/handle.go
gimantha: fixed

runtime/extern/handle.go
gimantha: Fixed

test_util/testharness/test_harness.go
gimantha: Fixed

semantics/type_resolver.go
gimantha: Fixed

semantics/symbol_resolver.go
gimantha: Fixed

runtime/internal/exec/handle.go
gimantha: Fixed

runtime/extern/handle.go
gimantha: Fixed

heshanpadmasiri (dismissed):
Frontend changes LGTM

runtime/internal/exec/handle.go
```diff
@@ -73,11 +83,101 @@ func parentFrameFromFunctionValue(fnValue *values.Function) *Frame {
...
+		signature.RestParam = &extern.Parameter{
+			Name: fn.RestParams.Name.Value(),
+			Type: fn.LocalVars[restLocalIndex].Type,
+		}
+	}
+	return signature, true
+}
+
+func describeFunctionMetadata(ctx *extern.Context, fn *bir.BIRFunction, firstParam int) (extern.FunctionMetadata, bool) {
+	if firstParam > len(fn.RequiredParams) || fn.ReturnVariable == nil {
```
warunalakshitha: we never use fn.ReturnVariable in this function so do we need to check null?

runtime/internal/exec/dispatch.go
```diff
@@ -50,7 +50,7 @@ func LookupFunction(env *extern.Env, org, module, name string) (any, bool) {
 		return NewBIRHandle(fn), true
 	}
 	if ef := reg.GetNativeFunction(key); ef != nil {
-		return NewNativeHandle(ef.Impl), true
+		return newNativeHandle(ef.Impl, reg.GetFunctionDescriptor(key)), true
```
warunalakshitha: reg.GetNativeFunction(key) is calling twice for same key

bir/bir_gen.go
```diff
@@ -1943,39 +2003,31 @@ func transformResourceMethodInner(root *funcBlock, rm *ast.BLangResourceMethod,
 			continue
 		}
 		root.addLocalVar(model.Name(name), ctx.CompilerContext.SymbolType(ref), ref)
-		requiredParams = append(requiredParams, BIRParameter{Name: model.Name(name)})
+		requiredParams = append(requiredParams, BIRParameter{
+			Name:        model.Name(name),
+			Annotations: values.NewAnnotationValues(),
```
warunalakshitha: should we add required annotations from required params like ctx.CompilerContext.SymbolAnnotationValues(param.Symbol()),

bir/bir_gen.go
```diff
@@ -448,6 +453,40 @@ func newFunctionRoot(ctx *Context, definedIn context) *funcBlock {
...
+	}
+	setFunctionLocals(root, birFunc)
+	birFunc.ErrorTable = root.fn.errorEntries
+	return birFunc
+}
+
+func transformNativeFunction(root *funcBlock, astFunc *ast.BLangFunction, selfSymbolRef *model.SymbolRef) *BIRFunction {
+	if _, ok := root.fn.birCx.CompilerContext.GetSymbol(astFunc.Symbol()).(model.DependentlyTypedFunctionSymbol); ok {
+		name := model.Name(astFunc.GetName().GetValue())
+		return &BIRFunction{
```
warunalakshitha: creating a function with LocalVars == nil, will that break pretty_print.go code

bir/pretty_print.go
```diff
@@ -96,15 +96,13 @@ func (p *PrettyPrinter) Print(tyCtx semtypes.Context, node BIRPackage) string {
 func (p *PrettyPrinter) PrintFunction(function BIRFunction) {
 	p.write(function.Name.Value())
 	p.write("(")
-	paramStart := 1
-	if function.Flags.Has(model.FlagAttached) {
-		paramStart = 2
-	}
+	paramStart := function.ParamLocalVarOffset()
 	for i, v := range function.LocalVars[paramStart:] {
```
warunalakshitha, line 104: there are cases where function.LocalVars == null. sholud we handle those or fix in birgen level

bir/bir_gen.go
```diff
@@ -1916,6 +1942,40 @@ func buildResourceMethodEntry(ctx *Context, rm *ast.BLangResourceMethod, fn *BIR
...
+
+func transformNativeResourceMethod(root *funcBlock, rm *ast.BLangResourceMethod, selfSymbolRef *model.SymbolRef) *BIRFunction {
+	if _, ok := root.fn.birCx.CompilerContext.GetSymbol(rm.Symbol()).(model.DependentlyTypedFunctionSymbol); ok {
+		name := model.Name(root.fn.birCx.CompilerContext.SymbolName(rm.Symbol()))
+		return &BIRFunction{
+			BIRNodeBase:       BIRNodeBase{Pos: root.fn.loc(rm.GetPosition())},
+			Name:              name,
+			OriginalName:      model.Name(rm.GetName().GetValue()),
+			Flags:             rm.Flags(),
+			FunctionLookupKey: buildFunctionLookupKeyFromSymbol(root.fn.birCx, rm.Symbol()),
```
warunalakshitha: same function.LocalVars == null

runtime/internal/exec/handle.go
```diff
@@ -73,11 +83,101 @@ func parentFrameFromFunctionValue(fnValue *values.Function) *Frame {
...
+	handle := &InvokableHandle{invoke: invoke}
+	if descriptor == nil {
+		return handle
+	}
+	handle.signature = func() (extern.FunctionSignature, bool) {
+		return describeFunctionSignature(descriptor, firstParam)
+	}
+	handle.metadata = func(ctx *extern.Context) (extern.FunctionMetadata, bool) {
+		return describeFunctionMetadata(ctx, descriptor, firstParam)
+	}
```
warunalakshitha: newInvokableHandle unconditionally builds signature/metadata closures whenever descriptor != nil.The two closures do nothing but capture descriptor/firstParam and forward to describeFunctionSignature/describeFunctionMetadata. Store the values directly instead of wrapping them in closures — same behavior, zero extra allocations:

warunalakshitha: ex // InvokableHandle struct
type InvokableHandle struct {
	invoke      func(ctx *extern.Context, args []values.BalValue) (values.BalValue, error)
	descriptor  *bir.BIRFunction
	firstParam  int
}

runtime/internal/exec/handle.go
warunalakshitha: newNativeHandle(builtin, fd)

bir/pretty_print.go
gimantha: Fixed in 267d26403. Confirmed it panics today — `bal run --dump-bir` on any module with a dependently typed `external` function dies with `slice bounds out of range [1:0]`. Fixed in the printer rather than bir_gen: those functions genuinely have no signature to materialize (`DependentlyTypedFunctionSymbol.TypedSignature()` panics with "must be Monomorphized"), so `LocalVars == nil` is legitimate and consumers have to tolerate it. Extracted the parameter list into `printFunctionParams` with an early return, and added a regression test.

bir/bir_gen.go
gimantha: Fixed in 267d26403 — yes, it broke pretty_print.go, see the reply there. Kept `LocalVars == nil` here since a dependently typed symbol has no materializable signature, and guarded the printer instead.

bir/bir_gen.go
gimantha: Fixed in 267d26403, same as `transformNativeFunction` — guarded in pretty_print.go.

bir/bir_gen.go
gimantha: Fixed in 0c2cbf0b3 — now reads `ctx.CompilerContext.SymbolAnnotationValues(ref)` like the required and rest parameters do. Heads up that it resolves to an empty set today: `BLangResourcePathSegment` has no annotation attachments and `TransformResourcePathParameter` never reads `resourcePathParameterNode.Annotations()`, so an annotation on a path parameter is silently dropped in the AST builder. I verified this with `resource function get value/[@parameterMeta {name: "segment"} string seg]()` — it compiles clean and the annotation never reaches BIR. So this change is forward compatible rather than a behavior change; wiring the AST side up looks like a separate issue, happy to file it if you agree.

runtime/internal/exec/dispatch.go
gimantha: Fixed in a7733f25a. Slight correction on the diagnosis: those are two different maps — `GetNativeFunction` reads `nativeFunctions` (Go impls registered by extern modules) and `GetFunctionDescriptor` reads `functionDescriptors` (BIR descriptors), and they are populated independently so they can't collapse into one access. What was genuinely duplicated is the whole resolve-then-build sequence, repeated verbatim in `LookupFunction` and `NewFunctionValueHandle`, so I extracted `nativeHandleFor(reg, lookupKey)` and both sites now go through it.

runtime/internal/exec/handle.go
gimantha: Fixed in 0fafa58a5 — exactly the struct you sketched. `InvokableHandle` now holds `descriptor` and `firstParam`, `newInvokableHandle` is a plain literal, and `FunctionSignature`/`FunctionMetadata` call `describeFunctionSignature`/`describeFunctionMetadata` directly, keying availability off `descriptor == nil` instead of the nil closures. Two fewer allocations per handle.

gimantha: Fixed in 0fafa58a5 — adopted this shape verbatim.

runtime/internal/exec/handle.go
gimantha: Fixed in 53d0708c5 — the runtime builtin branch now passes `reg.GetFunctionDescriptor(lookupKey)`, so builtins report signature and metadata like the extern branch below already did. `onGracefulStop` is declared in lang.runtime's .bal so the registry does hold a descriptor for it. That left `NewNativeHandle` with no callers, so I dropped it.

runtime/internal/exec/handle.go
gimantha: Fixed in 2aad99962. You're right that `describeFunctionMetadata` never reads `ReturnVariable`, but the check isn't dead — it's what rejects descriptors with no materialized signature (native dependently typed functions, which carry no locals and no return variable; same ones behind the pretty_print panic on this PR). Rather than delete it I named the intent: `hasMaterializedSignature(fn)`, used by both describe functions.

warunalakshitha (approved):
LGTM

PR #736, opened 2026-08-10
feat(avro): port ballerina/avro standard library

### LINKED ISSUE #733: Add basic support for avro package
> $Subject

TharmiganK:
## Summary

Fixes: https://github.com/ballerina-nutcracker/ballerina/issues/733

Ports the `ballerina/avro` standard library to the Go-native interpreter — Avro binary serialization and deserialization driven by an Avro schema string, covering every Avro type the module specification maps to a Ballerina type.

- `avro:Schema` class: `init(schema)`, `toAvro(data)`, `fromAvro(data, targetType?)`
- Full type coverage: `null`, `boolean`, `int`, `long`, `float`, `double`, `bytes`, `string`, `record`, `enum`, `array`, `map`, `union`, `fixed`
- Named-type references, namespaces, and recursive schemas
- `fromAvro` binds the decoded payload to the type inferred from the call site (or an explicit `targetType`) via `values.CloneWithType` — records, enums, tuples, singletons, numeric widening, `map<T>`/`T[]`/`json`/`anydata` targets all work

Schema parsing and the binary wire format are backed by `github.com/hamba/avro/v2` (MIT); the Ballerina↔Avro value mapping is implemented natively to preserve jBallerina parity, including its scalar coercion rules (`int` wraps to 32 bits, a `string` schema stringifies any value, `int`/`decimal` widen into `float`/`double`).

## Example

```ballerina
import ballerina/avro;
import ballerina/io;

type Person record {|
    string name;
    int age;
|};

    avro:Schema schema = check new (string `{
        "type": "record", "name": "Person", "namespace": "example.avro",
        "fields": [
            {"name": "name", "type": "string"},
            {"name": "age", "type": "int"}
        ]}`);

    byte[] serialized = check schema.toAvro({name: "Ann", age: 30});
    Person person = check schema.fromAvro(serialized);
    io:println(person.name); // Ann
}
```

## Notes

- Public contract validated against jBallerina's `module-ballerina-avro` (v1.2.2) — no breaking changes. `avro:Error` is a plain `error` alias rather than `distinct error` (interpreter limitation, already documented for other stdlibs).
- One deliberate divergence: union branch selection applies a single natural-type-first rule at every position. jBallerina's `MessageFactory` omits `UNION` from its top-level switch, so a top-level union falls through to Apache Avro's own branch resolution and silently rejects `string`/`bytes`/`record`/`enum`/`array`/`fixed` branches — even the common `["null","string"]` idiom. The Go-native version applies its record-field union rule everywhere instead, so those cases work; a `["double","long"]` union given an `int` selects the lossless `long` branch where a jBallerina record field would select `double`.
- Also returns `avro:Error` for an unparseable schema instead of panicking out of `init`, matching what the module's own specification (§2.1) says should happen.
- Everything else — including scalar coercion quirks like `int` wraparound and `string`-schema stringification — is reproduced exactly as jBallerina's `SerializeVisitor` implements it, even where the behaviour looks surprising, since those are deliberate lines in that code. A separate upstream issue is being raised to confirm intent on one fall-through case (`float` into an `int`/`long` schema, which isn't handled by any explicit case and reaches Apache Avro's `Number` cast by accident) and on the top-level union gap.
- See `lib/stdlibs/ballerina/avro/0.0.1/go1.26/README.md` for the full support matrix and behavioural-change notes.

## Test plan

- [x] go build ./...
- [x] go vet ./...
- [x] go test ./corpus/...
- [x] Corpus tests added under a new `library/subset3` milestone (documented in `doc/library/subset3.md`): avro-showcase-v, avro-primitives1-v, avro-record1-v, avro-enum1-v, avro-array1-v, avro-map1-v, avro-union1-v, avro-fixed1-v, avro-errors1-v, avro-target-types1-v, avro-infer-e
- [x] `native/` coverage measured at ~91% (target: 80%)
- [x] Scalar and union behaviour cross-checked against jBallerina 2201.13.4 via `bal run`

## New Features
- Added the `ballerina/avro` standard library for schema-based Avro serialization and deserialization.
- Supports primitives, records, arrays, maps, enums, unions, fixed values, nested structures, and recursive records.
- Added target-type conversion, numeric widening, schema validation, and structured error reporting.

## Documentation
- Added usage guidance, supported types, limitations, stream-based I/O, HTTP response binding, and behavioral notes.

## Tests
- Added comprehensive examples and integration coverage for successful conversions, type handling, and failure scenarios.


lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/avro.go
```diff
@@ -0,0 +1,191 @@
...
+			return schemaErr, nil
+		}
+		payload, _ := args[1].(*values.List)
+		target, _ := args[2].(*values.TypeDesc)
+		tc := ctx.TypeCtx()
+
+		native, _, err := schema.codec.NativeFromBinary(payload.ToByteSlice())
+		if err != nil {
+			return avroErrorFrom(deserializationError, err), nil
+		}
```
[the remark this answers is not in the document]
TharmiganK: Verified against jBallerina directly with bal run (both a primitive int schema and a record schema, each with garbage bytes appended after a valid encoding): jBallerina's own fromAvro silently ignores trailing bytes and returns the decoded value — identical to current behavior. Adding a trailing-bytes check would introduce a new divergence from jBallerina, not fix a bug. Per this project's parity rule, that's the wrong direction.

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
```diff
@@ -0,0 +1,322 @@
...
+
+// Schema-directed serialization of Ballerina values into goavro's native Go
+// representation, mirroring jBallerina's SerializeVisitor.
+//
+// encodeValue only ever builds the Go value goavro.Codec.BinaryFromNative
+// expects (map[string]any for record/map, []any for array, a wrapped
+// map[string]any{branchName: value} for a non-null union member, and a Go
+// type matching each primitive exactly); it never touches the wire format
+// itself. The one call to BinaryFromNative in toAvroExtern does that, and its
+// own type/membership/size checks are a backstop that should not normally
```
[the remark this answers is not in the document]
TharmiganK: invalid/stale. The module actually vendors github.com/linkedin/goavro/v2 (confirmed in lib/go.mod and the file's own imports), not hamba/avro/v2. The finding asks to revert accurate goavro references to a library that isn't used — applying it would make the comments wrong.

doc/library/subset3.md
```diff
@@ -1,6 +1,51 @@
...
+
+### Avro-to-Ballerina type mapping
+
+Every mapping the module specification defines is supported in both directions.
+
+| Avro type | Ballerina type | Notes |
+|---|---|---|
+| `null` | `()` | Encodes to zero bytes |
+| `boolean` | `boolean` | |
+| `int`, `long` | `int` | An `int` schema wraps to 32 bits, matching jBallerina; a `float` is rejected rather than truncated |
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [03feb20](https://github.com/ballerina-nutcracker/ballerina/pull/736/commits/03feb20d3f37942d28080806bbc89aaf1d945bcd)

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
```diff
@@ -0,0 +1,322 @@
...
+// applies write-side defaults either.
+func encodeRecord(s *shape, data values.BalValue) (any, error) {
+	record, ok := data.(*values.Map)
+	if !ok {
+		return nil, typeMismatch("record", data)
+	}
+	fields := make(map[string]any, len(s.fields))
+	for _, field := range s.fields {
+		value, _ := record.Get(field.name)
+		encoded, err := encodeValue(field.shape, value)
```
Nuvindu: Here, what if the `field.name` does not exist in the `record`. Will it captures the field as empty and correctly propagates the error?

Nuvindu:
Do we have test cases for deserializing Avro data into `record & readonly` types?

TharmiganK:
> Do we have test cases for deserializing Avro data into `record & readonly` types?

Added tests with [da818e7](https://github.com/ballerina-nutcracker/ballerina/pull/736/commits/da818e767d62a5c07c443d58cfce13d6a19341ea)

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
TharmiganK: Yes, the empty value is propagated to the encoders and the relevant encoder will return a proper error: Example:

```bal
import ballerina/avro;
import ballerina/io;

    avro:Schema needsBoth = check new (string `{
        "type": "record", "name": "Both", "fields": [
            {"name": "a", "type": "string"},
            {"name": "b", "type": "int"}]}`);
    byte[]|avro:Error incomplete = needsBoth.toAvro({a: "only"});
    if incomplete is avro:Error {
        io:println("message: ", incomplete.message());
        io:println("full: ", incomplete);
    }
}
```

**Nutcracker:**
```bal
error Error ("Avro serialization error", error("field 'b': expected a value matching the 'int' schema, found ''"))
```

**Swan Lake:**
```bal
error Error ("Avro serialization error", error("java.lang.NullPointerException: null value for (non-nullable) int at Both.b"))
```

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
```diff
@@ -0,0 +1,402 @@
...
+// panics as soon as the value is a list or a mapping.
+func stringOf(data values.BalValue) string {
+	return values.String(data, map[uintptr]bool{})
+}
+
+// encodeBytes also serves fixed: both take a byte[] as-is and leave size and
+// membership checks (fixed's exact length, enum's exact symbol) to goavro,
+// which already implements and tests them.
+func encodeBytes(data values.BalValue) (any, error) {
+	value, ok := data.(*values.List)
```
warunalakshitha, line 171: we are checking only list but not byte array even string array will become true and panic when calling  value.ToByteSlice()

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+// stringOf is values.String with the cycle-tracking map it needs: passing nil
+// panics as soon as the value is a list or a mapping.
+func stringOf(data values.BalValue) string {
+	return values.String(data, map[uintptr]bool{})
+}
+
+// encodeBytes also serves fixed: both take a byte[] as-is and leave size and
+// membership checks (fixed's exact length, enum's exact symbol) to goavro,
```
warunalakshitha: leave size and membership checks? Do we need this

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
warunalakshitha, line 171: should we use isByteList

corpus/bal/library/subset3/avro-errors1-v.bal
```diff
@@ -0,0 +1,134 @@
...
+    io:println(boolSchema.toAvro(1) is avro:Error); // @output true
+    avro:Schema longSchema = check new (string `{"type": "long"}`);
+    io:println(longSchema.toAvro("text") is avro:Error); // @output true
+    io:println(longSchema.toAvro(1.5d) is avro:Error); // @output true
+    avro:Schema doubleSchema = check new (string `{"type": "double"}`);
+    io:println(doubleSchema.toAvro(true) is avro:Error); // @output true
+    avro:Schema floatSchema = check new (string `{"type": "float"}`);
+    io:println(floatSchema.toAvro(1.5d) is avro:Error); // @output true
+    avro:Schema bytesSchema = check new (string `{"type": "bytes"}`);
+    io:println(bytesSchema.toAvro("text") is avro:Error); // @output true
```
warunalakshitha, line 81: Shall we add toAvro(["x"]) and toAvro([300])

corpus/bir/subset9/09-object/service-isolated-resource-remote-1-v.txt
warunalakshitha, line 1: why do we need this change?

corpus/bir/subset9/09-object/service-lock-field-1-v.txt
warunalakshitha, line 1: same

Nuvindu (dismissed):
Check against the current jBallerina implementation for unknown deviations. LGTM

corpus/bir/subset9/09-object/service-isolated-resource-remote-1-v.txt
TharmiganK: Adding avro to the builtin stdlib list bumps a counter the compiler uses for naming service lock keys, and these two files are the only ones that happen to bake that counter into their expected output. Nothing about their actual test content changed, they just needed regenerating.

corpus/bir/subset9/09-object/service-lock-field-1-v.txt
TharmiganK: Adding avro to the builtin stdlib list bumps a counter the compiler uses for naming service lock keys, and these two files are the only ones that happen to bake that counter into their expected output. Nothing about their actual test content changed, they just needed regenerating.

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
TharmiganK: Fixed the comment

corpus/bal/library/subset3/avro-errors1-v.bal
TharmiganK: Added the ["x"] one.

For [300], I checked what jBallerina actually does first — turns out it throws an NPE. In nutcracker, toAvro([300]) succeeds: 300 wraps to 44 (300 mod 256), same as we already do for int/long schemas when the value doesn't fit. So there's no working jBallerina behavior to match here, and I didn't want to guess at what's "correct" — even in-range values like [0, 255] hit the same NPE on their side, so it's not a real range check we can copy, just broken.

Adding a @Nuvindu to weigh in on whether wrapping is the right call here or if we should be doing something else.

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
TharmiganK: I actually tried that first, but it broke things. isByteList looks at the value's declared type, and something like schema.toAvro([9, 8, 7, 6]) — a plain literal, which is how most people will call this — doesn't come in as a "byte[]" type here, just a generic int array. That's why did not use isByteList. 

@Nuvindu can you check whether is it valid to checks each element is actually an int64 and support int[], or should be an error saying unsupported type? In jBallerina this example, returns a NPE

```bal
import ballerina/avro;
import ballerina/io;

    avro:Schema digestSchema = check new (string
        `{"type": "fixed", "name": "Md5", "namespace": "demo", "size": 4}`);

    // A perfectly normal call: a bare literal with valid byte values (0-255).
    byte[]|error result = digestSchema.toAvro([9, 8, 7, 6]);

    io:println(result is error); // expect false, actually true
    if result is error {
        error? cause = result.cause();
        if cause is error {
            io:println(cause.message()); // NullPointerException
        }
    }
}
```

TharmiganK: Had a chat with @Nuvindu and the jBallerina case is a bug. But we should not infer it as byte[] using the check. We should stick with whatever type inferred from runtime. Will fix this.

corpus/bal/library/subset3/avro-errors1-v.bal
TharmiganK: Had a chat with @Nuvindu and the jBallerina case is a bug. But we should not wrap the value since [300] is not byte[] from runtime. Will fix this

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
TharmiganK: This is fixed now

corpus/bal/library/subset3/avro-errors1-v.bal
TharmiganK: This is fixed now

lib/stdlibs/ballerina/avro/0.0.1/go1.26/native/serialize.go
TharmiganK: This is fixed now

lib/stdlibs/ballerina/avro/0.0.1/go1.26/README.md
```diff
@@ -0,0 +1,82 @@
...
+| String | Supported | |
+| Record | Supported | |
+| Enum | Supported | |
+| Array | Supported | |
+| Map | Supported | |
+| Union | Supported | Branch selection prefers the branch matching the value's natural Avro type — see Notable Behavioural Changes. |
+| Named type references and namespaces | Supported | |
+| Recursive schemas | Supported | |
+| Module error type | Partially Supported | `avro:Error` is declared, but as a plain `error` alias; the `distinct` type descriptor is not yet supported. |
+| Readonly and intersection target types | Supported | `readonly & T` binds and freezes the decoded value, for a record, an array, and a tuple. |
```
Nuvindu, line 73: We can add `Logical Types` as `Not supported` at the moment

lib/stdlibs/ballerina/avro/0.0.1/go1.26/avro.bal
```diff
@@ -0,0 +1,63 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents any error related to Ballerina Avro module
+// Note: distinct error types are not yet supported; Error is currently an alias for error.
+public type Error error;
```
Nuvindu, line 19: Is it okay to have the Error type and class in the same file. Shouldn't we need to have a separate file?

corpus/bal/library/subset3/avro-primitives1-v.bal
```diff
@@ -0,0 +1,96 @@
...
+import ballerina/avro;
+import ballerina/io;
+
+public function main() returns error? {
+    // null carries no bytes at all.
+    avro:Schema nullSchema = check new (string `{"type": "null"}`);
+    byte[] nullBytes = check nullSchema.toAvro(());
+    io:println(nullBytes.length()); // @output 0
+    () nothing = check nullSchema.fromAvro(nullBytes);
+    io:println(nothing is ()); // @output true
```
Nuvindu, line 26: `nothing is ()` always returns true here right? Since we used `check` in the above line.

lib/stdlibs/ballerina/avro/0.0.1/go1.26/README.md
TharmiganK: This is to track the parity between SwanLake and Nutcracker. Since this is not even supported in SwanLake, we do not need to track it here. The Avro package spec should document it

lib/stdlibs/ballerina/avro/0.0.1/go1.26/avro.bal
TharmiganK: We need a restructuring, previously there was a issue in nutcracker parser with multiple bal files. Now it is fixed, thinking of doing the restructure later

corpus/bal/library/subset3/avro-primitives1-v.bal
TharmiganK: Yes, we currently has corpus tests which tests the output by running these samples. I think it is ok for the corpus tests

Nuvindu (approved):
LGTM

PR #723, opened 2026-08-06
refactor(semtypes): clean up the machine-migrated interface

heshanpadmasiri:
## Summary
- clean up machine-migrated names in the semtypes API
- make unnecessary public functions private
- replace overloaded list and mapping helpers with functional options

Depends on https://github.com/ballerina-nutcracker/ballerina/pull/722

This is the stacked replacement for #714; the original PR remains open during the transition.

Closes #686

PR #738, opened 2026-08-10
feat: lax typing

### LINKED ISSUE #558: [New Feature]: Add support for LAX typing
### Description

We need to support https://ballerina.io/spec/lang/master/#lax_static_typing. I think this is purely a type resolution thing runtime should just work

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #668: [Bug]: Getting `unsupported container type for field access` error for `json` types
### Description

> $Subject

### Steps to Reproduce

Run this code:

```bal
import ballerina/io;

public function main() {
    json user = {
        name: "Tharmigan",
        age: 30
    };
    io:println(user.name);
    io:println(user.age);
}
```

Error:
```console
error[SEMANTIC_ERROR]: unsupported container type for field access
 --> test.bal:8:16
  |
8 |     io:println(user.name);
  |                ^^^^^^^^^

error[SEMANTIC_ERROR]: unsupported container type for field access
 --> test.bal:9:16
  |
9 |     io:println(user.age);
  |  
```

### Affected Version(s)

Ballerina Nutcracker - 0.6.0

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
$subject
Resolves #668 , #558
Depends on #723 

## Summary by CodeRabbit

* **New Features**
  * Added lax field access for JSON and mapping values, including nested and optional access.
  * Added clear handling for missing fields, nil values, invalid receivers, and propagated errors.
  * Added support for retrieving mapping values through the `map.get` operation.

* **Bug Fixes**
  * Ensured receivers are evaluated only once during lax field access.
  * Prevented lax access from being used as an assignment target or bypassing invalid field checks.

* **Tests**
  * Added comprehensive coverage for lax, optional, strict, nested, and invalid field-access scenarios.


semantics/internal/types/type_resolver.go
```diff
@@ -5762,6 +5762,53 @@ func resolveIndexBasedAccess(t typeResolver, chain *binding, expr *ast.BLangInde
 	return resultTy, defaultExpressionEffect(chain), true
 }
 
+func isLax(t typeResolver, ty semtypes.SemType) bool {
+	return isLaxInner(t, ty, semtypes.NewSemtypeInterner(), make(map[semtypes.InternHandle]bool))
+}
+
+func isLaxInner(t typeResolver, ty semtypes.SemType, interner *semtypes.SemTypeInterner, seen map[semtypes.InternHandle]bool) bool {
+	tyCtx := t.typeContext()
+	if semtypes.IsSameType(tyCtx, ty, semtypes.CreateJSON(tyCtx)) {
```
warunalakshitha, line 5771: json & readonly is not lax, so j.name is still a compile error

warunalakshitha: Shall we check following behaviour withjballerina

json & readonly j = {name: "a"}; j.name 

map<json> & readonly m = {name: "a"}; m.name

semantics/internal/types/type_resolver.go
```diff
@@ -5762,6 +5762,53 @@ func resolveIndexBasedAccess(t typeResolver, chain *binding, expr *ast.BLangInde
...
+	if !semtypes.IsSubtype(tyCtx, ty, semtypes.Mapping) {
+		return false
+	}
+	handle := interner.Intern(ty)
+	if seen[handle] {
+		return false
+	}
+	seen[handle] = true
+	defer delete(seen, handle)
+	return semtypes.AllMappingAtomMemberTypesMatch(tyCtx, ty, func(memberTy semtypes.SemType) bool {
```
warunalakshitha: Shall we check  following examples as well
```ballerina
type Person record {|
    json name;
    json...;
|};
public function main() {
    Person p = {name: "Ada", "extra": 1};
    json n = p.name;     // should be json
    json|error x = p.age; // should be a compile error
}
```

warunalakshitha: And this too
```ballerina
    json user = {name: "Tom", age: 30};
    string name = check user.name; 
    int age = check user.age;
}
```

semantics/internal/types/type_resolver.go
heshanpadmasiri: We don't treat json subtypes as lax, on the ground [spec](https://ballerina.io/spec/lang/master/#lax_static_typing) specifically states
> In this version of Ballerina, only the first step has been taken towards supporting this concept. There is a fixed set of type descriptors that are classified as lax: specifically json is lax, and map<T> is lax if T is lax. The only kinds of expression for which lax typing rules are specified are field-access-expr, optional-field-access-expr and checking-expr.

jBallerina treat json & readonly also as lax so "json & readonly j = {name: "a"}; j.name" works I think it's a bug given spec could have also stated `T&readonly` just like map case.

semantics/internal/types/type_resolver.go
heshanpadmasiri: I am not sure this is correct (jBallerina allows this)

> ```ballerina
> public function main() returns error? {
>     json user = {name: "Tom", age: 30};
>     string name = check user.name; 
>     int age = check user.age;
> }
> ```

Relevant section of the [spec](https://ballerina.io/spec/lang/master/#section_6.10) doesn't not specify a runtime type check for contextually expected type only allowing for the given field to be absent or container to be nil.

> A field-access-expr is evaluated as follows:
> 
> expression is evaluated resulting in a value v
> if v has basic type error, the result is v (this can only happen in the lax case)
> otherwise, if v does not have basic type mapping, the result is a new error value (this can only happen in the lax case)
> otherwise, if v does not have a member whose key is field-name, the result is a new error value (this can only happen in the lax case)
> otherwise, the result is the member of v whose key is field-name.

I can allow this if we think jBallerina implementation is correct and spec is underspecified (we already do this for new, check , checkpanic expressions)

heshanpadmasiri: Problem with allowing it is we need to specify how the type check is going to do is it just type test, is it a type cast (trigger numeric conversion) or something like cloneWithType (trigger numeric conversion for nested fields + readonly)

warunalakshitha: Lets keep this behaviour then. if this is a breaking change in current code if can check that separately. But still we can proceed with current impl since usage is low. if user require he can still use fromjsonwithtype with lax typing explicitly.

warunalakshitha (approved):
LGTM

PR #744, opened 2026-08-15
docs(architecture): update architecture diagram and related docs

warunalakshitha:
$Subject

## Summary by CodeRabbit

* **Documentation**
  * Updated terminology from “Platform Adaptation Layer” to “Platform Abstraction Layer.”
  * Clarified compilation stages, parallel processing, dependency ordering, and error handling.
  * Expanded architecture documentation covering interpretation, standalone builds, standard-library resolution, platform backends, and external system boundaries.
  * Added documentation for installing package archives in the local repository with `bal push`.
  * Improved the architecture diagram and descriptions of key components and workflows.


doc/guides/ARCHITECTURE.md
```diff
@@ -1,8 +1,8 @@
 # Architecture
 
-Ballerina Nutcracker compiles a `.bal` program to **Ballerina Intermediate Representation (BIR)** and then interprets that BIR. Almost everything below is a Go package inside the single static `bal` binary; only Ballerina Central and the host OS sit outside it.
+Ballerina Nutcracker compiles a `.bal` program to **Ballerina Intermediate Representation (BIR)** and then interprets that BIR (`bal run`), or embeds the BIR with the runtime into a standalone binary (`bal build`). Almost everything below is a Go package inside the single static `bal` binary; Ballerina Central, the local repository, the host OS, and the browser sit outside it.
```
heshanpadmasiri: Not entierly sure what is meant by a `static bal binary`, if it means a statically linked binary that is not true. dependencies like crypto and http, which we pack dynamically load system libararies if I am not mistaken

heshanpadmasiri (commented):
In the architecture diagram, it shows parser and ast pacakges to dependent on type system. Parser only indirectly depend on it (I belive it depends on decimal and that's the reason; could be wrong) st don't contain any semtypes, and AST only define fields that have types not perform type resolutions. Only other packages actually do type resolution. Perhaps it is worth it distinguish the two

heshanpadmasiri (commented):
Also I belive some front end passes (semantics and maybe desugar+bir gen) depends on values. 

Also it might be worth it seperate the parser and ast node gen from st and ast modules (this is how things are anyway in the code) That should make the dependencies clear. Also maybe worth it to indicate the context as well

heshanpadmasiri (commented):
Also I would move pal native out of the PAL box may be to HostOS (similar to browser having pal_wasm)

Perhaps image gen got confuced because PAL native is the same module (we should fix that)

heshanpadmasiri (approved):
LGTM

PR #750, opened 2026-08-17
feat: support defaultable fields when selecting mapping alternatives

### LINKED ISSUE #636: [Improvement]: Use mapping default expressions when selecting mapping alternatives
### Description

We need an equivalent of untyped function signature to record type declarations giving us the field name -> default closure mapping. We should then associate this with the type and should use it instead of current maps in type resolver to implement a proper lowering of mapping constructors. 

Similar to invocations after type resolution mapping constructors should have invocations to all their default closures

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

### LINKED ISSUE #766: frontend: partial mapping-constructor literal for a union-typed record field loses its default values
## Component

- [ ] Dependent stdlib: `ballerina/data.yaml`

## Summary

`analyzeMappingConstructorExpr` (`semantics/internal/analysis/semantic_analyzer.go:1477`) rejects a partially-specified nested record literal with "missing non-defaultable required record field", even though the field has a declared default, when all three of these hold:

1. the outer value being constructed is itself a mapping-constructor-expr for a record type (e.g. a record literal assigned to another record's field), and
2. the field's declared type is a union containing an anonymous (inline) record type, e.g. `record {...}|false`, and
3. the nested literal supplied for that field omits at least one of the anonymous record's defaultable fields.

Constructing the *same* anonymous-record-in-union type directly (as a standalone variable's type, not nested inside another record literal) works fine and fills in the omitted defaults correctly. So does supplying every field of the nested literal explicitly, or supplying `false` instead of a record. Only the specific combination above fails.

This was found while porting `ballerina/data.yaml`, whose `Options` record has exactly this shape:

```ballerina
public type Options record {|
    ...
    record {
        boolean nilAsOptionalField = false;
        boolean absentAsNilableType = false;
        boolean enableYamlStreamReorder = false;
    }|false allowDataProjection = {};
    ...
|};
```

A caller writing `yaml:Options opts = {allowDataProjection: {nilAsOptionalField: true}};` — a natural way to override one projection flag and take the other two defaults — fails to compile.

## Reproduction

Minimal repro, independent of `data.yaml`:

```ballerina
type Outer record {|
    record {
        boolean nilAsOptionalField = false;
        boolean absentAsNilableType = false;
    }|false allowDataProjection = {};
|};

public function main() {
    Outer o = {allowDataProjection: {nilAsOptionalField: true}};
    _ = o;
}
```

Run with: `go run ./cli/cmd run repro.bal`

```
error[SEMANTIC_ERROR]: missing non-defaultable required record field 'absentAsNilableType'
 --> repro.bal:9:37
  |
9 |     Outer o = {allowDataProjection: {nilAsOptionalField: true}};
  |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Verified on `main` @ `533244670c843207f1f2c333db95e36660ccef8a`.

### Confirmed NOT to fail (control cases)

```ballerina
// Same anonymous-record-in-union type, but not nested inside another record literal: fine.
public function main() {
    record {
        boolean nilAsOptionalField = false;
        boolean absentAsNilableType = false;
    }|false p = {nilAsOptionalField: true};
}
```

```ballerina
// Every field supplied explicitly: fine.
type Outer record {|
    record {
        boolean nilAsOptionalField = false;
        boolean absentAsNilableType = false;
    }|false allowDataProjection = {};
|};

public function main() {
    Outer o = {allowDataProjection: {nilAsOptionalField: true, absentAsNilableType: false}};
    _ = o;
}
```

```ballerina
// false instead of a record: fine.
type Outer record {|
    record {
        boolean nilAsOptionalField = false;
    }|false allowDataProjection = {};
|};

public function main() {
    Outer o = {allowDataProjection: false};
    _ = o;
}
```

## Expected vs actual

- **Expected:** compiles, with `absentAsNilableType` defaulted to `false`, matching jBallerina.
- **Actual:** `missing non-defaultable required record field 'absentAsNilableType'`.

### Why this looks like a bug, not a spec rule

Ballerina record fields with a declared default (`boolean x = false;`) may be omitted from any mapping-constructor literal of that record type; the compiler is responsible for injecting the default. The compiler already implements this correctly for the two control cases above.

### Suspected mechanism

`resolveMappingConstructorWithExpectedType` (`semantics/internal/types/type_resolver.go:4086`) populates `e.FieldDefaults` (consulted later by the `analyzeMappingConstructorExpr` check above) via one of two paths:

```go
// type_resolver.go:4110-4123
if ref, ok := t.getMappingAtomSymRef(mat); ok {
    if carrier, ok := t.getSymbol(ref).(model.MemberCarrier); ok {
        e.FieldDefaults = carrier.FieldDefaults()
    }
} else if bType, ok := t.getMappingAtomBType(mat); ok {
    // This happens for inline record type definitions given they don't have type symbol.
    if recTy, ok := bType.(*ast.BLangRecordType); ok {
        for name, field := range recTy.Fields() {
            if field.DefaultExpr != nil {
                e.FieldDefaults = append(e.FieldDefaults, model.FieldDefault{FieldName: name, FnRef: field.DefaultFnRef})
            }
        }
    }
}
```

For an anonymous (inline) record, only the second branch applies, and it depends on `t.getMappingAtomBType(mat)` finding an entry keyed by the exact `*semtypes.MappingAtomicType` pointer. That pointer is registered once, when the field's type descriptor (`record {...}|false`) is itself resolved (`type_resolver.go:7124-7125`, inside the `*ast.BLangRecordType` case of `resolveBType`):

```go
semType := d.Define(t.typeEnv(), fields, rest)
mat := semtypes.ToMappingAtomicType(t.typeContext(), semType)
t.setMappingAtomBType(mat, ty)
```

When the nested literal is later type-checked against the *decomposed* union member (extracting the record alternative out of `AnonRecord|false` to get `mat` for `selectMappingInherentType`), a fresh call to `semtypes.ToMappingAtomicType` for that decomposed member type appears not to return a pointer `==` to the one registered above, so `getMappingAtomBType` misses and `e.FieldDefaults` is left empty. When the same type is used directly (not nested, and not reached via union decomposition inside a field lookup), the pointer identity apparently survives, which matches why the two control cases succeed. This needs confirmation with a debugger/print-pointer trace — the above is inferred from behaviour, not verified against the atom-pool internals in `semtypes/`.

## Suggested fix

Either (a) make `getMappingAtomBType`'s association survive union decomposition (key by something stable derived from the `*ast.BLangRecordType` node itself, or re-derive/re-register the BType association whenever a mapping atomic type is extracted from a union member during field-type resolution), or (b) make `selectMappingInherentType` fall back to re-deriving field defaults from the AST type descriptor directly when the atom isn't found in either map, rather than silently treating every field as non-defaultable.

heshanpadmasiri:
$subject

Resolves #636
Closes #766

Depends on https://github.com/ballerina-nutcracker/ballerina/pull/737

## Summary by CodeRabbit

* **New Features**
  * Added support for preserving default values in inline, nested, union, and cross-module record types.
  * Added public type interning and mapping-type lookup capabilities.
  * Improved handling of omitted record fields when constructing mapping values.

* **Bug Fixes**
  * Mapping constructor diagnostics now identify incompatible or ambiguous field values more precisely.
  * Empty record literals can now resolve correctly when union alternatives provide default fields.

* **Tests**
  * Added coverage for record defaults across functions, narrowing, nested records, factories, and modules.


warunalakshitha (approved):
LGTM

PR #768, opened 2026-08-20
chore(lint): enable duplicate code detection

heshanpadmasiri:
Fixes #<Issue Number>

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logical expression evaluation while preserving correct short-circuit behavior.
  * Standardized validation for methods and type qualifiers.

* **Build & Quality**
  * Improved build, test, and lint execution across project modules.
  * Applied stricter, consistent code-quality checks throughout the codebase.

* **Documentation**
  * Generated source files now include clear notices indicating they should not be edited manually.


heshanpadmasiri:
Mechanical-refactor audit (excluding the later no-panic PR):

Reproduced this from base `ee36ba1e` with the pinned `golangci-lint` v2.10.1 built using Go 1.26.0. `dupl` reports 12 findings: 2 in `birgen`, 4 in `parser`, and 6 in generated `st` code. `golangci-lint --fix` makes no source edits.

The mechanical recipe is:

1. Select one duplicate with gopls `refactor.extract.function`.
2. Replace the second duplicate with the extracted helper.
3. Parameterize only the expressions that differ.

That produces the intended shapes: logical AND/OR through a shared `logicalExpression`; method definition/declaration validation through `validateMethodQualifiers`; and class/object qualifier validation through `createTypeQualNodeList`.

Files containing manual completion or broader edits that need review:

- `birgen/bir_gen.go`: gopls extracts only one body. Choosing the operator-based branch reversal and replacing the second function are manual.
- `parser/parser.go`: replacing the second copies, returning the modified `functionKeyword`, and introducing the qualifier predicate are manual.
- `compiler-tools/tree-gen/templates/st-node-util.go.tmpl` and `st/st_node_util_gen.go`: the generated duplication is suppressed with a generated-file marker rather than refactored. This intentionally removes the whole generated file from lint consideration.
- `Makefile` and `.github/workflows/golangci-lint.yml`: per-module targets and parallel lint execution are broader build-system changes.
- Commit `6ed83042`: Windows workspace-path normalization is a separate build fix, not part of the duplicate-code refactor.

The module config files/symlinks are lint activation plumbing; the files above are the parts that cannot be treated as an automatic `dupl` fix.

desugar/.golangci.yml
```diff
@@ -0,0 +1 @@
+../.golangci/strict.yml
```
warunalakshitha, line 1: lets add a newline all these files

heshanpadmasiri: It's a symlink not a source file (I don't know rational behind how github render these)

.golangci/strict.yml
```diff
@@ -0,0 +1,36 @@
...
+  enable:
+    - depguard
+    - dupl
+  settings:
+    depguard:
+      rules:
+        no-test-util-in-production:
+          files:
+            - "$all"
+            - "!$test"
```
warunalakshitha, line 14: I’m generally not in favor of skipping code from quality checks. Test code should be held to the same quality standards as production code. Otherwise, developers might misuse exclusion lists, similar to what happened with the jballerina Checkstyle exclusions.

heshanpadmasiri: This lints that there are no test utils in imports so by definition it needs to skip imports. Other lints are applied to tests as well

warunalakshitha (approved):
LGTM

Issue #777, opened 2026-08-21
resolver: module cannot import its own package's default/root module by name


TharmiganK:
## Component
- [x] Language / interpreter (compiler, BIR, runtime)

## Summary

A named module cannot `import` its own package's default (root) module by the package's bare name, when that package is consumed as a **dependency** (loaded via bala-style package resolution) rather than compiled as the top-level
project. The import is rejected as unresolved, even though the equivalent program is valid, cycle-free Ballerina and runs correctly on jBallerina. Discovered while porting `ballerina/protobuf`, whose `protobuf.types.any` submodule needs to reference a type declared in the package's default module (`ballerina/protobuf`).

Sibling-to-sibling submodule imports (e.g. `import org/pkg.other;` from within`org/pkg.thing`) already work correctly, including as a dependency — only importing the bare/default module of one's own package is broken.

Note: running the *provider* package directly as a project (`go run ./cli/cmd run <providerDir>`) does not reproduce this — if nothing in the provider's own default module imports the submodule, the submodule is never compiled at all (dead-code skip), so the bad self-import is never reached. The bug only surfaces once the package is loaded as a dependency of a separate consumer
project, which unconditionally compiles every module of the dependency.

## Reproduction

Zip of the reproducer: https://drive.google.com/file/d/17NE94OEkTS-STc_gSdV5sKge0PhL5XQH/view?usp=drive_link

Provider package `testorg/modimport:0.1.0`, installed into a local repository (`bal pack .` + `bal push --repository=local`, or the Go interpreter's equivalent bala-cache layout under `BAL_ENV`):

`Ballerina.toml`:
```toml
[package]
org = "testorg"
name = "modimport"
version = "0.1.0"
```

`main.bal` (default module — does **not** import `sub`, so there is no cycle):
```ballerina
public type Base record {|
    int x;
|};
```

`modules/sub/sub.bal`:
```ballerina
import testorg/modimport;

public function makeBase() returns modimport:Base {
    return {x: 1};
}
```

Consumer project depending on `modimport` via `repository = "local"`. Note it does **not** import `sub` either — only the default module:

`Ballerina.toml`:
```toml
[package]
org = "testorg"
name = "consumer"
version = "0.1.0"

[[dependency]]
org = "testorg"
name = "modimport"
version = "0.1.0"
repository = "local"
```

`main.bal`:
```ballerina
import ballerina/io;
import testorg/modimport;

public function main() {
[...]

PR #779, opened 2026-08-21
feat(stdlib): add `ballerina/protobuf` standard library support

### LINKED ISSUE #734: Add basic support for protobuf package
### Description

Port `ballerina/protobuf`. Depends on `time` (shipped). Already in flight as #779.

**Estimated effort:** Medium (2 native jars).

### Describe your problem(s)

`protobuf` is not implemented. `grpc` depends on it for the well-known type surface.

### Describe your solution(s)

- Well-known type wrappers: `Any`, `Duration`, `Timestamp`, `Struct`, `Empty`, and the scalar wrappers
- `protobuf:pack` and `protobuf:unpack` with typed targets
- Timestamp and duration conversion bridging to `time:Utc` and `time:Seconds`
- `@protobuf:Descriptor` annotation accepted; compile-time validation depends on the compiler-plugin platform (#453)

### Related area

Runtime

See the Relationships panel for blocking dependencies. Tracked by PR #779.

_No response_

_No response_

TharmiganK:
Related: https://github.com/ballerina-nutcracker/ballerina/issues/777

Fixes: https://github.com/ballerina-nutcracker/ballerina/issues/734

## Summary

Ports the `ballerina/protobuf` standard library to the Go-native interpreter — the protobuf well-known types (`Any`, `Struct`, `Timestamp`, `Duration`, `Empty`, and the scalar wrapper types) used by generated gRPC client/service code, structured as one real package with genuine sub-modules under `modules/` (`types.any`, `types.duration`, `types.empty`, `types.struct`, `types.timestamp`, `types.wrappers`), matching jBallerina's own package layout.

- `protobuf.types.any:pack`/`unpack` for every well-known `ValueType` member: `int`, `float`, `string`, `boolean`, `byte[]`, `()`, `time:Utc`, `time:Seconds`, `map<anydata>`
- `TypeMismatchError` raised when the requested target type doesn't match the packed value's type
- Full type declarations across all `protobuf.types.*` sub-modules (`Any`, `Empty`, `Struct`, `Timestamp`, `Duration`, and the scalar wrappers), plus their `Context*`/`Context*Stream` variants
- `protobuf:MessageDescriptor` and the `protobuf:Descriptor` annotation, declared for interface parity

Arbitrary user-defined message records (jBallerina's `@protobuf:Descriptor`-annotated `record {}` path) are **not yet supported**. This is a current interpreter gap rather than a permanent divergence — it will work once the interpreter catches up. The details:

- **`unpack` is already unblocked.** `unpack(anyValue, targetTypeOfAny)` receives a typedesc, and its `Annotations` already carry the attached `Descriptor` value — verified with a temporary native probe, which printed `ballerina/protobuf:0.0.1:Descriptor` alongside the `MessageDescriptor` value. At Ballerina level, `td.@protobuf:Descriptor` also resolves correctly today.
- **`pack` is the blocker.** `pack(ValueType message)` receives only the value. A record value here carries just its structural `semtypes.SemType` — no nominal identity, no annotation slot — so there is no route from the value back to the annotation on its declared type. jBallerina reaches it through the value's nominal inherent type (`RecordType.getAnnotations()`). The portable equivalent, `(typeof message).@protobuf:Descriptor`, works on jBallerina (verified with `bal run`) but panics here: `transformTypeofExpression unimplemented`.
- **So the missing `.proto`-to-Ballerina code generator is not the real blocker**, though it matters too: without it there is no realistic way to author or test a `@Descriptor`-annotated record.

Implementing only the `unpack` half would produce values that cannot round-trip, so descriptor support is deferred until both directions work. Until then an arbitrary `record {}` falls back to `google.protobuf.Struct`; the closed empty record `record {||}` needs no descriptor and packs as `google.protobuf.Empty`, matching jBallerina.

## Example

```ballerina
import ballerina/io;
import ballerina/protobuf.types.'any as pbany;

    pbany:Any packed = check pbany:pack(42);
    io:println(packed.typeUrl);   // type.googleapis.com/google.protobuf.Int64Value

    int unpacked = check pbany:unpack(packed);
    io:println(unpacked);         // 42
}
```

## Notes

- Public contract validated against jBallerina's `module-ballerina-protobuf` — no breaking changes. The one divergence: `protobuf.types.any:Error` is a standalone `distinct error` instead of `distinct protobuf:Error`, because a module can't currently import its own package's default/root module by name (filed as #777, with a jBallerina-verified reproduction); catching the root `protobuf:Error` will not also catch this module's errors. Like the descriptor gap above this is a temporary interpreter limitation, so both are tracked in the README support table rather than under "Notable Behavioural Changes", which the README format contract reserves for permanent architectural constraints.
- Wire encode/decode is backed by `google.golang.org/protobuf` (`wrapperspb`/`durationpb`/`timestamppb`/`structpb`/`emptypb`) in `native/any.go`, rather than a hand-rolled codec.
- Numeric widening (`Struct` ints → floats), nanosecond-precision `Duration`/`Timestamp` round-tripping, and the `TypeMismatchError` message format were all verified against real jBallerina.
- `test_util/testphases/phases.go`'s hand-rolled corpus-driver stdlib loader is generalized to support multi-file modules and multi-module packages (previously assumed exactly one `.bal` file per stdlib) — needed for real Ballerina sub-modules, not specific to protobuf.
- See `lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/README.md` for the full support matrix, and `doc/library/subset3.md` for the release-milestone summary.

## Test plan

- [x] `make build`
- [x] `make vet`
- [x] `go test ./...` across every workspace module
- [x] Corpus integration tests pass (`corpus/lib/subset3/protobuf-*.bal` — empty, wrappers, duration, timestamp, struct, any-scalars, any-errors, any-malformed, stream-fields)
- [x] `native/` coverage: 90.2%, exceeding the 80% target
- [x] `validate-stdlib-contract`: PASS (with the one documented divergence above)

## Summary by CodeRabbit

- **New Features**
  - Added Ballerina Protobuf standard-library support.
  - Added packing and unpacking for `Any`, including scalar wrappers, timestamps, durations, empty values, and `Struct` data.
  - Added context types for streaming and non-streaming protobuf values.
  - Added descriptor metadata and distinct error handling.
  - Closed empty records now map to `google.protobuf.Empty`.

- **Documentation**
  - Documented supported types, usage examples, limitations, and behavioral differences.

- **Tests**
  - Added coverage for conversions, malformed data, type mismatches, streams, and well-known types.


lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
```diff
@@ -0,0 +1,404 @@
...
+		return &emptypb.Empty{}, nil
+	case "BoolValue":
+		v, _ := message.(bool)
+		return wrapperspb.Bool(v), nil
+	case "Int64Value":
+		v, _ := message.(int64)
+		return wrapperspb.Int64(v), nil
+	case "FloatValue":
+		v, _ := message.(float64)
+		return wrapperspb.Float(float32(v)), nil
```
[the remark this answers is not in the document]
TharmiganK: invalid. I verified against real jBallerina earlier this session — it wire-encodes float as FloatValue (fixed32, confirmed via raw hex dump) with the expected float64→float32 truncation. Switching to DoubleValue would break parity with jBallerina, not fix a bug.

lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
```diff
@@ -0,0 +1,399 @@
...
+	hexVal, _ := anyMap.Get("value")
+	hexStr, _ := hexVal.(string)
+
+	raw, err := hex.DecodeString(hexStr)
+	if err != nil {
+		return newAnyError(ctx, "failed to decode Any value: "+err.Error())
+	}
+
+	suffix := strings.TrimPrefix(typeURL, typeURLPrefix)
+	decoded, naturalTy, err := wireMessageToValue(ctx, suffix, raw)
```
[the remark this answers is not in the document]
TharmiganK: Verified against real jBallerina (module-ballerina-protobuf's `AnyTypeCreator.unpack`) before acting, per policy for untrusted review data.

jBallerina dispatches on the **exact, full literal type-URL string** — `ProtobufConstants.WRAPPER_INT64_TYPE_URL = "type.googleapis.com/google.protobuf.Int64Value"`, etc. — looked up via an exact map (`Utils.isMatchingType` → `ModuleUtils.getAnyTypeMap().containsKey(typeUrl)`), not by stripping the domain and taking the last path segment. A `type.example.com/google.protobuf.Int64Value" value falls through jBallerina's own `default: break` in that switch too, and ultimately returns a `TypeMismatchError" (or the record/Struct fallback), exactly like this port's `strings.TrimPrefix` + `wireMessageToValue`'s `default` case does.

Generalizing to accept any URL prefix, as suggested, would make this port *more* lenient than jBallerina and diverge from the parity this PR is targeting — so I'm leaving the exact-prefix behavior as-is. No code or test changes made for this one.

daneshk:
Two things I found worth a look:

## 1. Possible precision loss in `decimalToSecNano` (native/any.go)

```go
func decimalToSecNano(sec *decimal.Decimal) (int64, int32) {
	intSec := int64(sec.Float64())        // <-- round-trips through float64
	intSecDec := decimal.FromInt64(intSec)
	fracDec, _ := sec.Sub(intSecDec)
	nanosDec, _ := fracDec.Mul(nanosPerSec)
	nanosInt, _, _ := nanosDec.Int64()
	return intSec, int32(nanosInt)
}
```

This backs `secondsToDuration`, used when packing a `time:Seconds` value. `float64` only carries ~15–17 significant decimal digits, but a realistic Unix-timestamp-scale `time:Seconds` value with full nanosecond precision (e.g. `1735689599.999999999`) has 19. I checked it numerically:

```
sec as float64          : 1735689600.0   (rounded up to the next whole second)
intSec via float64 path : 1735689600     (one second too many)
fracDec (sec - intSec)  : -1E-9          (fraction goes negative)
resulting Nanos field   : -1
```

So `pack()` can emit `Duration{Seconds: 1735689600, Nanos: -1}`. This still round-trips correctly through this module's own `secNanoToDecimal` (which just adds seconds + nanos back), but it violates the protobuf well-known-type spec, which requires `nanos` to have the same sign as `seconds`. A real gRPC/protobuf peer that calls `CheckValid()`/`AsDuration()` on the decoded message (a very common thing to do) could reject or misread it — and interop with real generated code is the whole point of this pack/unpack pair.

The existing corpus tests (`protobuf-duration-v.bal`) only use small values (`12.5`, `12.123456789`), so `float64` has more than enough precision there and this never triggers. Might be worth splitting the decimal into whole/fractional parts using decimal-only arithmetic (or via string formatting) instead of `Float64()`, and adding a corpus case with a large `time:Seconds` value with a fraction near `1` to cover it.

## 2. `record {||}` → `Empty` special case missing (any.bal)

jBallerina's `getUrlSuffixFromValue` has:

```ballerina
} else if anyMessage is record {||} {
    return "google.protobuf.Empty";
} else {
    return externGetNameFromRecord(<record {}>anyMessage);
}
```

The port's `typeUrlSuffix` doesn't have the `record {||}` branch, so an empty record value falls through to the final `else` and gets packed as `Struct` (with an empty field map) instead of `Empty`. The README's "Notable Behavioural Changes" documents the broader "arbitrary records pack as `Struct`" divergence, but that's specifically about `@protobuf:Descriptor`-annotated records needing schema resolution — it doesn't mention that a plain *empty* record, which needs no descriptor at all, also silently loses jBallerina's `Empty` fast path.

Minor in practice (`()` is the realistic way to express "no content" and is handled correctly), but since `record {}` is part of the documented `ValueType` union, might be worth either restoring the `record {||}` check or adding this to the documented divergences list.

Happy to be wrong on either of these — first real look at this codebase for me, so please sanity-check the reasoning.

TharmiganK:
> Two things I found worth a look:
> 
> ## 1. Possible precision loss in `decimalToSecNano` (native/any.go)
> ```go
> func decimalToSecNano(sec *decimal.Decimal) (int64, int32) {
> 	intSec := int64(sec.Float64())        // <-- round-trips through float64
> 	intSecDec := decimal.FromInt64(intSec)
> 	fracDec, _ := sec.Sub(intSecDec)
> 	nanosDec, _ := fracDec.Mul(nanosPerSec)
> 	nanosInt, _, _ := nanosDec.Int64()
> 	return intSec, int32(nanosInt)
> }
> ```
> 
> This backs `secondsToDuration`, used when packing a `time:Seconds` value. `float64` only carries ~15–17 significant decimal digits, but a realistic Unix-timestamp-scale `time:Seconds` value with full nanosecond precision (e.g. `1735689599.999999999`) has 19. I checked it numerically:
> 
> ```
> sec as float64          : 1735689600.0   (rounded up to the next whole second)
> intSec via float64 path : 1735689600     (one second too many)
> fracDec (sec - intSec)  : -1E-9          (fraction goes negative)
> resulting Nanos field   : -1
> ```
> 
> So `pack()` can emit `Duration{Seconds: 1735689600, Nanos: -1}`. This still round-trips correctly through this module's own `secNanoToDecimal` (which just adds seconds + nanos back), but it violates the protobuf well-known-type spec, which requires `nanos` to have the same sign as `seconds`. A real gRPC/protobuf peer that calls `CheckValid()`/`AsDuration()` on the decoded message (a very common thing to do) could reject or misread it — and interop with real generated code is the whole point of this pack/unpack pair.
> 
> The existing corpus tests (`protobuf-duration-v.bal`) only use small values (`12.5`, `12.123456789`), so `float64` has more than enough precision there and this never triggers. Might be worth splitting the decimal into whole/fractional parts using decimal-only arithmetic (or via string formatting) instead of `Float64()`, and adding a corpus case with a large `time:Seconds` value with a fraction near `1` to cover it.
> 
> ## 2. `record {||}` → `Empty` special case missing (any.bal)
> jBallerina's `getUrlSuffixFromValue` has:
> 
> ```ballerina
> } else if anyMessage is record {||} {
>     return "google.protobuf.Empty";
> } else {
>     return externGetNameFromRecord(<record {}>anyMessage);
> }
> ```
> 
> The port's `typeUrlSuffix` doesn't have the `record {||}` branch, so an empty record value falls through to the final `else` and gets packed as `Struct` (with an empty field map) instead of `Empty`. The README's "Notable Behavioural Changes" documents the broader "arbitrary records pack as `Struct`" divergence, but that's specifically about `@protobuf:Descriptor`-annotated records needing schema resolution — it doesn't mention that a plain _empty_ record, which needs no descriptor at all, also silently loses jBallerina's `Empty` fast path.
> 
> Minor in practice (`()` is the realistic way to express "no content" and is handled correctly), but since `record {}` is part of the documented `ValueType` union, might be worth either restoring the `record {||}` check or adding this to the documented divergences list.
> 
> Happy to be wrong on either of these — first real look at this codebase for me, so please sanity-check the reasoning.

Fixed via [b5d70ef](https://github.com/ballerina-nutcracker/ballerina/pull/779/commits/b5d70efd2270be4dcfd9039290394729cb27fa5c)

daneshk (approved):

lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
```diff
@@ -0,0 +1,425 @@
...
+// truncateTowardZero discards sec's fractional part, so the result never exceeds sec
+// in magnitude and keeps sec's sign.
+func truncateTowardZero(sec *decimal.Decimal) *decimal.Decimal {
+	var out *decimal.Decimal
+	if sec.Cmp(decimalZero) < 0 {
+		out, _ = sec.Ceiling()
+	} else {
+		out, _ = sec.Floor()
+	}
+	if out == nil {
```
warunalakshitha: can out be nil?

lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
```diff
@@ -0,0 +1,425 @@
...
+	case nil:
+		return nil, nil
+	case bool:
+		return val, nil
+	case int64:
+		return float64(val), nil
+	case float64:
+		return val, nil
+	case *decimal.Decimal:
+		return val.Float64(), nil
```
warunalakshitha, line 370: shall we add a test cover the decimal as well

lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
```diff
@@ -0,0 +1,425 @@
...
+// decimalToSecNano splits a Ballerina Seconds decimal into whole seconds and nanoseconds
+// using exact decimal arithmetic throughout. The whole part truncates toward zero, matching
+// jBallerina's BigDecimal.intValue(); routing it through float64 would round instead, and a
+// Unix-timestamp-scale value carrying full nanosecond precision needs 19 significant digits
+// where float64 offers ~15-17. Rounding there pushes the fraction past the second boundary
+// and flips its sign, which google.protobuf.Duration forbids: for durations of a second or
+// more, a non-zero nanos must carry the same sign as seconds.
+func decimalToSecNano(sec *decimal.Decimal) (int64, int32) {
+	if sec == nil {
+		return 0, 0
```
warunalakshitha: can sec be nil?

TharmiganK: Yes — this is a defensive guard, not dead code. `sec` comes from a type assertion at the call site (e.g. `dec, _ := message.(*decimal.Decimal)` in `valueToWireMessage`'s `Duration` case), which leaves `dec` nil if the assertion fails. `typeUrlSuffix` only dispatches to `Duration` when the value `is time:Seconds`, so this shouldn't trigger today, but the guard keeps the function safe instead of relying on that invariant forever holding.

lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
TharmiganK: Yes — `Decimal.Ceiling()`/`Floor()` return `(nil, *Error)` on an invalid/overflow condition (see `decimal/decimal.go`'s `validate`). For any realistic `time:Seconds` value (Unix-timestamp scale) this can't actually trigger — decimal128's representable range is far beyond that — but the guard avoids a nil-Decimal panic rather than depending on that never happening, falling back to zero seconds instead of crashing.

lib/stdlibs/ballerina/protobuf/0.0.1/go1.26/modules/types.any/native/any.go
TharmiganK: Good catch, added in 4c684bd1a. While verifying I found real jBallerina is actually broken for decimal values inside a Struct: a top-level decimal field silently becomes `null`, a decimal inside a nested map wipes out the *entire* nested value to `null`, and a decimal inside a list throws a `NullPointerException` on unpack (`BArray.size()` on a null array) — confirmed via `bal run` for all three shapes. Not worth replicating, so this keeps the port's existing choice: decimal widens to `float64` like `int` already does, since `google.protobuf.Struct` only carries doubles. Added flat/nested/list decimal cases to `protobuf-struct-v.bal`, verified round-tripping correctly through our own interpreter.

warunalakshitha (approved):
LGTM

PR #783, opened 2026-08-21
fix(annotations): expose record field annotations at runtime

### LINKED ISSUE #776: runtime: record-field-level annotations are not runtime-queryable
## Summary

Type-level annotation attachment already round-trips end to end: `annotation ... on type;` plus `@AnnotName {...} type T ...;` can be read back at runtime via `T.@AnnotName`. Record-field-level annotation attachment (`annotation ... on record field;` plus `@AnnotName {...}` directly on a field inside a `record {...}` type) parses and passes compile-time attach-point checking, but the resolved annotation value is never propagated anywhere a caller — Ballerina-level or native Go — can read it back. This blocks any stdlib (e.g. a port of `ballerina/constraint`) whose runtime behavior depends on inspecting per-field annotations of a record type.

## Reproduction

```ballerina
import ballerina/io;

annotation record {int min;} MinLen on type, record field;

@MinLen {min: 1}
type Name string;

type Person record {|
    @MinLen {min: 3}
    string name;
|};

public function main() {
    // Works today: type-level annotation is runtime-readable.
    record {int min;}? nameMeta = Name.@MinLen;
    io:println(nameMeta);

    // No equivalent exists for the `name` field's @MinLen inside Person.
    // Ballerina itself has no `.@` accessor for a field within a record
    // type (only whole-typedesc annotations are readable this way), and
    // there is no native-Go API either — the value is simply gone after
    // compile-time attach-point validation.
}
```

Running the type-level half through the interpreter confirms it works as expected:

```
$ go run ./cli/cmd run annot-repro.bal
{"min":1}
```

There is no way to extend this repro to demonstrate the field-level case succeeding, because no accessor exists at any layer — that absence is the defect. It is confirmed by direct inspection of the runtime representation and pipeline stages below, not by a failing run.

## Evidence

- `semtypes/field.go` — the `Field` struct that represents a record field's runtime shape carries only:
  ```go
  type Field struct {
      name     string
      typeOf   SemType
      readonly bool
      optional bool
  }
  ```
  There is no `annotations` field at all, so a record's `SemType` has nowhere to store per-field annotation values even if they were resolved.

- `semantics/internal/types/type_resolver.go` (`resolveAnnotationAttachments`, ~line 1196) — for a `record field` attach point, this only validates that the attach point and the annotation's declared type are compatible. The resolved annotation value is discarded after the check; it is never written onto the field's `SemType`.

- `birgen/bir_gen.go`, `bir/pretty_print.go`, `bir/codec/serializer.go` — none of these carry or serialize per-field annotation data. Type-level, class/service-level, object-field, and function/param/return annotations all flow into BIR and are runtime-visible; record-field annotations are the one attach point that is entirely absent from BIR.

- Ballerina-level access: there is no `.@` (or other) syntax to read a single field's annotation out of a record type descriptor — jBallerina itself only exposes whole-typedesc annotations via `T.@annotName`, so even a correct fix here would need a native-Go entry point (e.g. via `typedesc` introspection) rather than new surface syntax.

## Why this matters

A port of `ballerina/constraint` (jBallerina's validation library) is blocked by this. Its `validate()` function is driven almost entirely by walking a record's fields and reading each field's `@constraint:Int/Float/Number/String/Array/Date` annotation at runtime — this is not an edge case in that module, it is the primary and most common attachment point used throughout its spec and tests. Without a runtime-queryable per-field annotation table, there is no way to implement `validate()`'s core recursive-descent algorithm natively.

gimantha:
Fixes #776

Annotations attached to a record field were validated for their attach point and then thrown away, so nothing — Ballerina code or native Go — could read them back. This blocks a port of `ballerina/constraint`, whose `validate()` is driven almost entirely by reading per-field annotations at runtime.

The root cause is one gate in `resolveAnnotationAttachments`:

```go
storeValue := ownerSymbol != (model.SymbolRef{}) && sym.IsRuntimeVisibleAt(pointKey)
```

The record field call site passed the zero symbol, so `storeValue` was never true. That was not an oversight: record types are structural, so a field has no symbol of its own to key a value on.

Two notes correcting the issue description, since they affect how the fix is scoped:

- Record field is **not** the only attach point missing from BIR. `object field`, `return`, and function/method-level annotations are all in the same state — `bir.ObjectField` is `{Name, Ty}` and `extern.FunctionMetadata` is only `{Params, RestParam}`. What is runtime-visible today is type definitions, class/service objects, and parameters. This PR fixes record field only; the others are separate work.
- `model.AnnotationSymbol.IsRuntimeVisibleAt` already special-cases `"recordfield"` and `"objectfield"`, so the intent was there — only the plumbing was missing.

Key field annotations on the **enclosing type definition's symbol**, in a store parallel to the existing symbol annotation store, and carry them to the runtime on the typedesc constant the type already lowers to. That reuses the whole path type-level annotations already travel:

```
resolve → CompilerEnvironment store (type-def symbol + field name)
        → BIR gen looks the values up by the denoted type's symbol
        → values.TypeDesc.FieldAnnotations   (BIR ConstantLoad)
        → ctx.TypeAnnotations(td).Fields[field]
```

The AST carries no annotation data. `BLangTypedescExpr` no longer holds `AnnotationValues` either; BIR generation reads the compiler environment at lowering time, so it always observes the final state after the propagation pass described below.

Design decisions worth reviewing:

**A typed `FieldAnnotationValues map[string]AnnotationValues` rather than jBallerina's `$field$.name` key packing.** Packing into the existing flat annotation map would have been a smaller diff — symbol pool and BIR serialization would come for free — but a field name containing a `.` via a quoted identifier makes decoding ambiguous, and it pushes string surgery onto every Go consumer.

**An `annotationSink` closure replaces the `ownerSymbol` parameter of `resolveAnnotationAttachments`.** That function is ~90 lines handling repeated annotations, non-constant values, const checks and marker annotations. Redirecting where values are stored, rather than duplicating that logic for record fields, is why repeated and runtime-valued annotations on record fields work without any extra code. Call sites that only validate now pass `nil` instead of the zero symbol.

**Type inclusion inherits field annotations — a deliberate divergence from jBallerina.** jBallerina drops annotations at the inclusion boundary. Here `type Derived record {| *Base; |}` sees the annotations declared on `Base`'s fields, transitively and across module boundaries. Without this, validation driven by field annotations would silently skip every field of `record {| *BaseReq; ... |}`, which is a common shape.

Mechanics: the inclusion graph only exists in the AST (`RecordSymbol.Members()` is a flattened field list with no back-pointer to the included type), so the pass reads `BLangRecordType.Inclusions`, which the symbol resolver has already resolved to symbol refs. It runs as a **second pass** after every type definition has stored its own annotations, so the result does not depend on whether `Base` appears above or below `Derived` in the file. The same pass covers record aliases (`type Alias Base;`, transitively), and a field the record redeclares first drops the inherited annotations for that field before its own are applied, so `record {| *Base; string id; |}` reports nothing for `id` unless the redeclaration annotates it. Cross-module inclusions need no recursion: module A merges into `Base`'s stored entry when A is compiled, and the symbol pool carries the merged result to B.

Two behaviours the compiler decides, not this PR:

- A field reachable through more than one inclusion must be overridden (`included field 'tag' declared in multiple type inclusions must be overridden`), so the local declaration always resolves ambiguity and "later inclusion wins" is unreachable.
- Cyclic inclusion and alias chains are rejected upstream; the `visiting` guard in the merge is a fail-fast internal error for a violated invariant, not a recovery path.

Serialization was extended on both paths that carry annotations: the BIR codec (typedesc constants) and the symbol pool (record symbols, which also cover record aliases because the symbol resolver allocates an alias with its target's symbol kind). This branch bumped the pool version to 9; main bumped it independently to 10 in the meantime, and after merging main the branch sits at 10 with the new trailing block. The type-level annotation encoding is byte-identical to before — `writeLength` is `write(buf, int64(n))` — so only a new trailing block was added.

## Checklist

**Please confirm the inclusion-inheritance semantics.** It is the one intentional behaviour difference from jBallerina and the hardest thing to reverse later, since stdlibs would come to depend on it. It was requested explicitly, but it deserves a second opinion from whoever owns the `constraint` port.

Record field annotations are unreachable from Ballerina surface syntax, which shapes the testing: a `corpus/bal` test cannot observe them, so the behavioural tests live in `corpus/extern` with native Go readers. It also means the existing BIR round-trip corpus test — which compares program output — could never catch a codec regression here, hence the dedicated round-trip tests below.

Annotations on the fields of an **inline** record type (`record {| @A int x; |}` in a variable declaration) are still neither stored nor validated; `resolveTopLevelAnnotationAttachments` only walks `pkg.TypeDefinitions`. That gap predates this PR and is left alone.

New public API, in `runtime/extern`:

```go
type TypeAnnotations struct {
	Annotations values.AnnotationValues            // on the type itself
	Fields      map[string]values.AnnotationValues // per record field; only annotated fields have an entry
}

func (c *Context) TypeAnnotations(td *values.TypeDesc) (TypeAnnotations, bool)
```

The bool is false only if a runtime-valued annotation could not be loaded. Every value that did load is still returned, so one broken value does not hide the rest; `ObjectAnnotations` follows the same rule.

## Automation tests
- Unit tests
  > None kept. The `values/annotations_test.go` and `context/annotations_test.go` from an earlier revision were dropped on review; the `corpus/extern` tests below drive the same code end to end through `.bal` fixtures, which is the only way to observe record field annotations since they are unreachable from Ballerina surface syntax.
- Integration tests
  > Four tests in `corpus/extern` plus one in `corpus`:
  >
  > - `TestRecordFieldAnnotations` — multiple annotations on one field, optional fields, marker (`true`) annotations, unannotated fields, unknown field names, non-record typedescs, non-constant annotation values (the `$annotation$N` global dereference path), and source-only annotations correctly staying invisible at runtime.
  > - `TestRecordFieldAnnotationsStructural` — inclusion, transitive inclusion (`Root ← Middle ← Leaf`), a diamond with the required override, a local declaration overriding an inherited annotation, a redeclared field shadowing an inherited annotation, record aliases (`Alias`, `AliasOfAlias`), repeated (list-valued) annotations, and records with rest fields.
  > - `TestRecordFieldAnnotationsCrossModule` — a record type imported from another module, a local record that includes an imported one, and an imported record alias used as a typedesc, included into a local record, and aliased again locally.
  > - `TestRecordFieldAnnotationsSymbolPoolRoundtrip` — the same project, alias cases included, pushed through the symbol pool codec into a fresh environment and recompiled, which is the path a prebuilt `.bala` dependency takes.
  > - `TestRecordFieldAnnotationsSurviveBIRRoundtrip` — compiles a fixture, marshals and unmarshals the BIR, and compares the typedesc constants' field annotations.
  >
  > Both serialization tests were verified non-vacuous by breaking each codec in turn: the BIR one fails with `typedesc constant count changed: before 1, after 0`, and a deserializer that keeps every key but replaces each value with `999` fails the value digest comparison.

## Security checks

go 1.26.0, macOS (darwin/arm64). `make vet` and the full workspace test suite pass. `TestBalBuildNativeDependency` and `TestNativeRunner_ColdBuildAndCacheHit` fail locally on an unrelated pre-existing module-cache size limit when the test downloads the published module; they fail identically on unmodified `main`.

## Summary by CodeRabbit

* **New Features**
  * Added runtime support for reading annotations applied to record fields.
  * Preserved field annotations through type inclusions, aliases, compilation, and serialization round trips.
  * Unified type-level and record-field annotation access through a single metadata API.
  * Added support for repeated and computed record-field annotations.

* **Bug Fixes**
  * Corrected annotation propagation for inherited, overridden, aliased, and nested record fields.


ast/expressions.go
```diff
@@ -416,6 +416,9 @@ type (
 		// constant.
 		Constraint       semtypes.SemType
 		AnnotationValues values.AnnotationValues
+		// FieldAnnotationValues holds the annotations attached to the fields of
+		// the record type this typedesc denotes, keyed by field name.
+		FieldAnnotationValues values.FieldAnnotationValues
```
heshanpadmasiri: I am starting to feel this is the wrong place to store annotations given you need to also store these in compiler env. I think the correct approach is a single index (say `AnnotationHandle`). You store all the annotation information in the compiler env under that handle. Type symbols and expressions associate with that information by storing the index.
We have similar patterns for supporting things like defaultable params/args.

context/annotations_test.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
```
heshanpadmasiri: Can't we turn these into corpus tests

runtime/extern/handle.go
```diff
@@ -80,6 +80,12 @@ type MetadataHandles struct {
 	Signature         func(*Context, any) (FunctionSignature, bool)
 	Metadata          func(*Context, any) (FunctionMetadata, bool)
 	ObjectAnnotations func(*Context, *values.Object) (values.AnnotationValues, bool)
+	// RecordFieldAnnotations resolves the annotations attached to one field of
+	// a record type descriptor.
+	RecordFieldAnnotations func(*Context, *values.TypeDesc, string) (values.AnnotationValues, bool)
+	// AnnotatedRecordFields lists the annotated field names of a record type
+	// descriptor.
+	AnnotatedRecordFields func(*Context, *values.TypeDesc) []string
```
heshanpadmasiri: I am not sure if this is the common access pattern. I would imagine extern functions either need all the annotations attached or all the field annotations. Not just annotations for a perticular field. So I would prefer there to be either single function that return a struct with a map of annotations for each field + type annotations.

values/annotations.go
```diff
@@ -39,3 +41,32 @@ func (r *RuntimeAnnotationValueRef) GlobalLookupKey() string {
...
+}
+
+// SortedFields returns the annotated field names in a deterministic order.
+func (f FieldAnnotationValues) SortedFields() []string {
+	fields := make([]string, 0, len(f))
+	for field := range f {
+		fields = append(fields, field)
+	}
+	sort.Strings(fields)
+	return fields
```
heshanpadmasiri: Do we need to support this? record fields are anyway unordered according to spec AFAIK.

values/annotations_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
```
heshanpadmasiri: Can't we turn this to corpus test as well

semantics/internal/types/type_resolver.go
```diff
@@ -1302,24 +1304,117 @@ func resolveInvokableAnnotationAttachments(
...
+//
+// This runs as a second pass so that it does not depend on the order type
+// definitions appear in the source. Inclusions of types from other modules need
+// no recursion: those annotations were already merged when that module was
+// compiled, and reach us through the symbol pool.
+func inheritIncludedRecordFieldAnnotations(t typeResolver, pkg *ast.BLangPackage) {
+	inclusions := make(map[model.SymbolRef][]model.SymbolRef)
+	for i := range pkg.TypeDefinitions {
+		defn := pkg.TypeDefinitions[i]
+		record, ok := defn.GetTypeData().TypeDescriptor.(*ast.BLangRecordType)
```
heshanpadmasiri: This only propagates metadata for definitions whose descriptor is directly a `BLangRecordType`. A record alias has its own symbol and is skipped, so its field annotations disappear at runtime. For example:

```ballerina
type Base record {| @fieldMeta {name: "base-id"} string id; |};
type Alias Base;
io:println(fieldMetaName(Alias, "id")); // expected base-id, gets <absent>
```

Please copy or transitively resolve field annotations for record aliases as well.

semantics/internal/types/type_resolver.go
```diff
@@ -1302,24 +1304,117 @@ func resolveInvokableAnnotationAttachments(
...
+	visiting[symbol] = true
+	defer delete(visiting, symbol)
+	for _, included := range inclusions[symbol] {
+		for field, annotations := range includedRecordFieldAnnotations(t, included, inclusions, visiting) {
+			for key, value := range annotations {
+				merged.Set(field, key, value)
+			}
+		}
+	}
+	for field, annotations := range t.compilerContext().RecordFieldAnnotationValues(symbol) {
```
heshanpadmasiri, line 1381: Overlaying only the local annotation map cannot distinguish an inherited field from a locally redeclared field with no annotations. For example:

```ballerina
type Base record {| @fieldMeta {name: "base-id"} string id; |};
type Derived record {| *Base; string id; |};
io:println(fieldMetaName(Derived, "id")); // expected <absent>, gets base-id
```

Directly declared fields should first suppress inherited metadata for the same field, then apply their own annotations.

semantics/internal/types/type_resolver.go
```diff
@@ -1302,24 +1304,117 @@ func resolveInvokableAnnotationAttachments(
...
+// ones declared on its own fields. The visiting set keeps a cyclic inclusion
+// chain from recursing forever.
+func includedRecordFieldAnnotations(
+	t typeResolver,
+	symbol model.SymbolRef,
+	inclusions map[model.SymbolRef][]model.SymbolRef,
+	visiting map[model.SymbolRef]bool,
+) values.FieldAnnotationValues {
+	merged := values.NewFieldAnnotationValues()
+	if visiting[symbol] {
```
heshanpadmasiri, line 1361: Cyclic record inclusion is rejected by an earlier compiler stage, so returning an empty map here silently masks a violated post-resolution invariant and can compile incomplete runtime metadata. Please fail fast with an internal error, or remove this defensive recovery if the invariant is guaranteed.

heshanpadmasiri (changes requested):

semantics/internal/types/type_resolver.go
gimantha: Fixed in 3b2936467. Confirmed the repro: the pass only looked at definitions whose descriptor is a `BLangRecordType`, so `type Alias Base;` was skipped entirely. It now treats a `BLangUserDefinedType` descriptor as a source, so an alias picks up the aliased type's field annotations, and because the merge is recursive that also covers chains (`type AliasOfAlias Alias;`). Both are in the fixture now and return `base-id`.

semantics/internal/types/type_resolver.go
gimantha: Fixed in 3b2936467. You're right, and the old code got it wrong in exactly the way you describe. The pass now collects the field names each record declares directly (from `FieldPtrs()`, which excludes inclusion-inherited fields) and drops the inherited annotations for those fields before applying the record's own. So `type Derived record {| *Base; string id; |}` reports `<absent>` for `id`, while a redeclaration that does carry an annotation still wins. Both cases are covered — `Shadowing` for the suppress case and `Overriding` for the replace case.

semantics/internal/types/type_resolver.go
gimantha: Fixed in 3b2936467 — it now calls `t.internalError("cyclic type reference while propagating record field annotations", ...)`. I kept the guard rather than removing it because without it a violated invariant is a stack overflow rather than a diagnostic; failing fast was the first option you offered. For the record, I did verify the invariant holds today: `type A record {| *B; |}; type B record {| *A; |};` is rejected before this pass runs.

runtime/extern/handle.go
gimantha: Agreed and changed in 3b2936467. There is now a single accessor:

```go
type TypeAnnotations struct {
	Annotations values.AnnotationValues            // on the type itself
	Fields      map[string]values.AnnotationValues // per record field
}

func (c *Context) TypeAnnotations(td *values.TypeDesc) (TypeAnnotations, bool)
```

`RecordFieldAnnotations` and `AnnotatedRecordFields` are gone, and `MetadataHandles` carries one hook instead of two. It resolves every runtime annotation ref up front, so the bool now means only "a runtime value failed to load" rather than doubling as "this field has no annotations", which the per-field version conflated.

values/annotations.go
gimantha: Removed in 3b2936467. You're right that record fields are unordered, and with the API returning a map there was nothing left for it to order. The two serializers still need a deterministic byte stream, so they sort field names inline, next to where they already sort annotation keys. The tests that print field lists sort locally.

ast/expressions.go
gimantha: Addressed in 3b2936467, though by removing the AST storage rather than adding a handle. `BLangTypedescExpr` now carries neither `FieldAnnotationValues` nor the pre-existing `AnnotationValues`; BIR generation looks both up from the compiler environment by the denoted type's symbol. That is exactly what the sibling lowering site for a plain type reference (`bir_gen.go`, the `SymbolKindType` branch) already did, so the two paths now share one helper. The environment is the single source of truth and the AST holds no annotation data at all.

That also drops `annotationValuesForTypeDescriptor` from the resolver, and it is arguably more correct: the lookup now happens at lowering time, so it always observes the final merged state including the inclusion/alias propagation pass.

I did not introduce `AnnotationHandle` itself. With the AST fields gone the remaining association is just the symbol ref, and the indirection would only pay off as part of unifying the existing `symbolAnnotations` store — which backs class, service, parameter and object annotations and is serialized per symbol kind in the symbol pool. Happy to do that as a follow-up if you want the two maps in the environment collapsed under one handle, but it seemed wrong to fold that refactor into this fix. Let me know if you'd rather it happen here.

context/annotations_test.go
gimantha: Deleted in 3b2936467, along with `values/annotations_test.go`. The four `corpus/extern` tests already drive these paths end to end through .bal fixtures, so the unit tests were duplicating them at a lower level.

Worth flagging what is lost, since it is not zero: those tests covered the nil-map guards in the constructors and asserted that field annotations cannot leak into the plain symbol annotation store that backs type-level annotations. The nil-map paths are still hit through the fixtures (a non-record typedesc such as `int`), but the isolation property is now only implied rather than asserted directly.

The reason there is no plain `corpus/bal` test for any of this is that record field annotations are unreachable from Ballerina surface syntax — that absence is the bug being fixed — so the assertions have to come from a native reader, which is what `corpus/extern` provides.

values/annotations_test.go
gimantha: Deleted in 3b2936467 — same as the context one, see the reply there.

corpus/record_field_annotation_codec_test.go
```diff
@@ -0,0 +1,116 @@
...
+				td, ok := load.Value.(*values.TypeDesc)
+				if !ok || len(td.FieldAnnotations) == 0 {
+					continue
+				}
+				entry := make(map[string]string, len(td.FieldAnnotations))
+				for _, field := range td.FieldAnnotations.SortedFields() {
+					keys := slices.Sorted(maps.Keys(td.FieldAnnotations[field]))
+					entry[field] = joinKeys(keys)
+				}
+				result = append(result, entry)
```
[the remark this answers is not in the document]
gimantha: Good catch — fixed in 3b2936467. The assertion compared only annotation keys, so a codec that preserved every key while corrupting the values would have passed.

It now renders each typedesc constant as sorted `field key=value` lines with values rendered structurally (mappings field by field, lists element by element, runtime refs by their global lookup key) and compares those. The fixture gained a repeated annotation and a non-constant one, so the digest looks like:

```
age   ...:marker=true
dynamic ...:fieldMeta=ref($anon/record-field-annotations:$annotation$0)
name  ...:fieldMeta={name:person-name}
tags  ...:repeatable=[{name:one},{name:two}]
```

I verified it is load-bearing by making the deserializer keep every key but replace each value with `999`: the old assertion passed, the new one fails and prints the diff above.

corpus/extern/extern_test.go
```diff
@@ -732,6 +945,153 @@ func TestDependentlyTypedCrossModuleRoundtrip(t *testing.T) {
...
+	fieldMetaKey := model.AnnotationKey(model.PackageIdentifier{
+		Organization: org,
+		Package:      typesMod,
+		Version:      "0.1.0",
+	}, "fieldMeta")
+
+	ballerinaEnvFs := os.DirFS(getBallerinaEnvPath(t))
+	result, err := projects.Load(os.DirFS(absProjectDir), ".", projects.ProjectLoadConfig{
+		BallerinaEnvFs: ballerinaEnvFs,
+	})
```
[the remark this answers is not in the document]
gimantha: Not changing this one. `projects.Load` takes an `fs.FS`, which is the abstraction at this boundary, and `os.DirFS` is how every test in the repo supplies it — 67 call sites, including `runCompilePhase` in `corpus/integration_test.go` and `TestDependentlyTypedCrossModuleRoundtrip` directly above this test, which this one is modelled on. Converting only this call site would make it the odd one out without changing what is exercised.

The PAL rule in AGENTS.md covers platform interaction in product code (io, http, fs at runtime); loading fixture files from disk in a test is not that. If the intent is that tests should route fixture loading through PAL too, that is a repo-wide change and a separate PR.

heshanpadmasiri (dismissed):
Frontend changes generally LGTM let's rebase with main

corpus/record_field_annotation_codec_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/ballerina-nutcracker/ballerina/semtypes"
+	"github.com/ballerina-nutcracker/ballerina/values"
+)
+
+// TestRecordFieldAnnotationsSurviveBIRRoundtrip checks that the per-field
+// annotations baked into a typedesc constant are preserved by BIR
+// serialization, values included. Record field annotations are only reachable
+// from native code, so the corpus output-comparison roundtrip cannot observe
+// them.
+func TestRecordFieldAnnotationsSurviveBIRRoundtrip(t *testing.T) {
```
warunalakshitha: Can we move them to bal corpus?

semantics/internal/types/type_resolver.go
```diff
@@ -1311,24 +1313,155 @@ func resolveInvokableAnnotationAttachments(
...
+		t.compilerContext().SetSymbolAnnotationValue(symbol, key, value)
+	}
+}
+
+// recordFieldAnnotationSink stores annotation values against the field named
+// field of the record type defined by typeDefinition. Record types are
+// structural, so the field has no symbol of its own to key on.
+func recordFieldAnnotationSink(t typeResolver, typeDefinition model.SymbolRef, field string) annotationSink {
+	if typeDefinition == (model.SymbolRef{}) {
+		return nil
```
warunalakshitha: can we cover this?

model/symbolpool/serializer.go
warunalakshitha: writeTypeSymbol does not write field annotations. Exported type Alias Person loses them in a .bala round trip.

model/symbolpool/serializer.go
```diff
@@ -349,6 +360,9 @@ func (sw *symbolWriter) writeRecordSymbol(buf *bytes.Buffer, sym *model.RecordSy
 	if err := sw.writeAnnotationValues(buf, annotations); err != nil {
 		return err
 	}
+	if err := sw.writeFieldAnnotationValues(buf, fieldAnnotations); err != nil {
```
warunalakshitha, line 415: `writeFieldAnnotationValues` is only used for `*model.RecordSymbol`. A record alias (`type Alias Person`) is a `TypeSymbol`, so `writeTypeSymbol` / `readTypeSymbol` never persist the field map.

The second pass already copies field annotations onto the alias in the compiler env, which is why same-module `TypeAnnotations(Alias)` works. After a `.bala` / symbol-pool round trip those values are gone:

```ballerina
typedesc td = lib:Alias;                 // empty Fields
type Local record {| *lib:Alias; |};  // does not inherit Person's field annotations
```

`lib:Person` is fine. If cross-module record aliases should keep field annotations (re-exports, `*SomeAlias`), write and read the same field-annotation block on the `TypeSymbol` path. Skip if that is out of scope for this PR.

runtime/internal/exec/annotations.go
```diff
@@ -47,3 +47,23 @@ func resolveAnnotationValues(ctx *extern.Context, annotations values.AnnotationV
...
+// not be loaded.
+func TypeAnnotations(ctx *extern.Context, td *values.TypeDesc) (extern.TypeAnnotations, bool) {
+	annotations, ok := resolveAnnotationValues(ctx, td.Annotations)
+	if !ok {
+		return extern.TypeAnnotations{}, false
+	}
+	fields := make(map[string]values.AnnotationValues, len(td.FieldAnnotations))
+	for field, fieldAnnotations := range td.FieldAnnotations {
+		resolved, ok := resolveAnnotationValues(ctx, fieldAnnotations)
+		if !ok {
```
warunalakshitha, line 66: If `resolveAnnotationValues` fails for one field (a runtime-valued `$annotation$N` global that cannot be loaded), this returns `false` for the whole type and hides every other field. Consider still returning the fields that resolved.

values/values.go
warunalakshitha: shall we remove "a thin wrapper
// around a semtype." now has multiple fields

model/symbolpool/serializer.go
gimantha: Checked this and the premise does not hold, so there is no serializer change. `type Alias Person` is allocated as a `*model.RecordSymbol`, not a `TypeSymbol`: `allocateTypeSymbol` mirrors the target's symbol kind for a `BLangUserDefinedType` descriptor (`symbol_resolver.go`, the `case *model.RecordSymbol:` branch), and that holds for imported targets too because the kind comes from the deserialized symbol. A record alias therefore already goes through `writeRecordSymbol` and carries its field map.

To cover it rather than argue it, d279c2d9c adds `public type PersonAlias Person;` to the dependency module of the cross-module fixture and, in the consumer, a `typedesc` use of `types:PersonAlias`, a `record {| *types:PersonAlias; |}`, and `type LocalAlias types:PersonAlias;`. All three report `imported-name` in `TestRecordFieldAnnotationsSymbolPoolRoundtrip`, which pushes the dependency through `symbolpool.Marshal`/`Unmarshal` into a fresh environment before compiling the consumer. `TestRecordFieldAnnotationsCrossModule` runs the same fixture on the in-process path.

The only way a `TypeSymbol` ends up with field annotations is an alias whose target failed to resolve, which is a compile error anyway, so I left the `TypeSymbol` path alone rather than add a block nothing writes.

runtime/internal/exec/annotations.go
gimantha: Agreed, changed in d279c2d9c. `resolveAnnotationValues` now skips the entry it cannot dereference and keeps going, so `TypeAnnotations` returns every type-level and field-level value that did load, and the bool only says whether something was left out. A field whose only annotation failed to load gets no entry, so `Fields` still lists just the fields that carry annotations, as documented. `ObjectAnnotations` shares the helper and picks up the same behaviour. The doc comments on `Context.TypeAnnotations` and the exec implementation now spell out the contract.

values/values.go
gimantha: Done in d279c2d9c.

warunalakshitha (approved):
LGTM

PR #818, opened 2026-08-24
chore: disable panics in frontend

heshanpadmasiri:
Fixes #<Issue Number>

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Compiler and parser failures now produce diagnostics with accurate source locations instead of abruptly stopping.
  * Build, run, and package commands now halt safely when code generation reports errors.
  * Lock statements now correctly report missing restricted-variable access.
  * Diagnostic updates preserve location tracking when file contents change.

* **Quality Improvements**
  * Improved error recovery reports multiple compilation issues together.
  * Expanded linting rules help identify unsafe error handling and restricted imports.

### DIFF
--- a/.golangci/no-panic.yml
+++ b/.golangci/no-panic.yml
@@ -1,0 +1,43 @@
+version: "2"
+
+linters:
+  default: standard
+  enable:
+    - depguard
+    - dupl
+    - iface
+    - unparam
+    - forbidigo
+  settings:
+    forbidigo:
+      forbid:
+        - pattern: ^panic$
+          msg: return or report an internal error instead of panicking
+    depguard:
+      rules:
+        no-test-util-in-production:
+          files:
+            - "$all"
+            - "!$test"
+            - "!**/test_util/**"
+          deny:
+            - pkg: "github.com/ballerina-nutcracker/ballerina/test_util"
+              desc: "test_util is test-only; import only from *_test.go"
+            - pkg: "github.com/ballerina-nutcracker/ballerina/test_util/testphases"
+              desc: "test_util/testphases is test-only; import only from *_test.go"
+    staticcheck:
+      checks:
+        # Default
+        - all
+        - -ST1000
+        - -ST1003
+        - -ST1016
+        - -ST1020
+        - -ST1021
+        - -ST1022
+
+  disable:
+    - unused
+issues:
+  max-issues-per-linter: 0
+  max-same-issues: 0
--- a/parser/tokenizer.go
+++ b/parser/tokenizer.go
@@ -17,29 +17,48 @@
 package parser
 
 import (
+	compilercontext "github.com/ballerina-nutcracker/ballerina/context"
 	"github.com/ballerina-nutcracker/ballerina/st"
+	"github.com/ballerina-nutcracker/ballerina/tools/diagnostics"
 )
+
+func failedToken() st.STToken {
+	return st.CreateToken(st.EOF_TOKEN, st.CreateEmptyNodeList(), st.CreateEmptyNodeList())
+}
 
 type tokenReader struct {
 	lexer             tokenLexer
 	currentToken      st.STToken
 	tokenBuffer       tokenBuffer
 	currentTokenIndex int
+	ctx               *compilercontext.CompilerContext
+	fileName          string
+	location          diagnostics.Location
 }
 
-func createTokenReader(lexer tokenLexer) *tokenReader {
+func createTokenReader(ctx *compilercontext.CompilerContext, fileName string, lexer tokenLexer) *tokenReader {
+	location := diagnostics.NewLocation(ctx.DiagnosticEnv(), fileName, 0, 0)
 	return &tokenReader{
 		lexer:             lexer,
 		currentToken:      nil,
 		currentTokenIndex: 0,
+		ctx:               ctx,
+		fileName:          fileName,
+		location:          location,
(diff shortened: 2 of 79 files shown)


parser/tokenizer.go
```diff
@@ -2,46 +2,64 @@
 //
 // WSO2 LLC. licenses this file to you under the Apache License,
 // Version 2.0 (the "License"); you may not use this file except
-// in compliance with the License.
-// You may obtain a copy of the License at
+// in compliance with the License. You may obtain a copy of the License at
 //
-// http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
```
warunalakshitha: is this change intensional?

semantics/internal/common/deferred_method.go
```diff
@@ -36,21 +36,37 @@ func NewDeferredMethodSymbol(name string, space *model.SymbolSpace) *DeferredMet
 
 func (d *DeferredMethodSymbol) MethodName() string              { return d.name }
 func (d *DeferredMethodSymbol) SymbolSpace() *model.SymbolSpace { return d.space }
-func (d *DeferredMethodSymbol) Name() string                    { panic("method symbol has not been resolved yet") }
+
+//nolint:forbidigo // Intentional guard: deferred method symbols must be resolved before access.
+func (d *DeferredMethodSymbol) Name() string { panic("method symbol has not been resolved yet") }
+
+//nolint:forbidigo // Intentional guard: deferred method symbols must be resolved before access.
```
warunalakshitha, line 46: why this?

heshanpadmasiri: For internal transient symbols to avoid panics and give internal errors we need to change the symbol interface in model to pass in compiler context. This means model will depend on compiler context, which I don't like

parser/tokenizer.go
heshanpadmasiri: Fixed

.golangci/no-panic.yml
```diff
@@ -0,0 +1,43 @@
+version: "2"
```
warunalakshitha, line 1: strict.yml and this looks same. can we have one config? any reason have two files?

warunalakshitha:
Should we handle all panic cases? Some might never be happen? Code coverge still low but that is expected because most of runtime panics are abnormal behaviors.

warunalakshitha:
There are few more places might panic as per my local PR checkup

| Issue | File:line | Solution |
|---|---|---|
| Force type-assert after `transformSyntaxNode` can still panic once an internal error returns `nil` | `nodebuilder/node_builder.go:1380` | Use comma-ok assert; on failure report/return early instead of `.(T)` |
| Same force-assert on function body | `nodebuilder/node_builder.go:1439` | Same: comma-ok + early return / bad node |
| Same force-assert on else body | `nodebuilder/node_builder.go:2075` | Same: comma-ok + early return / bad stmt |
| Same force-assert on resource method body | `nodebuilder/node_builder.go:5352` | Same: comma-ok + early return / bad body |
| Parser/lexer internal errors always locate at file start (offset 0) | `parser/parser.go:126`, `parser/lexer.go:69`, `parser/tokenizer.go:60` | Update `location` from the current token/offset before calling `InternalError` |
| Token buffer overflow reports error then continues without storing, so lexer/buffer can diverge | `parser/tokenizer.go:136` | On overflow, stop filling and fail the parse (don’t keep calling `NextToken` / continue) |

.golangci/no-panic.yml
heshanpadmasiri: In addition to strict (I couldn't find a way to "inherit" or extend configs unfortunately so duplicated) it restricts explicit panics

heshanpadmasiri:
> Should we handle all panic cases? Some might never be happen? Code coverge still low but that is expected because most of runtime panics are abnormal behaviors.

Violations of invariants should be represented by internal errors, unimplemented features should go to unimplemented errors. Also if you use unimplemented errors you can write -fe or -fv tests already in the test harness, and harness ensure they fail with unimplemented errors already

heshanpadmasiri:
> There are few more places might panic as per my local PR checkup
> 
> Issue	File:line	Solution
> Force type-assert after `transformSyntaxNode` can still panic once an internal error returns `nil`	`nodebuilder/node_builder.go:1380`	Use comma-ok assert; on failure report/return early instead of `.(T)`
> Same force-assert on function body	`nodebuilder/node_builder.go:1439`	Same: comma-ok + early return / bad node
> Same force-assert on else body	`nodebuilder/node_builder.go:2075`	Same: comma-ok + early return / bad stmt
> Same force-assert on resource method body	`nodebuilder/node_builder.go:5352`	Same: comma-ok + early return / bad body
> Parser/lexer internal errors always locate at file start (offset 0)	`parser/parser.go:126`, `parser/lexer.go:69`, `parser/tokenizer.go:60`	Update `location` from the current token/offset before calling `InternalError`
> Token buffer overflow reports error then continues without storing, so lexer/buffer can diverge	`parser/tokenizer.go:136`	On overflow, stop filling and fail the parse (don’t keep calling `NextToken` / continue)

We are trying to prevent agents using panics to check for program invariants. Not necessarily triggering panics. I don't think any of these modules actually call panic function in any of these cases. I think if code should validate these as invariants is orthogonal to this

warunalakshitha (approved):

Issue #759, opened 2026-08-18
[Task]: Lower statement-producing expressions in BIR generation


heshanpadmasiri:
## Description

Expression desugaring currently represents some expressions as `initStmts` plus a replacement expression. This loses the expression's evaluation boundary and forces containing constructs to reposition generated statements manually.

This becomes incorrect or requires special handling when such expressions occur inside traps, lazily evaluated expressions, match guards, or other control-flow-sensitive expressions.

For example:

```ballerina
match x {
    1 if [...guardValues()].length() > 0 => {
        // ...
    }
    _ => {
    }
}
```

The spread-list constructor generates variable declarations and a loop. Those statements must run only after pattern `1` matches. Hoisting them before the match changes observable behavior.

### Current directly affected expressions

- Nil-lifted binary expressions: `+`, `-`, `*`, `/`, `%`, shifts, and bitwise operators.
- Nil-lifted unary expressions: `+`, `-`, and `~`.
- Optional field-access expressions.
- List constructors containing spread members.
- Query expressions.
- Function calls, method calls, resource accesses, and `new` expressions with omitted defaultable arguments.
- Type conversions and type tests with inline function-type descriptors that require default-value initialization.

Containing expressions transitively propagate these statements, including binary and Elvis expressions, grouping, indexed and field access, invocations, constructors, `check`, `checkpanic`, annotation access, templates, XML expressions, mappings, and error constructors.

Current problems include:

- `trap` emits an internal error if its operand produces initialization statements.
- Arrow-function bodies currently discard produced initialization statements.
- Lazy operands can have their statements evaluated eagerly.
- Match guards would require a desugar-only AST side channel to preserve conditional evaluation.

## Describe your task(s)

- Lower these expression forms directly during BIR generation at their actual evaluation point.
- Preserve short-circuiting, trap regions, conditional branches, and evaluation order.
- Remove expression-generated `initStmts` from desugaring once all affected forms are migrated.
- Add corpus coverage for statement-producing expressions inside traps, lazy operands, conditional expressions, and match guards.
- Promote the match-guard spread-list future test once BIR lowering supports it.

`check` and `checkpanic` have already been partially moved in this direction by [PR #746](https://github.com/ballerina-nutcracker/ballerina/pull/746).

## Related area

Compilation

## Related issue(s)

- #163
- PR #746

heshanpadmasiri:
[...]

PR #820, opened 2026-08-24
fix(bir): lower nil-lifted expressions as control flow

### LINKED ISSUE #760: [Bug]: Nil-lifted expressions fail inside trap expressions
## Description

Nil-lifted expressions cannot currently be used as the operand of a `trap` expression.

Nil-lifting desugaring produces temporary-variable declarations and conditional statements through `initStmts`. When `walkTrapExpr` receives those statements, it reports an internal compiler error because moving them outside the trap region would change the expression's semantics:

```text
fatal[INTERNAL_ERROR]: Init statements will be hoisted outside of trap region
```

This currently affects nil-lifted binary operators and nil-lifted unary operators.

## Steps to Reproduce

```ballerina
import ballerina/io;

public function main() {
    int? left = 1;
    int? right = 2;

    int?|error result = trap left + right;
    io:println(result);
}
```

Run:

```shell
go run ./cli/cmd run nillift-trap.bal
```

Observed result:

```text
fatal[INTERNAL_ERROR]: Init statements will be hoisted outside of trap region

ballerina: compilation contains errors
```

Expected result:

```text
3
```

The expression and all statements generated while lowering it must be evaluated within the trap region. They must not be hoisted ahead of `trap`.

## Affected Version(s)

Current `main` at `533244670c843207f1f2c333db95e36660ccef8a`.

## Related area

Compilation

## Related issue(s)

Sub-issue of #759.

heshanpadmasiri:
## Summary

- move nil-lifted unary and binary expression lowering from desugar to BIR generation
- preserve operand evaluation order and keep generated control flow inside trap regions
- add corpus coverage for traps, loop conditions, match guards, operand mutation, and evaluation order

## Testing

- `go test ./birgen/...`
- `go test ./corpus/...`

This is the root PR in the stack for #759.

Fixes #760

## Summary by CodeRabbit

* **New Features**
  * Nullable unary and binary expressions now support nil-lifting.
  * Nil-lifted operations produce `()` when applicable operands are nil while preserving operand evaluation.
  * Improved nullable-expression behavior in conditions, match guards, mutations, and trapped expressions.
  * Trapped expressions now correctly report errors such as divide-by-zero.

* **Tests**
  * Added coverage for nil-lifting, side effects, control flow, match guards, and error trapping.

### DIFF
--- a/birgen/bir_gen.go
+++ b/birgen/bir_gen.go
@@ -74,6 +74,11 @@ func (fn *functionContext) addBB() *bir.BIRBasicBlock {
 	return &bb
 }
 
+// lastBBNumber returns the number assigned to the most recently allocated basic block.
+func (fn *functionContext) lastBBNumber() int {
+	return len(fn.bbs) - 1
+}
+
 func (fn *functionContext) loc(pos diagnostics.Location) bir.Location {
 	return birLoc(fn.pkgCtx.CompilerContext.DiagnosticEnv(), pos)
 }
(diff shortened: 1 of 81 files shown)


birgen/bir_gen.go
```diff
@@ -1540,23 +1545,67 @@ func wildcardBindingPattern(ctx context, curBB *bir.BIRBasicBlock) expressionEff
...
+}
+
+func unaryExpression(ctx context, bb *bir.BIRBasicBlock, expr *ast.BLangUnaryExpr) expressionEffect {
+	if isNilLiftableUnaryOp(expr.Operator) {
+		ty := expr.Expr.GetDeterminedType()
+		if semtypes.ContainsBasicType(ty, semtypes.Nil) {
+			return nilLiftedUnaryExpression(ctx, bb, expr)
+		}
+	}
+	if expr.Operator == model.OperatorKind_ADD {
```
warunalakshitha: shall we move this to top so we can delete duplicate same check in nilLiftedUnaryExpression

warunalakshitha: may be we do not need this since it is hadled before birgen https://github.com/ballerina-nutcracker/ballerina/blob/58cbd97b8198aa30d8dc1eeac06d1571e6c1a704/desugar/expression.go#L223

heshanpadmasiri: Fixed

warunalakshitha (approved):
LGTM

PR #822, opened 2026-08-24
fix(bir): preserve check flow in expression thunks

### LINKED ISSUE #816: [Bug]: Support check expressions inside desugared expression closures
## Description

Some expressions are desugared into immediately invoked closures to preserve lazy evaluation, loop, and trap semantics. When such an expression contains a `check`, the generated closure creates a control-flow boundary that does not exist in the source program.

A `check` expression must discriminate an error result value and, when it is an error, jump to the error-return control flow of the parent source function. Returning from the generated closure instead would not preserve the source function's semantics.

The compiler currently rejects this case with an internal error:

```text
fatal[INTERNAL_ERROR]: check expression cannot be lowered inside a generated expression thunk
```

## Steps to Reproduce

```ballerina
import ballerina/io;

function checkedInt() returns int|error {
    return 10;
}

function add(int value, int increment = 1) returns int|error {
    return value + increment;
}

    (int|error)[] values = [];
    values = [add(check checkedInt())];
    io:println(values);
}
```

Run:

```shell
go run ./cli/cmd run example.bal
```

### Expected output

```text
[11]
```

The generated closure should propagate the `check` failure into the parent function's error-return control flow.

### Actual output

```text
fatal[INTERNAL_ERROR]: check expression cannot be lowered inside a generated expression thunk
```

## Related area

Compilation

heshanpadmasiri:
## Summary

- lower generated expression thunk bodies in the enclosing BIR function
- preserve parent error-return control flow for `check` expressions
- promote the expression-thunk check corpus case and update affected BIR goldens

## Testing

- `go test ./desugar/... ./birgen/... ./bir/...`
- `go test ./corpus/...`

Depends on https://github.com/ballerina-nutcracker/ballerina/pull/821.

Closes #816

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed expression-thunk handling for checked expressions.
  - Errors from checked expressions now propagate correctly instead of causing compilation failures.
  - Preserved package-level function declaration order in generated output.
  - Improved stack traces by including relevant expression-thunk frames.

- **Tests**
  - Added coverage for successful and failing checked-expression cases.
  - Updated expected output to verify successful results and propagated errors.


birgen/bir_gen.go
```diff
@@ -972,21 +967,29 @@ func loadLambdaFunction(ctx context, curBB *bir.BIRBasicBlock, expr *ast.BLangLa
...
-		Kind:      bir.CallKindFunctionPointer,
-		Name:      birFunc.Name,
-		FpOperand: fpEffect.result,
-	}
-	call := bir.NewCall(callSite, thenBB, resultOperand, ctx.function().loc(expr.GetPosition()))
-	fpEffect.block.Terminator = call
-	return expressionEffect{result: resultOperand, block: thenBB}
+	body, ok := expr.Lambda.Function.Body.(*ast.BLangBlockFunctionBody)
+	if !ok || len(body.Stmts) == 0 {
+		ctx.internalError("expression thunk must have a non-empty block body", expr.GetPosition())
```
warunalakshitha: will these not covered lines can happen since these expressionThunks are created from desuger never from bal files? so do we need to handle them ?

heshanpadmasiri: fixed

warunalakshitha (approved):
LGTM

PR #835, opened 2026-08-25
chore(lint): enable wasted assignment checks

heshanpadmasiri:
## Purpose

Enable the `wastedassign` linter so assignments whose values are overwritten before use are caught in CI.

## Approach

- Enable `wastedassign` in the strict and no-panic golangci-lint configurations.
- Remove assignments that were genuinely unused.
- Use the resolved intersection type when rebuilding a union type.
- Limit temporary variables to the scopes where their values are consumed.

## Testing

The existing CI workflows cover linting, native builds, WASM builds, and the affected parser and semantic-type code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected type merging during bracketed binding-pattern parsing to preserve intersection transformations.
  - Improved tuple rest-type resolution when no rest type is available.

- **Chores**
  - Enabled the `wastedassign` linter in additional linting configurations.
  - Simplified internal variable handling and removed redundant assignments without changing user-visible behavior.


warunalakshitha (dismissed):
LGTM

warunalakshitha (approved):

Issue #832, opened 2026-08-25
langlib: string:includes, string:indexOf and value:toBalString are missing


TharmiganK:
## Component

- [x] Language / interpreter (compiler, BIR, runtime)
- [ ] Dependent stdlib: `ballerina/persist`

## Summary

Three lang-library functions that `ballerina/persist` ([#784](https://github.com/ballerina-nutcracker/ballerina/issues/784)) needs are not implemented and are not covered by [#640](https://github.com/ballerina-nutcracker/ballerina/issues/640):

| Function | Needed by |
|---|---|
| `lang.string:includes` | `persist:filterRecord` — `'field.includes("[]")` and `'field.includes(".")` |
| `lang.string:indexOf` | `persist:filterRecord` — `<int>'field.indexOf(".")` |
| `lang.value:toBalString` | `persist:getNotFoundError`, `persist:getAlreadyExistsError` — `key.toBalString()` when the key is a record |

`#640` already covers the other two functions these bodies need, `lang.string:substring` and `lang.value:toString`, and [#656](https://github.com/ballerina-nutcracker/ballerina/pull/656) implements them — but it lists `indexOf` only for `lang.array`, not for `lang.string`, and does not mention `includes` or `toBalString` at all. This issue tracks just that remainder.

`lang.string` currently ships three functions in total — `length`, `toBytes`, `fromBytes` (`lib/langlibs/ballerina/lang.string/0.0.1/any/lang.string.bal`). `lang.value` ships two — `cloneWithType`, `fromJsonWithType` (`lib/langlibs/ballerina/lang.value/0.0.1/any/lang.value.bal`).

`toBalString` needs a renderer distinct from the existing `values.ToString` (`values/values.go:172`, which backs `io:println`): it produces the Ballerina *expression* form, so strings are quoted and escaped and a `()` renders as `()` rather than the empty string. That renderer does not exist in the tree today.

## Reproduction

```ballerina
import ballerina/io;

public function main() {
    string f = "location.buildingCode";
    io:println(f.includes("."));
    io:println(f.indexOf("."));

    record {} key = {id: 5};
    io:println(key.toBalString());
}
```

Run with: `go run ./cli/cmd run str.bal`

## Expected vs actual

Verified against jBallerina 2201.13.5 (Swan Lake Update 13).

**Expected:**

```
true
8
{"id":5}
```

**Actual** — each call is a separate compile error:

```
error[SEMANTIC_ERROR]: method not found: includes
[...]

PR #838, opened 2026-08-28
fix(semantics): fall back to lang.value in method-call resolution

warunalakshitha:
## Purpose

Method-call dispatch for langlib functions (`x.method(...)`) only tried the receiver's own basic-type module (`lang.map`, `lang.int`, ...) and never fell back to `lang.value`, so any `lang.value`-only method failed with "method not found" on a precisely-typed receiver — e.g. a `record`-typed value calling `cloneWithType`/`fromJsonWithType`. This is a prerequisite for #832, which needs `value:toBalString` to work on a record receiver.

Relates to #832

## Approach

Split `resolveLangLibImport` into a non-erroring `tryResolveLangLibImport` plus a thin wrapper that reports the diagnostic. `resolveMethodCall` now tries the receiver's own module first and only reports "method not found" if the `lang.value` fallback also misses.

## Checklist
- [x] Read the Contributing Guide
- [x] Added or updated corpus tests

## Summary by CodeRabbit

- **New Features**
  - `fromJsonWithType()` can now be called without explicitly providing the target type when it can be inferred.
  - Added JSON conversion support coverage for integer and record values, including scalar and object inputs.

- **Bug Fixes**
  - Improved diagnostics when the target type cannot be inferred.
  - Improved handling of value operations and missing methods across receiver types.
  - Invalid operations now report appropriate errors instead of unexpected failures.

### DIFF
--- a/semantics/internal/types/type_resolver.go
+++ b/semantics/internal/types/type_resolver.go
@@ -6389,9 +6389,14 @@ func resolveMethodCall(t typeResolver, chain *binding, expr *ast.BLangInvocation
 	case semtypes.IsSubtype(t.typeContext(), recieverTy, semtypes.XML):
 		symbolRef, pkgAlias, ok = resolveLangLibImport(t, "lang.xml", methodSymbol.MethodName(), expr)
 	default:
+		ok = false
+	}
+	// lang.value applies to any value, so it's the fallback module.
+	if !ok {
 		symbolRef, pkgAlias, ok = resolveLangLibImport(t, "lang.value", methodSymbol.MethodName(), expr)
 	}
 	if !ok {
+		t.semanticError("method not found: "+methodSymbol.MethodName(), expr.GetPosition())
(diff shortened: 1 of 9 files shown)


semantics/internal/types/type_resolver.go
```diff
@@ -6622,11 +6626,26 @@ func resolveRemoteMethodCallAction(t typeResolver, chain *binding, expr *ast.BLa
 }
 
 func resolveLangLibImport(t typeResolver, pkgName string, methodName string, expr *ast.BLangInvocation) (model.SymbolRef, ast.BLangIdentifier, bool) {
```
heshanpadmasiri, line 6738: Is there a reason for keeping resolveLangLibImport and tryResolveLangLibrImport seperate?

warunalakshitha: Fixed

heshanpadmasiri (approved):
LGTM

PR #849, opened 2026-08-31
feat(semantics): support dependent array return types

### LINKED ISSUE #831: semantics: a parameter reference in a return type is not dependent-typed unless it is the whole return type
## Component

- [ ] Dependent stdlib: `ballerina/persist`

## Summary

A `typedesc` parameter referenced from *inside* a constructed return type descriptor — `returns elementType[]` rather than `returns elementType` — is not recognised as a dependent return. Worse, the declaration is accepted with **no diagnostic at all**: the parameter name resolves to the parameter's own static type, so `returns elementType[]` silently becomes `typedesc<record {}>[]`, and every call site gets that nonsensical type instead of the intended one.

The same leak applies to a parameter that is not a `typedesc` at all. `function weird(int n, string s) returns n[]` compiles cleanly and yields `int[]`, where jBallerina reports `invalid parameter reference: expected 'typedesc', found 'int'`. So a parameter name is usable as a type descriptor anywhere in a return type, silently taking that parameter's type.

Found while porting `ballerina/persist` ([#784](https://github.com/ballerina-nutcracker/ballerina/issues/784)), whose only extern function is

```ballerina
public isolated function convertToArray(typedesc<record {}> elementType, record {}[] arr) returns elementType[] = external;
```

This is the one public symbol of that package that cannot be declared today. The rest of `persist` — the `distinct` error hierarchy, `distinct object` `AbstractPersistClient` with `*` inclusion into a `client class`, and `getKey` — already works end to end.

Root cause, two facets:

1. `returnTypeReferencesTypedescParam` (`semantics/internal/symbols/symbol_resolver.go:549`) and `buildReturnTypeOp` (`semantics/internal/types/type_resolver.go:2089`) each switch on only `BLangReturnTypeDescriptor`, `BLangUserDefinedType`, `BLangUnionTypeNode` and `BLangIntersectionTypeNode`. An array (or any other constructed) descriptor falls through, so `isDependentlyTyped` returns false and no `DependentlyTypedFunctionSymbol` is ever allocated. Relatedly, `model.TypeOp` has only `RefTypeOp`, `IdentityTypeOp` and `BinaryTypeOp` — there is no representation for "array of the dependent part", so even once detection recursed, the substitution could not be expressed.
2. Because detection declines, the node is resolved as an ordinary type. `resolveBType`'s `*ast.BLangUserDefinedType` case (`semantics/internal/types/type_resolver.go:7086`) returns `t.symbolType(symbol)` for whatever symbol the name bound to, without checking that it is a type symbol. A parameter symbol therefore yields the parameter's type rather than a diagnostic.

Fixing (1) alone would cover the `typedesc` case; (2) is what turns the non-`typedesc` case from an error into silent nonsense, and is worth guarding independently.

### A — `typedesc` parameter referenced through an array (the `persist` case)

```ballerina
import ballerina/io;

type Person record {|
    string name;
|};

public isolated function convertToArray(typedesc<record {}> elementType, record {}[] arr) returns elementType[] = external;

public function main() {
    record {}[] src = [{"name": "a"}];
    Person[] out = convertToArray(Person, src);
    io:println(out);
}
```

Run with: `go run ./cli/cmd run dep-array.bal`

### B — non-`typedesc` parameter used as a type descriptor

```ballerina
import ballerina/io;

public isolated function weird(int n, string s) returns n[] = external;

public function main() {
    string x = weird(1, "a");
    io:println(x);
}
```

Run with: `go run ./cli/cmd run rc.bal`

### C — control: the same reference as the *whole* return type is handled correctly

```ballerina
public isolated function convertOne(typedesc<record {}> elementType, record {} r) returns elementType = external;
```

This is accepted, allocated as a `DependentlyTypedFunctionSymbol`, and substituted at the call site as expected. Only the nested form fails.

## Expected vs actual

Verified against jBallerina 2201.13.5 (Swan Lake Update 13) with `bal build`.

**A — expected:** compiles; `convertToArray(Person, src)` has static type `Person[]`. jBallerina accepts the assignment on line 11 and rejects `int[] bad = convertToArray(Person, src);` with `incompatible types: expected 'int[]', found 'Person[]'` — i.e. the dependent substitution applies through the array.

**A — actual:** the *declaration* is accepted silently, and the call site fails with

```
error[SEMANTIC_ERROR]: incompatible type: expected [{| name: string, never... |}...], got [typedesc<{| anydata... |}>...]
  --> dep-array2.bal:11:20
   |
11 |     Person[] out = convertToArray(Person, src);
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The return type became `typedesc<record {}>[]` — the parameter's own type, arrayed.

**B — expected:** rejected at the declaration. jBallerina reports

```
ERROR [rc.bal:(3:57,3:58)] invalid parameter reference: expected 'typedesc', found 'int'
```

**B — actual:** the declaration compiles, `weird(...)` has static type `int[]`, and the only error surfaces downstream as `incompatible type: expected string, got [int...]`. Nothing points at the invalid parameter reference.

heshanpadmasiri:
## Summary

- recognize typedesc parameter references inside array return descriptors
- preserve open, fixed, zero-length, and nested array operations through symbol serialization
- infer typedesc defaults through array, union, and intersection operation trees
- validate explicit and inferred calls through corpus, extern-runtime, and cross-module tests

## Testing

- `go test ./model/... ./semantics/... ./corpus ./corpus/extern/...`
- `go test ./...`
- commit hook: `make -j4 build lint`

Depends on #835

Closes #831

## Summary by CodeRabbit

- **New Features**
  - Added support for dependently typed functions that infer and return open, fixed-size, zero-length, nested, union, and intersection-based arrays.
  - Extended dependent array return types across module boundaries, including nested fixed-size arrays.
  - Added XML filter expression resolution and improved readonly mapping-field mutation checks.

- **Bug Fixes**
  - Improved diagnostics for incompatible dependent array return types and invalid non-`typedesc` parameter references.
  - Fixed serialization and restoration of array type information, including nested array shapes.

### DIFF
--- a/corpus/bal/subset8/08-function/dependent-fn-array-e.bal
+++ b/corpus/bal/subset8/08-function/dependent-fn-array-e.bal
@@ -1,0 +1,44 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+function open(typedesc T = <>) returns T[] = external;
+function fixed(typedesc T = <>) returns T[2] = external;
+function zero(typedesc T = <>) returns T[0] = external;
+function nested(typedesc T = <>) returns T[][] = external;
+function outsideUnion(typedesc T = <>) returns T[]|error = external;
+function insideUnion(typedesc<anydata> T = <>) returns (T|error)[] = external;
+function insideIntersection(typedesc T = <>) returns (T & readonly)[] = external;
+function constrained(typedesc<int|string> T = <>) returns T[] = external;
+function fixedArraySibling(typedesc<string> T = <>) returns T[]|int[1] = external;
+
+type StringArray string[];
+
+public function main() {
+    string[] _ = open();
+    string[2] _ = fixed();
+    string[0] _ = zero();
+    string[][] _ = nested();
+    string[]|error _ = outsideUnion();
+    (string|error)[] _ = insideUnion();
+    (string & readonly)[] _ = insideIntersection();
+    string[][] _ = open(StringArray);
+    string[]|int[1] _ = fixedArraySibling();
+
+    string _ = open(); // @error
+    string[3] _ = fixed(); // @error
+    var _ = open(); // @error
+    float[] _ = constrained(); // @error
+}
--- a/semantics/internal/types/type_resolver.go
+++ b/semantics/internal/types/type_resolver.go
@@ -2176,7 +2176,7 @@ type param struct {
 
 // buildReturnTypeOp translates a return-type-descriptor AST node into a TypeOp tree.
 // A user-defined-type node whose name matches a typedesc parameter becomes a RefTypeOp.
-// Union and intersection nodes recurse. Everything else is resolved to a concrete semtype
+// Array, union, and intersection nodes recurse. Everything else is resolved to a concrete semtype
 // and wrapped in an IdentityTypeOp.
 func buildReturnTypeOp(t typeResolver, params map[string]param, node ast.BLangNode) (model.TypeOp, bool) {
 	switch n := node.(type) {
@@ -2202,6 +2202,24 @@ func buildReturnTypeOp(t typeResolver, params map[string]param, node ast.BLangNo
 			return nil, false
 		}
 		return &model.BinaryTypeOp{Kind: model.TypeOpIntersection, Lhs: lhs, Rhs: rhs}, true
+	case *ast.BLangArrayType:
+		element, ok := buildReturnTypeOp(t, params, n.Elemtype.TypeDescriptor.(ast.BLangNode))
+		if !ok {
+			return nil, false
+		}
+		for i := len(n.Sizes); i > 0; i-- {
+			lengthExpr := n.Sizes[i-1]
+			if lengthExpr == nil {
+				element = &model.ArrayTypeOp{Element: element, IsOpen: true}
+				continue
+			}
+			length, ok := resolveFixedArraySize(t, lengthExpr)
+			if !ok {
+				return nil, false
+			}
+			element = &model.ArrayTypeOp{Element: element, Length: length}
+		}
+		return element, true
 	case *ast.BLangUserDefinedType:
 		if n.PkgAlias.GetValue() == "" {
 			if p, ok := params[n.TypeName.Value]; ok && semtypes.IsSubtype(t.typeContext(), p.ty, semtypes.Typedesc) {
@@ -7082,7 +7100,7 @@ func lowerInvocationArgsInner(t typeResolver, args []ast.BLangExpression, sig mo
 				return nil, false
 			}
 			constraint := semtypes.TypedescConstraint(t.typeContext(), paramTypes[i])
-			defaultArg, ok := lowerInferredTypedescDefaultArg(t, constraint, expectedType, depSym.ReturnType().FixedPart(), pos)
+			defaultArg, ok := lowerInferredTypedescDefaultArg(t, constraint, expectedType, depSym.ReturnType(), i, pos)
 			if !ok {
 				return nil, false
 			}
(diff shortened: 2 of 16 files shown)


corpus/bal/subset8/08-function/dependent-fn-array-e.bal
```diff
@@ -0,0 +1,44 @@
...
+function nested(typedesc T = <>) returns T[][] = external;
+function outsideUnion(typedesc T = <>) returns T[]|error = external;
+function insideUnion(typedesc<anydata> T = <>) returns (T|error)[] = external;
+function insideIntersection(typedesc T = <>) returns (T & readonly)[] = external;
+function constrained(typedesc<int|string> T = <>) returns T[] = external;
+function fixedArraySibling(typedesc<string> T = <>) returns T[]|int[1] = external;
+
+type StringArray string[];
+
+public function main() {
```
warunalakshitha, line 29: Shall we add B — non-typedesc parameter used as a type descriptor as a error bal as well. 

According to the issue

import ballerina/io;

public isolated function weird(int n, string s) returns n[] = external;

public function main() {
    string x = weird(1, "a");
    io:println(x);
}

Should be an error.

ERROR [rc.bal:(3:57,3:58)] invalid parameter reference: expected 'typedesc', found 'int'

heshanpadmasiri:
[AI RESPONSE] failed to reproduce, can you present a test case for this (re: https://github.com/ballerina-nutcracker/ballerina/pull/849#pullrequestreview-5175513144)

The symbol-pool version field was not removed by this PR. This branch's only changes under `model/symbolpool/` are additive (`typeOpTagArray` plus the matching write/read cases for `ArrayTypeOp`); neither the header write path in `Marshal` nor the header read path in `Unmarshal` is touched. The version field was removed on `main` in commit 5e88c5e28 "refactor(codec): remove BIR and symbol pool versions", which is already an ancestor of `main`.

heshanpadmasiri:
[AI RESPONSE] ignoring due to the docstring-coverage threshold not being a convention of this repository (re: https://github.com/ballerina-nutcracker/ballerina/pull/849#issuecomment-5474657045)

`AGENTS.md` explicitly directs contributors away from prose comments in favour of self-documenting names, and neither the review guidelines nor `CONTRIBUTING.md` set a docstring requirement. Repo-wide density on `main` is ~17.5%, and the touched files sit lower still (`model/symbol.go` 3.9%, `semantics/internal/types/type_resolver.go` 9.0%). Every function this PR adds is unexported or an interface-method implementation; the `TypeOp` interface and its `FixedPart` method are already documented at the interface declaration, matching the surrounding file. The 80% figure is a generic tool default, not a repo rule.

corpus/bal/subset8/08-function/dependent-fn-array-e.bal
heshanpadmasiri: fixed

heshanpadmasiri:
@coderabbitai review

semantics/internal/types/type_resolver.go
```diff
@@ -7135,6 +7154,100 @@ func lowerInferredTypedescDefaultArg(t typeResolver, constraint, expectedType, f
 	return expr, true
 }
 
+func inferTypedescConstraint(t typeResolver, op model.TypeOp, paramIndex int, constraint, candidate semtypes.SemType, pos diagnostics.Location) (semtypes.SemType, bool, bool) {
```
warunalakshitha, line 7216: most of cases in this inferTypedescConstraint are not covered. is these are just defensive code then we can skip?

warunalakshitha: Generated some cases please check is it ok to cover these as well

These shapes are **not** covered by the current bals (we only have lhs-only `T[]|error`, inside `(T|error)[]` / `(T & readonly)[]`). They are real language cases and would help `codecov/patch` without chasing defensive exhaustiveness arms.

```ballerina
// RHS-only union (mirror of existing outsideUnion)
function rhsOnly(typedesc T = <>) returns error|T[] = external;

// Both sides depend on the typedesc
function bothSides(typedesc T = <>) returns T[]|T = external;

// Intersection outside the array (vs existing insideIntersection)
function outsideIntersection(typedesc T = <>) returns T[] & readonly = external;

public function main() {
    string[]|error _ = rhsOnly();
    string[]|string _ = bothSides();
    string[] & readonly _ = outsideIntersection();
}
```

Could live in `dependent-fn-array-e.bal` (success lines) and optionally mirrored in `dependently-typed-v.bal` for runtime. Defensive arms (wrong `RefTypeOp` index, unknown `TypeOp`, empty `ListProj`, neither-side binary) still need no bals — leave those uncovered.

heshanpadmasiri:
[AI RESPONSE] failed to reproduce, can you present a test case for this (re: https://github.com/ballerina-nutcracker/ballerina/pull/849#pullrequestreview-5175513144)

The symbol-pool version field was removed in commit `5e88c5e28` ("refactor(codec): remove BIR and symbol pool versions"), which is already on `main` and not part of this PR. This PR's only changes under `model/symbolpool/` are additive: a new `typeOpTagArray` appended to the end of the existing tag iota block, plus matching write/read cases for `ArrayTypeOp`. The header framing at `serializer.go:30` is untouched. Symbol-pool bytes are also never persisted across builds — the only callers outside the package are in-process round-trip tests.

semantics/internal/types/type_resolver.go
heshanpadmasiri: fixed

warunalakshitha (approved):
LGTM

Issue #290, opened 2026-03-23
Add support for xml navigation expressions


heshanpadmasiri:
Fallow on from #289. We need to also create a test corpus as well. Also we need to decide if we are going to represent these directly in BIR or have a internal library for this

PR #872, opened 2026-09-03
feat(xml): add XML filter expressions

### LINKED ISSUE #846: Support XML filter expressions

heshanpadmasiri:
## Purpose

Add support for XML filter expressions so XML values can be filtered by element name and namespace.

## Approach

- Add XML filter-expression nodes and resolution across the compiler pipeline.
- Lower filters to BIR and evaluate them in the runtime.
- Add corpus coverage for syntax, typing, namespaces, serialization, and execution.

## Checklist

## Remarks

Closes #846.
Part of #290.
Depends on #835.

## Summary by CodeRabbit

- **New Features**
  - Added XML navigation filters supporting wildcards, identifiers, qualified names, namespace prefixes, and nested selections.
  - Added namespace-aware XML element handling, preserving prefixes and local names during parsing, filtering, serialization, and display.
  - Added compile-time validation for invalid XML filter syntax, types, and namespace prefixes.

- **Bug Fixes**
  - XML output and comparisons now correctly distinguish namespace-qualified elements and preserve qualified names.


values/equal.go
```diff
@@ -169,7 +169,8 @@ func xmlElementDeepEqual(a, b *XMLElement, visited map[refPair]struct{}) bool {
 	if cycle {
 		return true
 	}
-	return a.Name == b.Name &&
+	return a.Prefix == b.Prefix &&
```
warunalakshitha, line 172: Should we use NamespaceURI? <a:foo xmlns:a="urn:x"/> and <b:foo xmlns:b="urn:x"/> should be equal

ast/pretty_printer.go
```diff
@@ -566,6 +568,37 @@ func (p *PrettyPrinter) printXMLTemplateExpr(node *BLangXMLTemplateExpr) {
...
+			p.PrintString("xml-name-pattern-qualified-identifier")
+			if pattern.NamespacePrefix != nil {
+				p.PrintString(pattern.NamespacePrefix.GetValue())
+			}
+			p.PrintString(pattern.Identifier.GetValue())
+		case NamePatternKindPrefix:
+			p.PrintString("xml-name-pattern-prefix-wildcard")
+			p.PrintString(pattern.NamespacePrefix.GetValue())
+		default:
+			panic("unsupported XML name pattern kind")
```
warunalakshitha: Should these be internal errors not crash with panic

warunalakshitha:
LGTM except one minor Nit

values/equal.go
heshanpadmasiri: [AI RESPONSE] failed to reproduce, can you present a test case for this.

Checked against the jBallerina reference implementation (2201.13.4). In jBallerina, `XmlItem.equals` compares the expanded name (URI + local, prefix ignored) **and** the attributes map — and `xmlns` declarations are synthesized into that attributes map. So prefix differences still surface, via the xmlns attributes.

Empirical results, jBallerina vs. this branch — they agree on every case:

| case | jBallerina | this branch |
|---|---|---|
| `<a:foo xmlns:a="urn:x"/> == <b:foo xmlns:b="urn:x"/>` | false | false |
| `<p:foo xmlns:p="urn:x"/> == <p:foo xmlns:p="urn:y"/>` | false | false |
| `<r xmlns:p="urn:x" xmlns:q="urn:x"><p:c/></r> == <r xmlns:p="urn:x" xmlns:q="urn:x"><q:c/></r>` | false | false |
| `<r xmlns:p="urn:x"><p:c/></r> == <r xmlns:p="urn:y"><p:c/></r>` | false | false |

So `<a:foo xmlns:a="urn:x"/>` and `<b:foo xmlns:b="urn:x"/>` are *not* equal in jBallerina — their xmlns attribute maps differ — and this branch already matches.

Neither failure mode is reachable as written:
- *different prefixes, same URI -> wrongly unequal*: `XMLElement.Namespaces` is compared too, and the xmlns keys embed the prefix (`xmlns:a` vs `xmlns:b`), so those pairs are unequal either way — matching jBallerina.
- *same prefix, different URI -> wrongly equal*: an element only has a `NamespaceURI` with no matching entry in its own `Namespaces` when the declaration is inherited from an ancestor (`xmlns_resolver.go` hoists every needed declaration onto the root literal). Isolating such a child as a standalone value needs XML step navigation (`root/*`), still unimplemented on this branch. Whole-tree comparison always hits the differing root `Namespaces` first, so the divergence is unobservable.

I applied the proposed diff verbatim in a scratch worktree and reran the battery: case 3 flips from `false` to **`true`**, diverging from jBallerina. Swapping `Prefix` for `NamespaceURI` is a regression as written. At most `NamespaceURI` could be added *alongside* `Prefix` — a no-op today, and only meaningful once child navigation lands, so it belongs in the navigation PR.

Happy to take a concrete failing case if one exists.

values/equal.go
```diff
@@ -169,7 +169,8 @@ func xmlElementDeepEqual(a, b *XMLElement, visited map[refPair]struct{}) bool {
...
+		a.LocalName == b.LocalName &&
```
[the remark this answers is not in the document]
heshanpadmasiri: [AI RESPONSE] failed to reproduce, can you present a test case for this.

Checked against the jBallerina reference implementation (2201.13.4). In jBallerina, `XmlItem.equals` compares the expanded name (URI + local, prefix ignored) **and** the attributes map — and `xmlns` declarations are synthesized into that attributes map. So prefix differences still surface, via the xmlns attributes.

Empirical results, jBallerina vs. this branch — they agree on every case:

| case | jBallerina | this branch |
|---|---|---|
| `<a:foo xmlns:a="urn:x"/> == <b:foo xmlns:b="urn:x"/>` | false | false |
| `<p:foo xmlns:p="urn:x"/> == <p:foo xmlns:p="urn:y"/>` | false | false |
| `<r xmlns:p="urn:x" xmlns:q="urn:x"><p:c/></r> == <r xmlns:p="urn:x" xmlns:q="urn:x"><q:c/></r>` | false | false |
| `<r xmlns:p="urn:x"><p:c/></r> == <r xmlns:p="urn:y"><p:c/></r>` | false | false |

So `<a:foo xmlns:a="urn:x"/>` and `<b:foo xmlns:b="urn:x"/>` are *not* equal in jBallerina — their xmlns attribute maps differ — and this branch already matches.

Neither failure mode is reachable as written:
- *different prefixes, same URI -> wrongly unequal*: `XMLElement.Namespaces` is compared too, and the xmlns keys embed the prefix (`xmlns:a` vs `xmlns:b`), so those pairs are unequal either way — matching jBallerina.
- *same prefix, different URI -> wrongly equal*: an element only has a `NamespaceURI` with no matching entry in its own `Namespaces` when the declaration is inherited from an ancestor (`xmlns_resolver.go` hoists every needed declaration onto the root literal). Isolating such a child as a standalone value needs XML step navigation (`root/*`), still unimplemented on this branch. Whole-tree comparison always hits the differing root `Namespaces` first, so the divergence is unobservable.

I applied the proposed diff verbatim in a scratch worktree and reran the battery: case 3 flips from `false` to **`true`**, diverging from jBallerina. Swapping `Prefix` for `NamespaceURI` is a regression as written. At most `NamespaceURI` could be added *alongside* `Prefix` — a no-op today, and only meaningful once child navigation lands, so it belongs in the navigation PR.

Happy to take a concrete failing case if one exists.

warunalakshitha (approved):
LGTM

PR #875, opened 2026-09-04
[don't merge]: improve front end perf

heshanpadmasiri:
## Summary by CodeRabbit

* **New Features**
  * Added a unified compilation pipeline with workspace and dependency discovery, concurrent module processing, cancellation support, and compilation statistics.
  * Improved `bal run` handling for workspace-relative paths, project diagnostics, native dependencies, and compiler output.
  * Added parser debug options for token, syntax-tree, and recovery tracing.
  * Added richer syntax diagnostics, including codes, messages, severity, properties, and source content.

* **Bug Fixes**
  * Improved duplicate import detection and unresolved-import diagnostics.
  * Prevented diagnostic output from using incorrect source files.
  * Stabilized generated names during concurrent compilation.

PR #878, opened 2026-09-04
[Don't merge]: measure interpreter code layout sensitivity

heshanpadmasiri:
## Not for merge

Control experiment for the execution-benchmark regressions reported on #875. **Please don't merge this** — it exists only to produce one benchmark comment.

### What #875 showed

Its benchmark table moves in two directions at once, and both directions reproduced across two independent runs on the same head SHA:

| Case | run 1 | run 2 | noise range (13 control PRs) |
|---|---|---|---|
| 08-bench/sieve | 1.529 | 1.606 | 0.961–1.177 |
| 08-bench/closure2 | 1.497 | 1.511 | 0.970–1.019 |
| 08-bench/closure1 | 1.374 | 1.354 | 0.980–1.029 |
| 07-bench/2 | 1.156 | 1.176 | 0.975–1.026 |
| 09-bench/huge-project | 0.650 | 0.651 | 0.974–1.020 |

The 09-bench wins are the intended front-end gain and reproduce locally at 1.48–1.52×. The regressions are the puzzle: they are all execution-dominated cases (compilation is under 5% of their runtime; sieve is ~97% interpretation), and three of the four are the suite's only hot-loop closure benchmarks.

But #875 changes **zero lines** in `runtime/`, `values/` and `bir/`, and `--dump-bir` output is byte-identical for all four cases. The same program runs on the same interpreter.

### What this PR tests

On a linux/amd64 cross build of #875, the interpreter's hot functions are identical machine code at different addresses. `executeBasicBlock` is 805 bytes with an identical mnemonic and instruction-length sequence in both binaries; the only 13 differing instructions are `CALL`/`CMPL`/`LEAQ` RIP-relative displacements. Across 622 `runtime`/`values`/`bir` symbols the shifts are +1920, +2400 and +544 bytes, flipping 64-byte alignment for roughly 288 of them.

So the hypothesis is that the movement is **binary code layout**, not compiler behaviour.

This PR adds 1200 generated, never-executed functions to `bir`, kept alive by a sentinel-guarded reference in `printVersion` (`Version` is `dev` or an ldflags value, never the sentinel). Nothing here runs. It reproduces the same alignment signature on exactly the same functions:

```
FUNC                       main       align   +padding   align
executeBasicBlock          0x815520   32      0x834080   0
executeBasicBlockWithTrap  0x815080    0      0x833be0   32
executeCall                0x81f720   32      0x83e280   0
executeFunction            0x814420   32      0x832f80   0
executeFunctionNoTrap      0x815020   32      0x833b80   0
executeFunctionWithTrap    0x814e40    0      0x8339a0   32
```

`executeBasicBlock` remains 805 bytes of identical instructions. Benchmark outputs verified correct (sieve `524287`, closure2 `25000160000000`).

### How to read the result

- **closure1 / closure2 / 07-bench/2 / sieve regress here too** → layout is the mechanism, #875's compiler changes are exonerated, and the suite needs rethinking for execution-dominated cases (they cannot distinguish a real interpreter regression from a relink).
- **They don't regress** → layout is ruled out and the cause is genuinely inside #875.

One caveat on fidelity: because the padding sits in `bir`, which links *before* `runtime/internal/exec`, exec shifts by the full +125,792 bytes here, whereas in #875 it shifts +2400 with the bulk of the added code landing after it. The 64-byte alignment flip is identical; the absolute displacement is not.

The `Lint PR` title check will fail on the `[Don't merge]` prefix, same as on #875.

## Summary by CodeRabbit

- **Chores**
  - Updated internal version handling without changing the information displayed by the version command.
  - No user-visible features, behavior changes, or configuration updates are included.

### DIFF
--- a/bir/layout_pad.go
+++ b/bir/layout_pad.go
@@ -1,0 +1,14435 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Generated inert code for a benchmark layout experiment. Nothing here is
+// reachable at run time; its only purpose is to occupy .text so that the
+// packages linked after this one land at different addresses. See the PR
+// description for what this is controlling for.
+
+//nolint:all // generated padding, deliberately meaningless
+package bir
+
+// layoutPadTable keeps the padding functions reachable so the linker retains them.
+var layoutPadTable = [...]func(int) int{
+	layoutPad0,
+	layoutPad1,
+	layoutPad2,
+	layoutPad3,
+	layoutPad4,
+	layoutPad5,
+	layoutPad6,
+	layoutPad7,
+	layoutPad8,
+	layoutPad9,
+	layoutPad10,
+	layoutPad11,
+	layoutPad12,
+	layoutPad13,
+	layoutPad14,
+	layoutPad15,
+	layoutPad16,
+	layoutPad17,
+	layoutPad18,
+	layoutPad19,
+	layoutPad20,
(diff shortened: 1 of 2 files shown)


heshanpadmasiri:
@coderabbitai pause

heshanpadmasiri:
Control experiment done — result recorded below. Code layout is **not** the mechanism.

A +125,792-byte shift that reproduces the identical 64-byte alignment flip on the identical interpreter functions moved nothing: closure1 1.009, closure2 1.021, sieve 0.980, 07-bench/2 0.996, intmergesort 0.997, mergesort 0.999 (against 1.35–1.61× on #875). Closing; not for merge.

PR #879, opened 2026-09-04
[Don't merge]: bisect frontend API commit for exec regression

heshanpadmasiri:
## Not for merge

Bisect step for the execution-benchmark regression on #875. **Please don't merge** — it exists only to produce one benchmark comment.

This branch is `main` plus **exactly one commit**, `refactor(frontend): expose explicit compilation APIs` (`e38cfb080`), whose parent is current `main` (`bb5298bba`). Single variable, no cherry-pick.

### Why

After the rebase, #875 is just two commits. The regression is reproducible across four independent CI runs (two before the rebase, two after):

| Case | pre r1 | pre r2 | rebase1 | rebase2 | mean | noise range |
|---|---|---|---|---|---|---|
| 08-bench/sieve | 1.529 | 1.606 | 1.493 | 1.496 | 1.531 | 0.961–1.177 |
| 08-bench/closure2 | 1.497 | 1.511 | 1.443 | 1.488 | 1.485 | 0.970–1.019 |
| 08-bench/closure1 | 1.374 | 1.354 | 1.291 | 1.357 | 1.344 | 0.980–1.029 |
| 07-bench/2 | 1.156 | 1.176 | 1.161 | 1.148 | 1.160 | 0.975–1.026 |
| 09-bench/huge-project | 0.650 | 0.651 | 0.689 | 0.613 | 0.651 | 0.974–1.020 |

Noise ranges are per-case, from the benchmark tables of 13 unrelated PRs. The 09-bench win is the intended front-end gain and reproduces locally at 1.48–1.52×.

What makes the regressions strange: the affected cases are execution-dominated (compilation is under 5% of their runtime; sieve is ~97% interpretation), yet the PR changes **zero lines** in `runtime/`, `values/`, `bir/`, `semtypes/`, `lib/` and `decimal/`, and `--dump-bir` is byte-identical for all four. The same program runs on the same interpreter. Three of the four are also the suite's only hot-loop closure benchmarks.

Already ruled out by experiment: worse codegen (BIR diff), langlib changes, runtime construction, retained compiler heap (`gctrace` identical), GC/GOMAXPROCS tuning (absent from the diff), extra OS threads (`schedtrace`: 18 both), core count (GOMAXPROCS 1/2/4/12), measurement artifact (a second run reproduced everything), and **binary code layout** — #878 added 1200 inert never-executed functions reproducing the identical 64-byte alignment flip on the identical interpreter functions, and came back flat (closure1 1.009, closure2 1.021, sieve 0.980).

It also does not reproduce on darwin/arm64 at any core count: against this same pair, sieve 0.965, closure2 0.983, 07-bench/2 0.894 — HEAD equal or faster.

This commit touches no interpreter code either (mostly `parser`, `st`, `semantics`, `tools/diagnostics`), so on the face of it it should be flat.

### How to read the result

- **Regressions appear here** → the cause is in this frontend API refactor, which would be genuinely surprising and narrows the search to 36 files.
- **Flat** → the cause is `feat(driver): replace project-based run compilation`, which is where concurrent per-module compilation lives. That points at the execution phase inheriting a differently-allocated heap, and the follow-up is to measure `AnonHugePages` from `/proc/self/smaps_rollup` and the type-env size at `rt.Init` on linux.

The `Lint PR` title check will fail on the `[Don't merge]` prefix, same as on #875. The commit itself is unmodified from #875, so `Validate Commit Messages` should pass.

## Summary by CodeRabbit

* **New Features**
  * Added configurable parser diagnostics for token dumps, syntax-tree output, and recovery tracing.
  * Added parser support for custom source identities and debug output destinations.
  * Added strict and recovery-based compilation options for source processing.
  * Diagnostic environments can now separate internal file identities from displayed filenames.

* **Bug Fixes**
  * Improved source snippets in diagnostics by using in-memory document content when available.
  * Duplicate import aliases are now reported consistently.
  * Improved handling of invalid top-level syntax during recovery.

### DIFF
--- a/ast/statements.go
+++ b/ast/statements.go
@@ -127,8 +127,9 @@ type (
 		// LockKey is the content-addressed identifier of the restricted
 		// variable, set by the lock analyzer. For a module-level isolated
 		// variable it has the form "org/pkg:varName"; for a non-immutable
-		// field of an isolated class accessed via self it has the form
-		// "org/pkg:ClassName.fieldName". Empty when the lock body has no
+		// field of an isolated class or service accessed via self it has the
+		// form "org/pkg:ClassName.fieldName" or
+		// "org/pkg:$service$N.fieldName". Empty when the lock body has no
 		// restricted variable (we treat this as a semantic error though
 		// spec doesn't).
 		LockKey string
--- a/birgen/bir_gen.go
+++ b/birgen/bir_gen.go
@@ -41,6 +41,7 @@ func birLoc(de *diagnostics.DiagnosticEnv, pos diagnostics.Location) bir.Locatio
 type packageContext struct {
 	CompilerContext *compilerctx.CompilerContext
 	packageID       *model.PackageID // Current package ID
+	sourcePrefix    string
 	birPkg          *bir.BIRPackage
 	typeCtx         semtypes.Context
 	// PR-TODO: extract them to memoized types struc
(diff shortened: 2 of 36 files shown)


heshanpadmasiri:
Bisect result: **flat**. The frontend API refactor commit is not the cause.

| Case | #875 (4 runs) | this commit alone | noise |
|---|---|---|---|
| sieve | 1.493–1.606 | **1.013** | 0.961–1.177 |
| closure2 | 1.443–1.511 | **1.039** | 0.970–1.036 |
| closure1 | 1.291–1.374 | **0.980** | 0.980–1.029 |
| 07-bench/2 | 1.148–1.176 | **1.027** | 0.975–1.026 |
| 09-bench/huge-project | 0.613–0.689 | **1.002** | 0.974–1.020 |

So both the regression *and* the 09-bench compile win come entirely from `feat(driver): replace project-based run compilation`. Closing; not for merge.

Issue #876, opened 2026-09-04
[Bug]: readonly fields in mapping constructors are silently ignored


heshanpadmasiri:
### Description

The `readonly` prefix on a mapping-constructor specific field is parsed but has no effect on the constructed value. The field's read-only bit is never set, so a field declared `readonly` in a mapping constructor can still be mutated.

```ballerina
import ballerina/io;

public function main() {
    map<int> m = {readonly a: 1};
    m["a"] = 2;
    io:println(m);
}
```

Expected: a runtime panic on the update of a read-only field.

Actual: prints `{"a":2}` — the field was mutated, with no diagnostic at compile time or runtime.

### Steps to Reproduce

1. Save the program above as `readonly.bal`.
2. Run `go run ./cli/cmd run readonly.bal`.
3. Observe `{"a":2}` instead of a panic.

### Affected Version(s)

Reproduced on `main` as of v0.07.0 development.

### Root cause

`BLangMappingKeyValueField.Readonly` is populated in the node builder (`nodebuilder/node_builder.go`) but never read by any later stage, so the prefix is silently discarded. The BIR-level `IsReadonly` on `NewMap` reflects only the whole-mapping read-only bit derived from the contextual type being a subtype of `readonly`, not per-field readonly.

Per the specification: "A specific-field that starts with `readonly` will also be constructed with its read-only bit on. The read-only bit of the mapping value or field will be set only after the mapping value has been fully constructed, including any field specified by a computed-name-field."

As of #859 both `{readonly x}` and `{readonly x: 1}` are rejected with an `UNIMPLEMENTED_ERROR`, so the gap is no longer silent — but it is still unimplemented.

### Suggested fix

- Carry per-field readonly through to BIR and set the field's read-only bit after the mapping value is fully constructed, per the spec.
- Apply to both the explicit `{readonly x: v}` and shorthand `{readonly x}` forms.
- Remove the `UNIMPLEMENTED_ERROR` guard added in #859 and promote `corpus/bal/subset10/10-mapping/readonly-*-fv.bal` to `-v` tests.
- Add a `-p` test covering mutation of a readonly field.

### Related area

Compilation

### Related issue(s) (optional)

Parent: #856
Blocked by / follows: #859

Issue #858, opened 2026-09-03
[New Feature]: Support spread fields in mapping constructors


heshanpadmasiri:
### Description

The parser recognizes mapping spread fields, but the node builder panics in `transformMappingConstructorExpression`.

```ballerina
import ballerina/io;

public function main() {
    record {| int x; int y; |} first = {x: 1, y: 2};
    map<int> result = {...first, z: 3};
    io:println(result);
}
```

Expected output:

```text
{"x":1,"y":2,"z":3}
```

The source uses a closed record so its static type excludes `z`. Verified with jBallerina 2201.13.4: this example produces the output above. The original example declared `first` as `map<int>` and is a compile-time error, because that type permits `z`:

```text
invalid usage of mapping constructor expression: key 'z' may duplicate a key specified via spread field '...first'
```

### Describe your problem(s)

Valid mapping and record spread fields cannot pass AST construction. The compiler crashes rather than validating the spread operand.

### Describe your solution(s)

- Build an AST representation for mapping spread fields.
- Require spread operands to be subtypes of `map<any|error>`.
- Reject statically possible duplicate keys between a spread and a specific field, or between spreads. Optional fields can collide; semantically empty fields such as `never x?` cannot.
- Preserve insertion order.
- Evaluate each spread expression exactly once.
- Enforce contextual typing and readonly rules.
- Lower spreads before BIR generation where practical.
- Add valid, semantic-error, and runtime corpus tests.

### Related area

Compilation

### Related issue(s) (optional)

Parent: #856

Issue #856, opened 2026-09-03
[New Feature]: Remove restrictions on mapping constructors


heshanpadmasiri:
### Description

Mapping constructors currently support explicit key/value fields and computed-name fields, but AST construction panics for spread fields and shorthand variable-name fields.

```ballerina
import ballerina/io;

public function main() {
    map<int> base = {x: 1};
    int count = 2;
    map<int> result = {...base, count};
    io:println(result);
}
```

Expected output:

```text
{"x":1,"count":2}
```

### Describe your problem(s)

The parser recognizes all mapping-constructor field forms, but the node builder rejects spread fields and shorthand fields. This prevents otherwise valid record and map construction and produces a Go panic rather than a compiler diagnostic.

### Describe your solution(s)

Track completion of all mapping-constructor field forms:

- Support spread fields such as `{...base}`.
- Support shorthand variable-name fields such as `{count}`.
- Preserve specified field ordering and conflict behavior.
- Evaluate every field expression exactly once.
- Enforce contextual typing and readonly rules.
- Add valid, error, and runtime corpus tests.

The individual field forms will be implemented as native sub-issues.

### Related area

Compilation

### Related issue(s) (optional)

Sub-issues will track spread fields and shorthand fields.

PR #880, opened 2026-09-04
feat: support shorthand fields in mapping constructors

### LINKED ISSUE #859: [New Feature]: Support shorthand fields in mapping constructors
### Description

The shorthand `{name}` form is recognized by the parser, but the node builder panics when a specific field has no explicit value expression.

```ballerina
import ballerina/io;

public function main() {
    string name = "Ada";
    int age = 36;
    var person = {name, age};
    io:println(person);
}
```

Expected output:

```text
{"name":"Ada","age":36}
```

### Describe your problem(s)

Mapping-constructor shorthand fields cannot be compiled even though equivalent explicit fields work.

### Describe your solution(s)

- Translate a shorthand field into an identifier key and a variable-reference value.
- Preserve quoted-identifier behavior.
- Resolve the referenced symbol normally.
- Diagnose undefined variables and duplicate keys.
- Support contextual record and map construction.
- Add valid, semantic-error, and runtime corpus tests.

### Related area

Compilation

Parent: #856

heshanpadmasiri:
## Purpose

The shorthand `{name}` form parses fine, but the node builder bailed out with
`UNIMPLEMENTED_ERROR: mapping constructor var-name field not implemented`, so shorthand
fields could not be compiled even though the equivalent explicit fields worked.

Fixes #859

## Approach

The specification defines `{name}` as exactly equivalent to `{name: name}`, so the fix is a
total expansion in the node builder — no parser, semantics, desugar or BIR work is needed.

- **Expand the shorthand.** In the `st.SPECIFIC_FIELD` branch of
  `transformMappingConstructorExpression`, a field with `ValueExpr() == nil` now takes its
  value from `n.createExpression(specificField.FieldName())`.
  `createActionOrExpressionInner` already turns a bare `IDENTIFIER_TOKEN` into a
  `*ast.BLangVarRef` positioned on the token, so no helper is added and no var-ref is built
  by hand. The key side is untouched: `createSpecificFieldNameLiteral` already produces the
  identifier-key literal, and both sides run through `createBLangNameReference`, so the
  quoted-identifier prefix is stripped consistently (`{'default}` → key `"default"`).
- **Nothing downstream can tell the forms apart.** After the node builder a shorthand field
  is an ordinary `*ast.BLangMappingKeyValueField`, so symbol resolution, contextual record
  and `map<T>` construction, duplicate-key detection, const evaluation, CFG, desugar and
  BIR generation all apply unchanged.
- **Reject the per-field `readonly` prefix** with an `UNIMPLEMENTED_ERROR` positioned on the
  keyword. See the behaviour change below.
- **Drop the write-only `Readonly` field** from `ast.BLangMappingKeyValueField`. It was
  assigned in the node builder and read by no stage; the BIR-level `IsReadonly` on `NewMap`
  comes from the whole-mapping read-only bit, not from this field.

## Behaviour change

`{readonly x: 1}` and `{readonly "k": 1}` compile today with the modifier **silently
discarded** — the read-only bit is never set, so the runtime value is wrong. Both now report
`fatal[UNIMPLEMENTED_ERROR]: readonly mapping constructor field not implemented`. This turns
two currently-"working" programs into compile errors, deliberately: making the gap loud is
better than producing wrong runtime behaviour. Implementing the semantics is #876, and the
three `-fv` tests below are written in their eventual `-v` shape so that issue promotes them
by renaming the file and regenerating the golden.

`{_}` now reports `INTERNAL_ERROR: invalid identifier` instead of the var-name error,
matching what `{_: 1}` already reports today. That `_`-as-a-key problem predates this change
and affects both spellings equally, so it is left alone rather than special-cased for the
shorthand form.

## Checklist

## Tests

16 corpus tests in a new `corpus/bal/subset10/10-mapping/`:

- **5 valid** — `shorthand-v` (the issue's program, plus a const and a quoted identifier),
  `shorthand-mixed-forms-v` (shorthand interleaved with explicit and string-literal-key
  fields in both orders, and a contextual record), `shorthand-contextual-v` (record, `map<int>`,
  reassignment, and construction-time value capture), `shorthand-const-v` (the only test
  reaching `evaluateMappingConstructor`), `shorthand-symbol-kinds-v` (a function-valued and a
  typedesc-valued field).
- **8 error** — undefined symbol; the rest-field key rule; a duplicate-key matrix covering
  every pairing of field forms, both key kinds, and the quoted/unquoted spellings of one
  identifier; and `shorthand-stmt-start-e`, which is the only test reaching the two
  statement-start producers of the shorthand shape (`parseIdentifierRhsInStmtStartingBrace`
  and `parseMappingFieldRhs`).
- **3 future** — one per `readonly` spelling, each asserting the error lands on the keyword.

Every `-v` program was also run under jBallerina and produces **byte-identical stdout** to
its golden; every `-e` program's diagnostic **position matches jBallerina exactly**; and all
three `-fv` programs compile and run under jBallerina printing exactly their `@output`
markers.

## Remarks

- Spread fields (`{...base}`) remain unimplemented (#858), so the parent issue #856's
  motivating example `{...base, count}` still will not compile.
- Mapping binding patterns (`{name, age} = person`) take the assignment path and are
  unaffected — they still stop at `binding-pattern assignment is not implemented`.
- `doc/lang/` is not updated: subset10 has no subset doc yet, and the stale spread-field
  bullet in `subset8.md`/`subset9.md` is #858's to remove.

## Summary by CodeRabbit

- **New Features**
  - Added support for shorthand fields in mapping constructors, including contextual values, constants, mixed explicit/shorthand forms, functions, and typedescs.
  - Mapping fields without explicit values now use their field names as value expressions.

- **Bug Fixes**
  - Duplicate mapping keys are reported consistently across shorthand, quoted, string, and explicit forms.
  - Invalid shorthand usage produces clearer compiler diagnostics.
  - Readonly mapping constructor fields are explicitly reported as unsupported.

- **Tests**
  - Added comprehensive coverage for valid shorthand mappings and invalid mapping-constructor cases.

### DIFF
--- a/corpus/bal/subset10/10-mapping/shorthand-v.bal
+++ b/corpus/bal/subset10/10-mapping/shorthand-v.bal
@@ -1,0 +1,32 @@
+// Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+
+import ballerina/io;
+
+const string GREETING = "hi";
+
+public function main() {
+    string name = "Ada";
+    int age = 36;
+    string 'default = "d";
+    var person = {name, age};
+    io:println(person); // @output {"name":"Ada","age":36}
+    var c = {GREETING};
+    io:println(c); // @output {"GREETING":"hi"}
+    var q = {'default};
+    io:println(q); // @output {"default":"d"}
+}
--- a/ast/expressions.go
+++ b/ast/expressions.go
@@ -514,7 +514,6 @@ type (
 		bLangNodeBase
 		Key       *BLangMappingKey
 		ValueExpr BLangExpression
-		Readonly  bool
 	}
 
 	BLangMappingConstructorExpr struct {
--- a/corpus/ast/subset10/10-mapping/shorthand-const-v.txt
+++ b/corpus/ast/subset10/10-mapping/shorthand-const-v.txt
@@ -1,0 +1,20 @@
+(compilation-unit
+(package-id $anon . 0.0.0)
+  (import-package ballerina io (as io))
+  (const A (
+    (value-type int)) (
+    (literal 1)))
+  (const M (
+    (constrained-type
+      (builtin-ref-type map)
+      (value-type int))) (
+    (mapping-constructor-expr
+      (key-value
+        (literal A)
+        (simple-var-ref A)))))
+  (function main () (
+    (value-type null))
+    (block-function-body
+      (expression-stmt
+        (invocation io println (
+          (simple-var-ref M)))))))
--- a/corpus/ast/subset10/10-mapping/shorthand-contextual-v.txt
+++ b/corpus/ast/subset10/10-mapping/shorthand-contextual-v.txt
@@ -1,0 +1,60 @@
+(compilation-unit
+(package-id $anon . 0.0.0)
+  (import-package ballerina io (as io))
+  (type-definition Person
+    (record-type
+      (field name
+        (value-type string))
+      (field age
+        (value-type int))))
+  (function main () (
+    (value-type null))
+    (block-function-body
+      (var-def
+        (variable name (type
+          (value-type string)) (expr
+          (literal Ada))))
+      (var-def
+        (variable age (type
+          (value-type int)) (expr
+          (literal 36))))
+      (var-def
+        (variable p (type
+          (user-defined-type Person)) (expr
+          (mapping-constructor-expr
+            (key-value
+              (literal name)
+              (simple-var-ref name))
+            (key-value
+              (literal age)
+              (simple-var-ref age))))))
+      (expression-stmt
+        (invocation io println (
+          (simple-var-ref p))))
+      (var-def
+        (variable m (type
+          (constrained-type
+            (builtin-ref-type map)
+            (value-type int))) (expr
+          (mapping-constructor-expr
+            (key-value
+              (literal age)
+              (simple-var-ref age))))))
+      (expression-stmt
+        (invocation io println (
+          (simple-var-ref m))))
+      (assignment
+        (simple-var-ref m)
+        (mapping-constructor-expr
+          (key-value
+            (literal age)
+            (simple-var-ref age))))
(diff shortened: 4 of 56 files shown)


corpus/bal/subset10/10-mapping/shorthand-v.bal
```diff
@@ -0,0 +1,32 @@
...
+public function main() {
+    string name = "Ada";
+    int age = 36;
+    string 'default = "d";
+    var person = {name, age};
+    io:println(person); // @output {"name":"Ada","age":36}
+    var c = {GREETING};
+    io:println(c); // @output {"GREETING":"hi"}
+    var q = {'default};
+    io:println(q); // @output {"default":"d"}
```
warunalakshitha, line 31: shall we cover these scenarios as well.
```
int age = 36;
map<int> m = {age};   // {"age":36} — accepted

string name = "Ada";
map<int> m = {name};  // error: string value, int expected
```

heshanpadmasiri:
[AI RESPONSE] ignoring due to the readonly handling being a forced consequence of the shorthand change rather than independent scope creep (re: https://github.com/ballerina-nutcracker/ballerina/pull/880#issuecomment-5538906034)

Detail on the "Out of Scope Changes" pre-merge check:

- **The readonly guard is not separable.** `{readonly age}` is a single `specific-field` production combining the readonly qualifier with a var-name field. On `main` it is rejected by the `mapping constructor var-name field not implemented` guard — the very guard #859 removes. Removing it *newly* makes `{readonly age}` compile with the modifier silently dropped, because `Readonly` was write-only. So this PR cannot be readonly-free; it can only choose between silently-wrong output and a loud error, on the same lines it already had to rewrite.
- **The AST field removal is dead-code removal.** `BLangMappingKeyValueField.Readonly` was assigned in the node builder and read by no stage (`birgen` derives `NewMap`'s `isReadonly` from `semtypes.IsSubtype(..., ValReadonly)` on the whole mapping type). Once the guard makes it unconditionally false, it is a stored-but-never-read field.
- **The `-fv` tests are required by convention.** `AGENTS.md` defines the `-fv` suffix as exactly this shape: valid tests in scope that we have decided not to support yet, triggering an unimplemented error at the marked location. A new `UNIMPLEMENTED_ERROR` path must ship with its corpus coverage in the same PR.
- **The alternative resolution the check itself offers is already met.** Issue #876 ("readonly fields in mapping constructors are silently ignored") is a sibling sub-issue of the same parent #856 and is already linked from the PR description, which explains that the three `-fv` tests are written in their eventual `-v` shape so #876 can promote them by renaming and regenerating goldens.

Splitting this out would mean either landing a shorthand PR that silently miscompiles `{readonly age}`, or hand-splitting one `if` condition across two PRs and re-touching the same lines.

corpus/bal/subset10/10-mapping/shorthand-v.bal
heshanpadmasiri: fixed

heshanpadmasiri:
Superseded by #904, which is the same commit pushed to `ballerina-nutcracker/ballerina` so that the readonly-field work (#898) can be stacked on top of it.

PR #887, opened 2026-09-07
chore(build): update Go toolchain to 1.27.1

heshanpadmasiri:
Bump the go directive to 1.27 and pin toolchain go1.27.1 across `go.work` and all 24 module `go.mod` files, and move CI (9 workflows) to go-version 1.27.1. Raise nativerunner `MinGoVersion` to 1.27 and the generated driver workspace to go 1.27 so native interpreter builds match the shipped tree.

Update golangci-lint from v2.10.1 to v2.13.2 in CI, the commit-msg hook and DEVELOPING.md. Rename the Bala platform identifier `go1.26` -> `go1.27` across the bundled stdlibs, corpus and projects testdata, and the `.agents` skill docs that reference that layout.

Fix the two findings the newer toolchain and linter surface:

- `reflect.Ptr` -> `reflect.Pointer` in `common/tomlparser`; Go 1.27 vet adds the inline analyzer and reports the deprecated alias. Same constant value.
- Drop the never-forwarded `suffixes` parameter from `getNextAnonymousTypeKey`, flagged by `unparam`.

### Regression tests

The reason the stale references in #887's first round survived review is that
nothing in the suite was coupled to the toolchain version. Confirmed by
mutation: reverting `MinGoVersion` to `1.26` passed the entire `./cli/...`
suite. The version lived in five uncoupled copies — `go.work`, the per-module
`go.mod` directives, `MinGoVersion`, the CI `go-version` pins, and the `go1.X`
platform segment of the bundled stdlib balas — with nothing making them agree.

New `cli/internal/nativerunner/toolchain_consistency_test.go` couples them,
anchored on `MinGoVersion` because it is the one copy compiled into the binary:

| Test | Catches |
|---|---|
| `TestMinGoVersionMatchesWorkspace` | `MinGoVersion` drifting from `go.work`'s `go` directive |
| `TestModuleDirectivesMatchWorkspace` | any of the 24 `go.mod` files left behind, or a module added later |
| `TestCIGoVersionMatchesToolchain` | a workflow `go-version` disagreeing with the `toolchain` pin |
| `TestBundledStdlibPlatformMatchesToolchain` | a bundled bala claiming a toolchain the tree no longer uses |
| `TestGolangciLintPinsAgree` | the linter pin splitting across CI, the commit-msg hook and DEVELOPING.md |

Each is mutation-verified — reverting any single copy to 1.26 now fails.

Why `MinGoVersion` is the anchor: it gates `Available()` and is stamped into
the `go` directive of every generated native module, so drift below the tree's
own directive green-lights a toolchain that cannot build the interpreter —
papered over by a `GOTOOLCHAIN=auto` download, but a hard
`go.mod requires go >= X` failure under `GOTOOLCHAIN=local` or offline,
surfacing far from its cause. A stale bala platform directory is invisible to
resolution, which matches on the bare `go` prefix. A split linter pin shows up
only as a contributor whose pre-commit lint disagrees with CI.

Also addresses @warunalakshitha's review: the nativerunner tests no longer
hardcode a threshold literal, which is what let stale values sit in them
unnoticed. The `goVersionAtLeast` call sites pass `MinGoVersion`; the
`versionAtLeast` table moves to deliberately synthetic versions (it tests a
general comparator, and pinning it to the real toolchain made it *read* as a
`MinGoVersion` assertion without being one) and gains the multi-digit
component case it lacked; and new boundary tests derive their stubs from
`MinGoVersion` to pin both sides of the gate — `TestAvailable_GoTooOld`'s
`go1.20` stub sat far enough below any plausible minimum to survive a stale
constant. The generated manifests are asserted too: the emitted workspace `go`
directive, its `MinGoVersion` fallback, and the per-payload `go.mod` via a
`nativeModuleGoMod` helper extracted from `buildOrReuse`, previously written
straight into a temp dir before `go build` with nothing reading it back.

---

Part 1 of 3 of a stack split out of #875:

1. **#887** `chore/go-toolchain-1.27` (this PR, base `main`)
2. #888 `refactor/frontend-compile-api`
3. #889 `feat/driver-run-compilation`


heshanpadmasiri:
CodeRabbit skipped this PR: 173 changed files against a 100-file limit. The change is a mechanical toolchain/lint version bump plus the two findings the new toolchain surfaces, so it is going forward without an automated review. The two behavioural hunks worth human eyes are `reflect.Ptr` -> `reflect.Pointer` in `common/tomlparser` and dropping the unused `suffixes` parameter from `getNextAnonymousTypeKey`; everything else is version strings and a `go1.26` -> `go1.27` identifier rename.

heshanpadmasiri (commented):
This masks a perf regression for some compute bound benchmarks I observed #889 that in linux that I couldn't reproduce in darwin. Also I couldn't reproduce it in linux in go 1.27

cli/internal/nativerunner/local_executor_test.go
warunalakshitha: Should we change these?

warunalakshitha: Should these be 1.27?

warunalakshitha: same

heshanpadmasiri: Yes — fixed, this now passes `MinGoVersion` instead of the literal.

Worth noting the literal was never just cosmetic here: it meant the test could not observe `MinGoVersion` at all. I mutation-tested it — setting `MinGoVersion` back to `"1.26"` passed the entire `./cli/...` suite. Passing the constant makes this drift-proof rather than needing a bump every release.

heshanpadmasiri: Same fix — passes `MinGoVersion` now. This test only cares that malformed `go version` output is rejected, so the threshold was arbitrary; taking it from the constant means it stops going stale.

heshanpadmasiri: Deliberately not, and I have added a comment saying so. This table exercises `versionAtLeast(a, b)`, a general dot-separated comparator — the versions are sample data, so pinning them to the real toolchain would make the table *read* as a `MinGoVersion` assertion without being one, which is what let the stale values sit here unnoticed. I have moved them to obviously-synthetic `1.7`/`1.8` and added the multi-digit-component case the table was missing (`{"1.10", "1.9", true}` — a lexicographic comparator gets that wrong).

The gap your three comments exposed is that nothing was actually tracking the constant, so I have added tests that do:

- `TestAvailable_GoAtMinVersion` / `TestAvailable_GoJustBelowMinVersion` pin both sides of the gate, deriving their stubs from `MinGoVersion`. `TestAvailable_GoTooOld` could not: its `go1.20` stub sits far enough below any plausible minimum to pass under a stale constant.
- A new `toolchain_consistency_test.go` couples the five uncoupled copies of the version — `go.work`, the 24 `go.mod` directives, `MinGoVersion`, the 9 CI `go-version` pins, and the `go1.X` platform segment of the bundled stdlib balas. Each of those four tests is mutation-verified: reverting any one copy to 1.26 now fails.

The one that matters most is `MinGoVersion` == `go.work`s `go` directive. `MinGoVersion` gates `Available()` and is stamped into the `go` directive of every generated native module, so drift below the tree`s own directive green-lights a toolchain that cannot build the interpreter — silently papered over by a `GOTOOLCHAIN=auto` download, but a hard `go.mod requires go >= X` failure under `GOTOOLCHAIN=local` or offline, surfacing far from its cause.

azinneera (approved):
LGTM

warunalakshitha (approved):

PR #900, opened 2026-09-08
refactor(semantics): fold constants to a value or a diagnostic

### LINKED ISSUE #896: [Task]: Refactor constant folding logic
### Description

According to spec all results of the constant expression have a compile time know shape. (IE there is no runtime semantics for these). This means for any valid constant expression constant folding should produce a value or folding error (example: 1/0). We need to refactor the code to reflect that. 

### Describe your task(s)

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

heshanpadmasiri:
## Purpose

Resolves #896.

Spec §6.4 gives every constant expression a compile-time shape — there is no runtime semantics for one. So for any structurally valid constant expression, folding must either produce a value or report a compile-time error. The code did neither reliably:

- `evaluateConstantExpression` returned an opaque `error`. `resolveConstant` collapsed every one of them into a single generic `expression is not a constant expression` diagnostic pointed at the whole initializer, `resolveAnnotationAttachments` prefixed them with `cannot evaluate annotation constant expression:`, and `resolveXMLNS` discarded the message entirely.
- When folding failed, the evaluator silently fell back to reading a singleton shape off the *determined type* (`constantSingleShapeValue`) for constant references, string concatenation and string templates. That made a fold failure invisible rather than an error, and made the value depend on whether type resolution happened to produce a singleton.
- Structural non-constness was checked in two unrelated places — `common.ValidateConstantExpr` from the constant analyzer and `resolveXMLNS`, and `errNotConstantExpression` from the evaluator — which is why the analyzer had to re-walk the expression after type resolution had already walked it.

This is the second PR in the stack; it sits on #897.

## Approach

Split the two questions the old `error` was conflating.

**`isConstantExpression` (`constant_evaluator.go`)** — answers only whether an expression is *structurally* a constant expression, returning the first offending subexpression. It reports no diagnostic, because `resolveAnnotationAttachments` needs the answer without one: a non-constant value on a non-`const` annotation is legal and becomes a runtime annotation global. It replaces `common.ValidateConstantExpr` (deleted), so both callers now share one walk.

**`foldConstant` (`constant_evaluator.go`)** — called only on an expression `isConstantExpression` has accepted, so a failure is never structural. It reports its own diagnostic at the *failing subexpression* and returns `false`; the whole evaluator now returns `(values.BalValue, bool)` instead of `(values.BalValue, error)`. A nil value with `ok` is a successful fold of `()`, which is why failure is signalled by the bool alone. It also absorbs the panic recovery that used to live in `evaluateAnnotationValue`.

Failures are now classified by who is at fault:

| kind | cases |
|---|---|
| `semanticError` | what the program got wrong: `1 / 0`, `1 % 0`, integer/decimal overflow, `<int>(1.0 / 0.0)`, a list spread member that is not a list, a missing filler value |
| `unimplemented` | constant forms we do not support yet: computed mapping keys, mapping spread fields, xml templates |
| `internalError` | states type resolution should already have excluded: an unexpected expression node, a non-boolean condition operand, an unknown operator |

`constantSingleShapeValue` and the `errNotConstantExpression` sentinel are gone. Dropping the singleton fallback also means `resolveConstant` no longer needs the initializer type as a backstop — the symbol type is always `values.SemTypeForValue(value)` — and `constantAnalyzer` no longer re-validates constness, since reaching it means type resolution found nothing to report.

Two smaller consequences of pointing diagnostics at the offending node: `evaluateConstantReference` and the mapping-key helper now take a position rather than a `SemType`, and `constantConversionError` includes the offending float value (`float value Infinity cannot be converted to int`) now that the message is no longer swallowed.

## Automation tests

Every changed diagnostic is pinned by an integration golden. The new cases cover each of the three failure classes and the fold paths the removed fallback used to mask.

Semantic errors — `corpus/bal/subset8/08-const/`: `div-zero-e`, `mod-zero-e`, `add-overflow-e`, `sub-overflow-e`, `mul-overflow-e`, `decimal-overflow-e`.

Successful folds that previously depended on the singleton fallback or on nil handling: `float-to-int-v` (`<int>` of a float, covering round-half-to-even at `2.5`/`3.5` and a value past 2^53), `nil-member-v` (`()` as a list member and as a map value), and `08-nil/const1-v` extended with an explicitly `()`-typed constant.

`unimplemented` forms, as future-valid (`-fv`) cases: `08-const/computed-key-fv` (`{[K]: 1}`), `08-const/mapping-spread-fv` (`{...Y}`), `08-const/xml-template-fv` and `09-annotation/xml-template-value-fv` (`xml \`<root>${I}</root>\``).

Non-constant expressions, now reported at the offending subexpression rather than the whole initializer. `subset10/10-conditional-expr/const-e` was one file asserting one diagnostic; it is replaced by `const1-e` … `const6-e` so each rejected operand position of the conditional and nil-conditional operators is asserted separately. `08-xml/namespace19-e` and `namespace20-e` cover a non-constant `xmlns` URI, the path where the message used to be discarded.

Updated goldens: `08-const/19-e` … `22-e` and `09-annotation/14-e` now assert the specific message and the narrower span; `09-annotation/21-v` keeps the runtime-annotation-global path and its comment is corrected to describe why (the reference is not a constant expression, rather than the deleted singleton check).

`make build`, `make vet`, `make lint` and `make test` all pass.

## Security checks

## Documentation

N/A — internal refactor plus more precise diagnostics; no user-facing doc change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compiler diagnostics for invalid constant expressions, including arithmetic overflow, division by zero, non-constant values, and invalid numeric conversions.
  - Constant evaluation now reports errors at the relevant expression with clearer, more specific messages.
  - Added validation coverage for conditional expressions, XML templates, computed mapping keys, and XML namespace values.

- **Tests**
  - Added comprehensive compiler and integration tests for constant evaluation, nil values, numeric conversions, annotation values, and XML behavior.
  - Updated expected outputs to reflect the improved diagnostics.


heshanpadmasiri:
Superseded by #902 — same commit (42fc053), reopened with its head branch on `ballerina-nutcracker/ballerina` so it can sit in a proper GitHub stack (stack #903, on top of #901).

Issue #869, opened 2026-09-03
[New Feature]: Support byte-array literals


heshanpadmasiri:
### Description

Base16 and base64 byte-array literals are parsed but panic during AST construction.

```ballerina
import ballerina/io;

public function main() {
    byte[] bytes = base16 `4869`;
    io:println(bytes);
}
```

Expected output:

```text
[72,105]
```

### Describe your problem(s)

Valid byte-array literals cannot be transformed into AST literal nodes.

### Describe your solution(s)

- Transform `base16` and `base64` literals into AST nodes.
- Decode content according to the specification.
- Validate characters, whitespace, length, and base64 padding.
- Produce the specified inherent and readonly types.
- Preserve literal text for diagnostics.
- Add valid, malformed-input, contextual-type, and runtime corpus tests.

### Related area

Compilation

PR #901, opened 2026-09-08
fix(semantics): give constants the readonly shape type

### LINKED ISSUE #891: [Bug]: constants of list or map type are not readonly-typed
### Description

Spec §8.8 says *"The type of the constant is the intersection of `readonly` and the singleton type containing just the shape of the value named by the constant."* For a constant of a structured type we give it neither the `readonly` intersection nor the singleton — the constant gets the initializer's plain mutable type. `const string` and `const int` are unaffected, because their singleton types are inherently immutable.

The value itself *is* readonly (mutating through an alias panics with `inherent type violation: cannot mutate readonly value`); only its type says otherwise.

### Steps to Reproduce

**1. `is readonly` returns `false`**

```ballerina
import ballerina/io;

const byte[] CL = [1, 2];
const map<int> CM = {"a": 1};

public function main() {
    any a = CL;
    any m = CM;
    io:println(a is readonly);
    io:println(m is readonly);
}
```

| | list | map |
|---|---|---|
| nutcracker | `false` | `false` |
| jBallerina 2201.13.4 | `true` | `true` |

**2. A constant cannot be used where `readonly` is required**

```ballerina
import ballerina/io;

const byte[] CL = [1, 2];

public function main() {
    readonly & byte[] r = CL;
    io:println(r);
}
```

nutcracker:

```
error[SEMANTIC_ERROR]: incompatible type: expected readonly&[int:Unsigned8...], got [int:Unsigned8...]
 --> b.bal:4:27
  |
4 |     readonly & byte[] r = CL;
  |                           ^^
```

jBallerina 2201.13.4: compiles, prints `[1,2]`.

`[int:Unsigned8...]` in that message is `CL`'s type as we have it — a plain mutable `byte[]` with a rest type. One consequence is that a const byte array cannot be passed as an `io:Block`, since that is `readonly & byte[]`.

**3. The type is not the singleton either**

```ballerina
import ballerina/io;

const byte[] CL = [1, 2];

public function main() {
    any a = CL;
    io:println(a is byte[2]);
    io:println(a is readonly & byte[]);
}
```

| | nutcracker | jBallerina 2201.13.4 |
|---|---|---|
| `a is byte[2]` | `false` | `true` |
| `a is readonly & byte[]` | `false` | `true` |

### Analysis

Two independent defects:

- **Compile time.** `resolveConstant` (`semantics/internal/types/type_resolver.go:7883-7887`) sets the constant's symbol type to the initializer's type verbatim — `expectedType := exprTy; t.setSymbolType(symbol, expectedType)` — with no `readonly` intersection and no narrowing to the singleton. The adjacent `// TODO: I am not sure if this is strictly correct given expression type would have changed based on the contextually expected type...` comment is about this. This is what produces the error in repro 2.
- **Runtime.** `values.List` holds readonly-ness twice: a `Type semtypes.SemType` and an `isReadonly bool` (`values/list.go:26`). `checkMutable` (`values/list.go:127`) reads the bool; `x is readonly` reads the semtype (`runtime/internal/exec/non_terminators.go:204` → `values.SemTypeForValue` → `v.Type`). The constant evaluator sets them inconsistently:

  ```go
  // semantics/internal/types/constant_evaluator.go:158
  return values.NewList(expr.GetDeterminedType(), &expr.AtomicType, true /*readonly*/, restFiller, ...)
  ```

  `isReadonly` is hardcoded `true` while `Type` comes from resolution, built with the default `CellMutabilityLimited`. Same at `:140` for maps. This is what produces `false` in repros 1 and 3, and it needs `semtypes.ListMutability(semtypes.CellMutabilityNone)` / the mapping equivalent.

Fixing the compile-time side will move goldens for every existing constant test.

### Affected Version(s)

`main` as of 683266d60.

### Related area

Type checker

### Related issue(s)

Blocks spec-correct const handling for #869 (byte-array literals) — ``const byte[] C = base16 `0102`;`` inherits both defects.

heshanpadmasiri:
## Purpose

Resolves #891.

Spec §8.8: *"The type of the constant is the intersection of `readonly` and the singleton type containing just the shape of the value named by the constant."* For a constant of a structured type we gave it neither the `readonly` intersection nor the singleton — the constant got the initializer's plain mutable type. The value itself was already readonly at runtime (mutating through an alias panics), only its type disagreed.

That showed up as:

- `CL is readonly` → `false` for `const byte[] CL = [1, 2];`
- `readonly & byte[] r = CL;` → `error[SEMANTIC_ERROR]: incompatible type: expected readonly&[int:Unsigned8...], got [int:Unsigned8...]`, so a const byte array could not be passed where `io:Block` is expected
- `CL is byte[2]` → `false`; the type was not the singleton either

This is the first of a stacked set of PRs; the byte-array-literal work (#869) sits on top of it and inherits both defects.

## Approach

Two sites, matching the two defects in the issue.

**`semantics/internal/types/constant_evaluator.go`** — the constant evaluator hardcoded `isReadonly = true` on the value while taking its `SemType` from resolution, which builds it with the default `CellMutabilityLimited`. So `values.List`/`values.Map` carried readonly-ness two ways that disagreed, and `x is readonly` (which reads the semtype) said `false`.

`readonlyListShapeType` / `readonlyMappingShapeType` now build the type from the *evaluated members* with `CellMutabilityNone`, so it is the readonly singleton shape type. They define it as a single list or mapping atom rather than an intersection, because both the BIR deserializer and the inherent type checks on the value require an atomic type.

**`semantics/internal/types/type_resolver.go`** — `resolveConstant` set the symbol type to the initializer's type verbatim (the site of the pre-existing `// TODO: I am not sure if this is strictly correct...`). It now takes the symbol type from the evaluated value, so the symbol type and the value's inherent type are the same type by construction. `exprTy` remains the fallback for a constant whose value could not be evaluated — always an already-reported error.

## Automation tests

Four corpus cases under `corpus/bal/subset8/08-const/`, covering the repros in the issue plus the combinations that the shared shape-type path could break:

- `readonly-list-v.bal` — `const byte[]`, `const int[]`, inferred nested `const`; `is readonly`, `readonly & byte[]` assignment, singleton `is byte[2]`, and that a nested member is readonly too
- `readonly-tuple-v.bal` — `[int, string]` and a rest-typed `[int, int...]`
- `readonly-map-v.bal` — `map<int>`, nested `map<map<int>>`, and that the map constant matches a closed record type
- `readonly-record-v.bal` — a record-typed constant against `R`, `readonly & R`

`make build`, `make vet` and the full test suite pass.

## Security checks

## Documentation

N/A — bug fix bringing constant typing in line with the existing spec, no user-facing doc change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type handling for constant lists, maps, records, and tuples.
  - Constants now consistently preserve readonly characteristics and value shapes, including nested values.
  - Improved compatibility when assigning constants to general and readonly types.
  - Improved resolution of module-level constants with declared or inferred types.

- **Tests**
  - Added coverage for readonly constant lists, maps, records, and tuples.
  - Added expected runtime outputs confirming readonly behavior and type compatibility.


warunalakshitha (approved):
LGTM

PR #902, opened 2026-09-08
refactor(semantics): constant folding

### LINKED ISSUE #896: [Task]: Refactor constant folding logic

heshanpadmasiri:
## Purpose

Resolves #896.

Spec §6.4 gives every constant expression a compile-time shape — there is no runtime semantics for one. So for any structurally valid constant expression, folding must either produce a value or report a compile-time error. The code did neither reliably:

- `evaluateConstantExpression` returned an opaque `error`. `resolveConstant` collapsed every one of them into a single generic `expression is not a constant expression` diagnostic pointed at the whole initializer, `resolveAnnotationAttachments` prefixed them with `cannot evaluate annotation constant expression:`, and `resolveXMLNS` discarded the message entirely.
- When folding failed, the evaluator silently fell back to reading a singleton shape off the *determined type* (`constantSingleShapeValue`) for constant references, string concatenation and string templates. That made a fold failure invisible rather than an error, and made the value depend on whether type resolution happened to produce a singleton.
- Structural non-constness was checked in two unrelated places — `common.ValidateConstantExpr` from the constant analyzer and `resolveXMLNS`, and `errNotConstantExpression` from the evaluator — which is why the analyzer had to re-walk the expression after type resolution had already walked it.

This is the second PR in the stack; it sits on #901.

## Approach

Split the two questions the old `error` was conflating.

**`isConstantExpression` (`constant_evaluator.go`)** — answers only whether an expression is *structurally* a constant expression, returning the first offending subexpression. It reports no diagnostic, because `resolveAnnotationAttachments` needs the answer without one: a non-constant value on a non-`const` annotation is legal and becomes a runtime annotation global. It replaces `common.ValidateConstantExpr` (deleted), so both callers now share one walk.

**`foldConstant` (`constant_evaluator.go`)** — called only on an expression `isConstantExpression` has accepted, so a failure is never structural. It reports its own diagnostic at the *failing subexpression* and returns `false`; the whole evaluator now returns `(values.BalValue, bool)` instead of `(values.BalValue, error)`. A nil value with `ok` is a successful fold of `()`, which is why failure is signalled by the bool alone. It also absorbs the panic recovery that used to live in `evaluateAnnotationValue`.

Failures are now classified by who is at fault:

| kind | cases |
|---|---|
| `semanticError` | what the program got wrong: `1 / 0`, `1 % 0`, integer/decimal overflow, `<int>(1.0 / 0.0)`, a list spread member that is not a list, a missing filler value |
| `unimplemented` | constant forms we do not support yet: computed mapping keys, mapping spread fields, xml templates |
| `internalError` | states type resolution should already have excluded: an unexpected expression node, a non-boolean condition operand, an unknown operator |

`constantSingleShapeValue` and the `errNotConstantExpression` sentinel are gone. Dropping the singleton fallback also means `resolveConstant` no longer needs the initializer type as a backstop — the symbol type is always `values.SemTypeForValue(value)` — and `constantAnalyzer` no longer re-validates constness, since reaching it means type resolution found nothing to report.

Two smaller consequences of pointing diagnostics at the offending node: `evaluateConstantReference` and the mapping-key helper now take a position rather than a `SemType`, and `constantConversionError` includes the offending float value (`float value Infinity cannot be converted to int`) now that the message is no longer swallowed.

## Automation tests

Every changed diagnostic is pinned by an integration golden. The new cases cover each of the three failure classes and the fold paths the removed fallback used to mask.

Semantic errors — `corpus/bal/subset8/08-const/`: `div-zero-e`, `mod-zero-e`, `add-overflow-e`, `sub-overflow-e`, `mul-overflow-e`, `decimal-overflow-e`.

Successful folds that previously depended on the singleton fallback or on nil handling: `float-to-int-v` (`<int>` of a float, covering round-half-to-even at `2.5`/`3.5` and a value past 2^53), `nil-member-v` (`()` as a list member and as a map value), and `08-nil/const1-v` extended with an explicitly `()`-typed constant.

`unimplemented` forms, as future-valid (`-fv`) cases: `08-const/computed-key-fv` (`{[K]: 1}`), `08-const/mapping-spread-fv` (`{...Y}`), `08-const/xml-template-fv` and `09-annotation/xml-template-value-fv` (`xml \`<root>${I}</root>\``).

Non-constant expressions, now reported at the offending subexpression rather than the whole initializer. `subset10/10-conditional-expr/const-e` was one file asserting one diagnostic; it is replaced by `const1-e` … `const6-e` so each rejected operand position of the conditional and nil-conditional operators is asserted separately. `08-xml/namespace19-e` and `namespace20-e` cover a non-constant `xmlns` URI, the path where the message used to be discarded.

Updated goldens: `08-const/19-e` … `22-e` and `09-annotation/14-e` now assert the specific message and the narrower span; `09-annotation/21-v` keeps the runtime-annotation-global path and its comment is corrected to describe why (the reference is not a constant expression, rather than the deleted singleton check).

`make build`, `make vet`, `make lint` and `make test` all pass.

## Security checks

## Documentation

N/A — internal refactor plus more precise diagnostics; no user-facing doc change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compile-time validation of constant expressions, including conditional operators, arithmetic overflow, division by zero, invalid conversions, and non-constant operands.
  - Added clearer diagnostics for invalid constant values and unsupported constant constructs.
  - Improved constant-expression handling in annotations and XML namespace declarations.

- **Tests**
  - Expanded coverage for constant evaluation, nil values, numeric conversions, XML templates, mapping constructs, and conditional operators.
  - Updated expected diagnostics and runtime output for affected scenarios.


corpus/integration/subset8/08-const/mapping-spread-fv.txtar
```diff
@@ -0,0 +1,7 @@
+-- stdout --
+-- stderr --
+fatal[UNIMPLEMENTED_ERROR]: mapping constructor spread field not implemented
```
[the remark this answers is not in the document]
heshanpadmasiri: [AI RESPONSE] failed to reproduce, can you present a test case for this

`TestIntegration/subset8/08-const/mapping-spread-fv.bal` passes as written. The spread field is rejected in `nodebuilder/node_builder.go:2252` with `mapping constructor spread field not implemented`, before `evaluateMappingConstructor` ever sees it, so the fixture text is correct.

semantics/internal/types/constant_evaluator.go
```diff
@@ -29,31 +29,131 @@ import (
...
+		return nil, true
+	case *ast.BLangTemplateExpr:
+		for _, insertion := range e.Insertions {
+			if offender, ok := isConstantExpression(t, insertion); !ok {
+				return offender, false
+			}
+		}
+		return nil, true
+	case *ast.BLangAnnotAccessExpr:
+		return isConstantExpression(t, e.Expr)
```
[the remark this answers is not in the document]
heshanpadmasiri: [AI RESPONSE] failed to reproduce, can you present a test case for this

The `*ast.BLangAnnotAccessExpr` case in `isConstantExpression` recurses into `e.Expr`, which for annotation access is a typedesc-valued expression. A type reference resolves to a type symbol, never a `*model.ConstantValueSymbol`, so the `*ast.BLangVarRef` arm returns false and the annot-access arm can never report true. Tried `const boolean? M = Person.@marker;` (reports `expression is not a constant expression`, never reaching `evaluate`), plus the annotation-value and record-field-default paths; annotation values only accept mapping constructors, and the nested field value hits the same var-ref rejection. Per AGENTS.md, a scenario with no corpus test is one that cannot happen, so there is nothing to fix here.

semantics/internal/types/constant_evaluator.go
```diff
@@ -213,165 +291,180 @@ func (e *constantExpressionEvaluator) evaluateListConstructor(expr *ast.BLangLis
...
 }
 
-func (e *constantExpressionEvaluator) evaluateBinaryExpression(expr *ast.BLangBinaryExpr) (values.BalValue, error) {
-	lhs, err := e.evaluate(expr.LhsExpr)
-	if err != nil {
-		return nil, err
+func (e *constantExpressionEvaluator) evaluateBinaryExpression(expr *ast.BLangBinaryExpr) (values.BalValue, bool) {
+	lhs, ok := e.evaluate(expr.LhsExpr)
+	if !ok {
+		return nil, false
```
warunalakshitha: IMO these can be covered using simple tests like
```ballerina
const A = (1 / 0) + 1; // @error
```

semantics/internal/types/constant_evaluator.go
```diff
@@ -213,165 +291,180 @@ func (e *constantExpressionEvaluator) evaluateListConstructor(expr *ast.BLangLis
...
 	default:
 		// Other operators require both operands.
 	}
 
-	rhs, err := e.evaluate(expr.RhsExpr)
-	if err != nil {
-		return nil, err
+	rhs, ok := e.evaluate(expr.RhsExpr)
+	if !ok {
+		return nil, false
```
warunalakshitha:
```ballerina
const A = 1 + (1 / 0); // @error
```

semantics/internal/types/constant_evaluator.go
```diff
@@ -555,17 +648,17 @@ func constantExactEqual(lhs, rhs values.BalValue) bool {
...
 }
 
-func (e *constantExpressionEvaluator) evaluateTypeConversion(expr *ast.BLangTypeConversionExpr) (values.BalValue, error) {
-	value, err := e.evaluate(expr.Expression)
-	if err != nil {
-		return nil, err
+func (e *constantExpressionEvaluator) evaluateTypeConversion(expr *ast.BLangTypeConversionExpr) (values.BalValue, bool) {
+	value, ok := e.evaluate(expr.Expression)
+	if !ok {
+		return nil, false
```
warunalakshitha:
```ballerina
const A = <int>(1 / 0); // @error
``` 
will cover this

semantics/internal/types/constant_evaluator.go
```diff
@@ -589,31 +682,32 @@ func constantConversionError(value values.BalValue, targetType semtypes.SemType)
...
 	}
 	var result strings.Builder
 	for i, insertion := range expr.Insertions {
 		result.WriteString(expr.Strings[i])
-		value, err := e.evaluate(insertion)
-		if err != nil {
-			return nil, err
+		value, ok := e.evaluate(insertion)
+		if !ok {
+			return nil, false
```
warunalakshitha:
```ballerina
const string A = string `x${1 / 0}`; // @error
```

semantics/internal/types/constant_evaluator.go
```diff
@@ -213,165 +291,180 @@ func (e *constantExpressionEvaluator) evaluateListConstructor(expr *ast.BLangLis
...
+		if err != nil {
+			return e.semanticFailure(err.Error(), expr.GetPosition())
+		}
+		return value, true
 	case model.OperatorKind_AND:
-		return lhs.(bool) && rhs.(bool), nil
+		return lhs.(bool) && rhs.(bool), true
 	case model.OperatorKind_OR:
-		return lhs.(bool) || rhs.(bool), nil
+		return lhs.(bool) || rhs.(bool), true
```
warunalakshitha, line 418: We can cover this with
```ballerina
const boolean A = false || true; // @output true
```

semantics/internal/types/constant_evaluator.go
```diff
@@ -213,165 +291,180 @@ func (e *constantExpressionEvaluator) evaluateListConstructor(expr *ast.BLangLis
...
 }
 
-func (e *constantExpressionEvaluator) evaluateNilConditionalExpression(expr *ast.BLangNilConditionalExpr) (values.BalValue, error) {
-	lhs, err := e.evaluate(expr.LhsExpr)
-	if err != nil {
-		return nil, err
+func (e *constantExpressionEvaluator) evaluateNilConditionalExpression(expr *ast.BLangNilConditionalExpr) (values.BalValue, bool) {
+	lhs, ok := e.evaluate(expr.LhsExpr)
+	if !ok {
+		return nil, false
```
warunalakshitha:
```ballerina
const A = (1 / 0) ?: 2; // @error
```

semantics/internal/types/constant_evaluator.go
```diff
@@ -213,165 +291,180 @@ func (e *constantExpressionEvaluator) evaluateListConstructor(expr *ast.BLangLis
...
-	condition, err := e.evaluate(expr.Condition)
-	if err != nil {
-		return nil, err
-	}
-	conditionValue, ok := condition.(bool)
+func (e *constantExpressionEvaluator) evaluateTernaryExpression(expr *ast.BLangTernaryExpr) (values.BalValue, bool) {
+	condition, ok := e.evaluate(expr.Condition)
 	if !ok {
-		return nil, fmt.Errorf("constant conditional operand has type %T", condition)
+		return nil, false
```
warunalakshitha: we can cover this with 

```ballerina
const A = (1 / 0) ? 1 : 2; // @error
```

semantics/internal/types/constant_evaluator.go
```diff
@@ -213,165 +291,180 @@ func (e *constantExpressionEvaluator) evaluateListConstructor(expr *ast.BLangLis
 	for i := len(initial); i < expr.AtomicType.FixedLength(); i++ {
 		filler, ok := values.FillerFactoryFor(e.resolver.typeContext(), expr.AtomicType.MemberAtInnerVal(i))
 		if !ok {
-			return nil, fmt.Errorf("constant list member %d has no filler value", i)
+			return e.semanticFailure(fmt.Sprintf("constant list member %d has no filler value", i), expr.GetPosition())
```
warunalakshitha, line 294:
```ballerina
type R record {| int x; |};
const [R, R] A = []; // @error
```

warunalakshitha:
We can add following which will cover most of not coverd cases
```ballerina
const A = [string `x${1 + <int>(-((1 / 0) ?: 2))}`]; // @error
```

warunalakshitha:
# Review report — #902 refactor(semantics): constant folding

> [!NOTE]

## Summary
**1 Important finding — fix first**, and it's process-only: `INTENT.md` didn't exist on this branch, so this run drafted one from the (unusually thorough) PR description. This is an excellent refactor, and one parity check that initially looked like a real problem turned out to be the opposite: the float-to-int constant-folding tests (`<int>1.0`, round-half-to-even at 2.5/3.5) are rejected outright by the locally-installed jBallerina (2201.13.2) as "not a constant expression" — but the official language spec (§6.4, fetched directly from `ballerina-platform/ballerina-spec`) explicitly lists `type-cast-expr` in the `const-expr` grammar. jBallerina itself has a spec-compliance gap here; Nutcracker is correct. Separately, while probing a coverage gap I found a genuine, serious bug — but confirmed it's pre-existing and unrelated to this PR (see Flagged), so it's reported to you directly rather than entered as a finding against this PR.

## Must fix (Important)
|---|---|---|---|---|
| 1 | Process | `docs/changes/v0.7.0/902/INTENT.md` | Missing on this branch | This run drafted it summarizing the PR's own already-thorough description. Per the pre-pr-review process, write `INTENT.md` (can closely follow the PR body) before merging — an agent-drafted `INTENT.md` is itself an Important finding even when the source material is this good. |

None — every claim below was confirmed by direct repro, source trace, or the official spec text, not left as a suspicion.

| Kind | Detail |
|---|---|
| **Pre-existing crash, unrelated to this PR — please triage separately** | `const V = 1; const A = [...V];` (spreading a non-list constant into a constant list) crashes with `panic: interface conversion: semtypes.allOrNothingSubtype is not semtypes.bdd: missing method canonicalKey` in `semtypes.listProjInnerVal`, during *type resolution* of the list constructor — before `constant_evaluator.go`'s own graceful "constant list spread member has type %T" semantic-error path is ever reached. Confirmed this is **not** introduced or worsened by this PR: built the exact base commit (`3d9b25672`) and reproduced the identical panic there. The equivalent *non-constant* case (`int[] a = [...v];` with `v` a non-list local) correctly produces a graceful `SEMANTIC_ERROR`, not a panic — so this is specific to the constant-list-constructor type-resolution path. Worth its own issue; out of scope for this PR. |
| PR merge status | `gh pr view 902` reports `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING` — `main` has moved since this branch was created (it's the second PR in a stack, sitting on #901). Needs a rebase before merging; not a code-correctness concern. |
| jBallerina divergence confirmed as *not a bug* | `float-to-int-v.bal` (`<int>` of a float literal, including round-half-to-even at 2.5→2 and 3.5→4) is rejected by jBallerina 2201.13.2 as "not a constant expression" for every case, including the simplest (`<int>1.0`). Verified this isn't a local environment issue (ruled out with a minimal control case). Checked the official spec grammar directly: `const-expr := ... | type-cast-expr | ...` (§6.4) — type-cast-expr is explicitly a valid constant-expression form. This is a jBallerina gap, not a Nutcracker one; per this PR's own description, this exact fold path already worked before this PR too (via the old, buggy singleton-fallback mechanism this PR replaces) — this PR makes it correct and robust (proper round-half-to-even) rather than introducing it. |
| Design note (not a defect) | `isConstantExpression` returns only the *first* offending subexpression per call, unlike the deleted `ValidateConstantExpr`'s callback which could report every non-const position in one expression. Explicitly stated in the PR description as the intended design (paired with `foldConstant` reporting exactly one diagnostic at the failing subexpression) — a deliberate simplification, not an oversight. |
| Unnecessary / large comments | None — every new comment is a non-obvious "why" (the fold-vs-structural split, the panic-recovery rationale, the fault-classification table mirrored in code comments). |
| New dependency | none |

## Fixed during review
None — this is a review of an already-open external PR (#902), not a branch this session can push to, so findings are reported rather than fixed.

## Nits (suggested, not applied)
None found.

## Evidence
| Pass | Result |
|---|---|
| Branch / scope | PR902 head `0deb337d4` vs merge-base `3d9b25672` (confirmed a real ancestor via `git merge-base --is-ancestor`, and *is* `main` at the point this branch was created — despite the PR currently showing `CONFLICTING` against current `main`, see Flagged); 1 commit, 65 files (+1,183 / −391) |
| INTENT.md | missing — drafted by this run (from an already-thorough PR description), flagged Important |
| Commits | Conventional ✔ (`refactor(semantics): fold constants to a value or a diagnostic`) · no Co-authored-by ✔ · no AI author ✔ (single author/committer, human email) |
| Lint, fmt, license headers | `gofmt -l -s` clean; no new `.go` files (license headers N/A); `go vet` clean; `staticcheck@latest` on `semantics/internal/types/...`, `semantics/internal/common/...`, `semantics/internal/analysis/...` found 2 unused symbols, both confirmed pre-existing and untouched by this diff (0 hits in `git diff`) — out of scope |
| Dead code | none introduced; `common.ValidateConstantExpr` (50 lines) correctly deleted as the walker it duplicated (`isConstantExpression`) took over both its callers |
| `go test` + corpus | `go build ./...`, `go vet ./...` clean; `go test ./semantics/...` passes; full `go test ./corpus/...` passes |
| Reuse before reimplement (mechanical pass) | ~13 new/rewritten functions in `constant_evaluator.go` (mostly renamed/refactored evaluator methods, same shape as before with a changed return signature) — each exactly one definition repo-wide. `isConstantExpression` correctly replaces (not duplicates) the deleted `common.ValidateConstantExpr`, unifying what were two separate structural-constness checks (analyzer + evaluator) into one |
| Impossible-condition scan | Sampled representative uncovered branches across all three of the PR's own failure categories: the `foldConstant` panic-recovery `defer` (122-124) and every `internalFailure`-classified branch checked (`internalFailure` itself at 141-143 uncovered — expected, it's the "should never happen" case; the mapping-key/unary-op/ternary-condition internal errors at 253-255, 268-269, 340-341, 351-352 match the PR's own stated examples "an unexpected expression node, a non-boolean condition operand, an unknown operator" verbatim) are all correctly exempt per the skill's own internal-error rule. One hypothesis was formed and *disproven* by direct repro rather than assumed: line 82 (`isConstantExpression`'s mapping-spread-field skip) looked like it should be exercised by `mapping-spread-fv.bal`, but tracing the actual error showed that test is rejected at `nodebuilder/node_builder.go` (AST-construction stage, "mapping constructor spread field not implemented") — two stages earlier, for a general limitation unrelated to constants. Not every one of the ~30 uncovered single-line ranges was individually traced (patch coverage is exactly at the 80% target, not below it); the ones sampled all fit the same well-disclosed pattern. |
| Design | Cleanly separates "is this structurally a constant expression" (no diagnostic) from "does folding it succeed" (always a value or a diagnostic) — exactly the conflation the PR set out to fix, confirmed by reading both functions in full |
| Patch coverage | 80% of 203 changed statement lines — meets target exactly. `constant_evaluator.go` alone is 77%; see Impossible-condition scan for the sampling done on its uncovered ranges |
| jBallerina parity | 5 spot-checks: `div-zero-e` matches exactly (line, category); `nil-member-v` matches exactly (stdout); `const1-e` (conditional non-const operand) matches on line/message, narrower underline span than jBallerina's (matches the PR's own stated design); `computed-key-fv` confirmed as `UNIMPLEMENTED_ERROR` category as intended; `float-to-int-v` diverges from jBallerina but confirmed *not a bug* against the official spec text (see Flagged) |
| Verification pass | Every claim in this report is backed by a repro, a source trace, a base-commit comparison, or the fetched spec text — including one hypothesis (mapping-spread) that was checked and dropped when the repro contradicted it |
| Commands | `go build ./...`, `go vet ./...`, `gofmt -l -s`, `go run honnef.co/go/tools/cmd/staticcheck@latest ./semantics/...`, `go test ./semantics/...`, `go test ./corpus/...`, `patch_coverage.py`, `go run ./cli/cmd run` + `bal run` on 5 spot-check programs plus 2 crash-isolation repros, a base-commit worktree build to confirm the crash predates this PR, a fetch of `raw.githubusercontent.com/ballerina-platform/ballerina-spec/master/lang/spec.html` for the `const-expr` grammar |

|---|---|---|---|
| semantics/internal/analysis/semantic_analyzer.go | 1 | 100% | — |
| semantics/internal/common/common.go | 0 | n/a (no statements) | — |
| semantics/internal/types/constant_evaluator.go | 179 | 77% | 82, 97, 122-124, 141-143, 150, 156, 176, 194, 209, 253-255, 268-269, 277, 282, 294, 300, 308, 340-341, 347-348, 351-352, 363, 374, 380, 388, 399, 418, 450, 467, 654, 699-700, 707 |
| semantics/internal/types/type_resolver.go | 23 | 100% | — |

<details>

```
issue: 902
reviewed_diff: sha256:9416cacd55c7f5a2d1426d98baf4fdb5ded4f939db78f8d36f9711d65e156970
skill_version: sha256:c5b1e3d5b07608dc303a686f000dc18de61f20f35748258d00edc2007bdda996
date: 2026-09-24T13:39:30Z
patch_coverage: 80
important: 1
verify: 0
verdict: "1 Important finding — fix first"
```
</details>

PR #905, opened 2026-09-08
fix(semantics): enforce readonly record field immutability

### LINKED ISSUE #898: [Bug]: readonly record fields can be reassigned — no compile error and no runtime panic
### Description

Nutcracker does not enforce the immutability of `readonly` fields in record values. Both the compile-time check and the runtime check are missing.

**Case 1 — assigning to a `readonly` field through the declaring type should be a compile error**

```ballerina
import ballerina/io;

type MR record {|
  readonly int value;
  int otherValue;
|};

public function main() {
  MR m = { value: 5, otherValue: 5 };
  m.otherValue = 10;
  io:println(m);
  m.value = 10; // expected: compile error
  io:println(m);
}
```

Expected: a compile error on `m.value = 10` (cannot update `readonly` field `value`).
Actual: compiles and runs.

```
{"value":5,"otherValue":10}
{"value":10,"otherValue":10}
```

**Case 2 — assigning through a wider type should panic at runtime**

```ballerina
import ballerina/io;

type MR record {| readonly int value; int otherValue; |};
type M  record {| int value; int otherValue; |};

public function main() {
  MR mr = { value: 5, otherValue: 5 };
  M m = mr;
  m.otherValue = 10;
  io:println(m);
  m.value = 10; // expected: panic
  io:println(m);
}
```

Here `MR` is a subtype of `M`, so the assignment cannot be caught statically; the update must panic at runtime because the underlying value's `value` field is `readonly`.

Expected: panic on `m.value = 10`.
Actual: the field is silently mutated.

```
{"value":5,"otherValue":10}
{"value":10,"otherValue":10}
```

### Possible cause

The `readonly` field qualifier appears to be parsed and accepted but never carried into the checks that guard field assignment — neither into the semantic check on member-access assignment, nor onto the runtime mapping value as per-field immutability that the field-set path consults.

### Related area

Semantics

### Related issue(s)

- Sub-issue of #856
- Blocks #876

heshanpadmasiri:
## Purpose

`readonly` record fields were parsed and accepted but the qualifier never reached either
check that guards a field assignment, so a `readonly` field could be reassigned with no
compile error and no runtime panic.

Fixes #898

Stacked on #904 — review only the top commits; the base commit belongs to that PR.

## Approach

The `readonly` bit already exists in the semtype representation as the *cell mutability* of
a mapping field's cell. Both checks are made to consult it, through one shared exposure:

- **`refactor(semtypes): expose mapping field cells and cell mutability`** — export
  `cellInnerVal` and `cellMut`, add `MappingAtomicType.FieldCell` (mirroring the existing
  `ListAtomicType.MemberAt`) and `MappingAlternative.Atomic`, so callers outside the package
  can read a mapping field's cell mutability. No behaviour change.
- **`fix(runtime): panic when mutating a readonly mapping field`** — the inherent type check
  on a mapping store widens from *member type only* to *member type and cell mutability*, so
  a `readonly` field mutated through a widened static type (issue case 2) panics instead of
  silently mutating. A cell that cannot hold a value at all — a closed record's rest cell —
  is deliberately not treated as readonly, so its existing diagnostic is unchanged.
- **`fix(semantics): reject assignment to a readonly mapping field`** — an lvalue whose cell
  is `CellMutabilityNone` in *every* mapping alternative is rejected (issue case 1). One
  mechanism covers `readonly` record fields, `T & readonly`, and `readonly & map<T>`. Reads
  are untouched: the check is gated on `IsLexpr`.

## Checklist

## Tests

New corpus tests under `corpus/bal/subset10/10-record/` and `10-map/`:

- **panic (`-p`)** — `readonlyfield1..6,9` cover mutation through a widened static type for
  the plain record case, intersections and `map<T>`; `readonlyfield11-p` covers *removing* a
  readonly optional field.
- **error (`-e`)** — `readonlyfield6..8,11,13`, `memberlvalue1-e` and `10-map/readonly2-e`
  cover the compile-time rejection and the untouched not-a-valid-key diagnostic.
- **valid (`-v`)** — `readonlyfield7,9,10,12` pin down what must keep working: reads of
  readonly fields, and writes to the mutable fields of a record that has readonly ones.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive readonly-field and readonly-map handling for record and mapping values.
  - Prevents modification or removal of readonly and required fields, with clear compile-time or runtime errors.
  - Supports readonly behavior across direct access, indexed access, unions, nested records, optional fields, and map conversions.

- **Bug Fixes**
  - Improved detection and reporting of invalid mutations involving readonly values and undeclared fields.

- **Tests**
  - Added extensive compiler and runtime coverage validating readonly access, mutation, removal, diagnostics, and expected output.


warunalakshitha (approved):
LGTM

PR #2, opened 2025-07-22
Set up initial project structure

snelusha:
This pull request sets up the initial structure for the `ballerina-lang-go` project, including configuration for CI, project metadata, and a basic implementation. The most important changes include adding a GitHub Actions workflow for Go CI, creating a `README.md` with project goals and usage instructions, initializing the Go module, and adding a basic `main.go` file with a "hello world" program.

### CI Setup:
* [`.github/workflows/go-ci.yml`](diffhunk://#diff-14e19ae558e62d3b5deed86deba28975aafaa2393584a9bb8591683f656406b0R1-R32): Added a GitHub Actions workflow to run tests and check Go code formatting using `gofmt`.

### Documentation:
* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L2-R24): Added project goals, implementation plan, usage instructions, and testing guidelines for the `ballerina-lang-go` project.

### Project Initialization:
* [`go.mod`](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R1-R3): Initialized the Go module for the project with the module name `ballerina-lang-go` and Go version `1.24.4`.

### Code Implementation:
* [`main.go`](diffhunk://#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R1-R25): Added a basic "hello world" program with an Apache 2.0 license header.

PR #907, opened 2026-09-08
fix(semantics): construct readonly constant fillers

### LINKED ISSUE #894: [Bug]: constant folding for list constructors does not handle filler values correctly
### Description

Constant folding of a list constructor is supposed to fill the unspecified members of a fixed-length list type with filler values. What it actually does depends on the member type, and is wrong in two separate ways.

**1. A filler is not produced when the member type is a record; the declaration is rejected.**

```ballerina
type R record {| int x = 5; |};
type RA R[2];

const RA F = [];
```

```
error[SEMANTIC_ERROR]: expression is not a constant expression
 --> a.bal:4:14
  |
4 | const RA F = [];
  |              ^^
```

jBallerina 2201.13.4 compiles the equivalent declaration and prints `[{"x":5},{"x":5}]`.

The diagnostic is wrong as well. The failure is detected as "constant list member 0 has no filler value", but that text never reaches the user, so nothing in the message mentions filler values or the member type. The same substitution happens for a member type that genuinely has no filler value: `const U V = [1];` with `type U (int|string)[2];` also reports `expression is not a constant expression`, while the non-const form reports the real reason:

```
error[SEMANTIC_ERROR]: missing required member at index 1: type 'int|string' has no filler value
 --> b.bal:4:11
  |
4 |     U v = [1];
  |           ^^^
```

**2. When a filler is produced for a list or mapping member type, it is a writable value, so the constant can be modified while the program runs.**

```ballerina
import ballerina/io;

type L2 int[2][2];

const L2 X = [];

public function main() {
    int[2] slot = X[0];
    slot[0] = 7;
    io:println(X);        // [[7,0],[0,0]]
}
```

Expected: the write fails with `cannot mutate readonly value` and `X` stays `[[0,0],[0,0]]`. That is what happens when the members are written out in the source instead of being filled, for the very same constant type:

```ballerina
const L2 P = [[1, 2], [3, 4]];

int[2] slot = P[0];
slot[0] = 7;   // panics: inherent type violation: cannot mutate readonly value
```

Writing to the constant itself is still blocked — `int[2][2] o = X; o[0] = [9, 9];` panics correctly. Only the contents of a filled member are open.

Seen from the type side, the same fact shows up as the constant not being `readonly`. Measured per member type, with the member left unspecified:

| member type | filler produced | value | `is readonly` | contents writable |
|---|---|---|---|---|
| `int` | yes | `0` | `true` | no |
| `string` | yes | `""` | `true` | no |
| `int[2]` | yes | `[0,0]` | `false` | **yes** |
| `map<int>` | yes | `{}` | `false` | **yes** |
| `record {\| \|}` | yes | `{}` | `true` | no |
| `record {\| int x = 5; \|}` | **no — rejected** | — | — | — |
| `int\|string` | no — rejected (correct, the type has no filler value) | — | — | — |

Scalar members are the only ones that come out right, which is what every existing corpus case uses (`const int[3]`-style), and is why neither of these was caught.

### Steps to Reproduce

1. Run the `const RA F = [];` snippet above — expected `[{"x":5},{"x":5}]`, actual a semantic error that does not mention filler values.
2. Run the `const L2 X = [];` program above — expected a panic on `slot[0] = 7`, actual `[[7,0],[0,0]]`.

### Affected Version(s)

`main` as of 1982e7096.

Observation 2 predates #891: 683266d60 prints `[[7,0],[0,0]]` for the same program. Before #891 every structured constant reported `is readonly == false`, so only the write was observable there; #891 made explicitly written members `readonly` and left filled members behind, which is why the two now differ.

### Possible diagnosis

For observation 2, the filler comes from `values.FillerFactoryFor` → `listFillerFactory` / `mapFillerFactory` (`values/values.go:100`), which builds the value from the member type as written and takes its readonly flag from `IsSubtype(cx, memberTy, ValReadonly)` — `false` for `int[2]` and `map<int>`, `true` for `int`, `string` and `record {| |}`, which matches the table exactly. `evaluateListConstructor` (`semantics/internal/types/constant_evaluator.go:184-190`) appends what the factory returns unchanged, whereas the members it evaluates itself are constructed readonly, hence the split between written and filled members. A constant's filled members presumably need to be built readonly, the same way the evaluated ones are.

For observation 1, a record type's filler value is not reachable at constant folding time: the defaults become default-value functions only during desugar (`desugar/desugar.go:1347`, stage 9), while folding runs at stage 4, and the semtype the filler factory works from carries field types but not defaults. Note the non-const path fails the same way — `RA v = [];` reports `missing required member at index 0: type '{| x: int, never... |}' has no filler value`, with the default already absent from the printed type — so that part is not specific to constants.

For the misleading message, `evaluateListConstructor` returns a plain `fmt.Errorf("constant list member %d has no filler value", i)` (`constant_evaluator.go:186`) and `resolveConstant` replaces every error that is not `errNotConstantExpression` with `"expression is not a constant expression"` (`semantics/internal/types/type_resolver.go:7879`), discarding the text.

### Related area

Semantics

### Related issue(s)

#891 (constant type is the readonly shape of its value) made observation 2 visible in the type. #892 (const rejected when its inline type descriptor has an array length or field default) blocks an inline `int[3]` from reaching this path at all.

The mapping-constructor counterpart is #895 (mapping constructors in constant expressions do not use record field defaults).

heshanpadmasiri:
## Purpose

Closes #894.

Constant list constructors used runtime filler values directly. Structured fillers therefore remained mutable inside otherwise readonly constants, and missing fillers were reported as generic constant-expression failures.

This is the third PR in the stack; it sits on #902.

## Approach

- Validate filler eligibility through the semantic filler descriptor.
- Recursively rebuild list and mapping fillers as readonly constant values with singleton shape types.
- Preserve fresh mutable fillers for non-constant list construction.
- Construct XML fillers as typed empty XML text values.
- Report unsupported and missing constant fillers through the diagnostic paths introduced by the preceding PR.
- Update stale Go APIs that blocked repository linting under Go 1.26.

## Checklist

## Tests

- `go test ./ast/... ./semantics/... ./desugar/... ./bir/... ./corpus/...`
- `go test ./common/...`
- `go test ./nodebuilder/...`
- Full pre-commit build and lint checks

## Summary by CodeRabbit

- **New Features**
  - Constant evaluation now supports nested scalar, array, tuple, and map filler values with readonly semantics.
  - Empty fixed-size arrays and composite constants receive correctly initialized values.
  - Array growth fills skipped positions with default values.

- **Bug Fixes**
  - Readonly values now consistently reject mutation attempts.
  - Nested constant values preserve independent mutable copies where appropriate.
  - Filler validation reports clearer errors for missing or unsupported values.

- **Tests**
  - Added coverage for constant evaluation, readonly behavior, array gaps, XML handling, and expected compiler diagnostics.


warunalakshitha:
# Review report — #907 fix(semantics): construct readonly constant fillers

> [!NOTE]

## Summary
**2 Important findings — fix first**, neither of which needs new code logic. `INTENT.md` is missing (drafted here). The 67% patch coverage number needs a closer look, not new tests: every uncovered line was verified line-by-line — each is either a pre-existing branch that was only relocated (not new behavior), an exhaustive `default:` internal error, or an object/table/stream filler path confirmed unreachable from any `.bal` program today (table type descriptors aren't implemented in the frontend at all). That last one just needs a tracking issue and a Deferred entry, not a test that can't be written. Everything else is clean: correct design, well-tested (10 new corpus cases, including two mutation-panic regressions and a real bug fix for XML filler typing), and 2/2 jBallerina parity spot-checks match exactly.

## Must fix (Important)
|---|---|---|---|---|
| 1 | Process | `docs/changes/v0.7.0/907/INTENT.md` | Missing on this branch | This run drafted it from the diff and PR description (see file). Per the pre-pr-review process, rewrite Problem/Approach in your own words before opening the PR — an agent-drafted INTENT.md is itself an Important finding. |
| 2 | Tests (coverage / Deferred) | `semantics/internal/types/constant_evaluator.go:311-313` (and the same-cause propagation at 332-333, 341-346) | Patch coverage is 67%, below the 80% target, but not because of a missing-but-writable test: `constantFillerValue`'s "not implemented" branch (and its propagation through nested list/mapping fillers) is only reachable for an object/table/stream member type in a constant. Verified experimentally: table type descriptors aren't implemented in the frontend (`transformTableTypeDescriptor unimplemented`), and no plain object shape both qualifies for `semtypes.ObjectFiller` and can appear as a constant's member type. This is exactly the pattern `INTENT.md`'s Deferred section exists for — but I found no existing tracking issue for it. | Open an issue for object/table/stream constant-filler support and add it to `INTENT.md`'s Deferred table with this reasoning; no test needs to be written today. |

None.

| Kind | Detail |
|---|---|
| PR description mismatch | The PR body's Approach list includes "Update stale Go APIs that blocked repository linting under Go 1.26," but this diff touches only `semantics/internal/types/constant_evaluator.go` and `values/values.go` — no Go-API/lint-related change is present. Likely copied forward from an earlier point in the stack and not trimmed. Harmless, but worth cleaning up before merge so the description matches the diff. |
| Coverage detail (not a gap) | `constant_evaluator.go:354` (`default:` internal error in `readonlyFillerValue`'s type switch) is an exhaustive case: `values.FillerValue` can only ever return the six types already handled (scalars, `*values.List`, `*values.Map`, `values.XMLValue`) once object/table/stream are rejected upstream — genuinely unreachable, correctly an internal error, no test needed. |
| Coverage detail (not a gap) | `constant_evaluator.go:365, 377-378` (`readonlyListValue`/`readonlyMappingValue`'s `err != nil` branches) are pre-existing code relocated from the old inline `evaluateListConstructor`/`evaluateMappingConstructor` bodies, not new behavior introduced by this PR — same "moved, not added" situation as the PR977 review found in `runner.go`. |
| New dependency | none |
| Corpus fixture changes | 10 new `.bal` + 10 new `.txtar` (1:1), all additions — no existing golden was changed |
| Benchmark | not run — this PR touches constant-folding at compile time, not the interpreter's hot execution loop; `corpus/bench` has no relevant case |

## Fixed during review
None — this is a review of an already-open external PR (#907), not a branch this session can push to, so findings are reported rather than fixed.

## Evidence
| Pass | Result |
|---|---|
| Branch / scope | PR907 head `e1f5765b2` vs its stacked base `0deb337d4` (`refactor/const-folding-diagnostics`, not `main` — this is the third PR in a stack, sits on #902); 1 commit, 54 files (+1,936 / −12) |
| INTENT.md | missing — drafted by this run, flagged Important |
| Commits | Conventional ✔ (`fix(semantics): construct readonly constant fillers`) · no Co-authored-by ✔ · no AI author ✔ (single author/committer, human email) |
| Lint, fmt, license headers | `gofmt -l -s` clean; no new `.go` files (license headers N/A); `go vet` clean; `staticcheck@latest` on affected packages found 4 unused symbols, all pre-existing in files this PR doesn't touch — none in scope |
| Dead code | none introduced |
| `go test` + corpus | `go build ./...`, `go vet ./...` clean; `go test` across `semantics`, `values`, `ast`, `bir`, `desugar`, `nodebuilder`, `common` passes; full `go test ./corpus/...` (integration + per-stage goldens) passes |
| Reuse before reimplement (mechanical pass) | New function names (`constantFillerValue`, `readonlyFillerValue`, `readonlyListValue`, `readonlyMappingValue`) each have exactly one definition repo-wide — no duplicate. The diagnostic helpers (`semanticFailure`/`unimplementedFailure`/`internalFailure`) and `values.FillerValue`/`semtypes.FillerValue` are all pre-existing and correctly reused, not reimplemented |
| Impossible-condition scan | `readonlyFillerValue`'s `default:` case is exhaustive-and-unreachable by construction (see Flagged); the object/table/stream "not implemented" path is unreachable today but for a different, real reason (unimplemented frontend feature, not a logic bug) — distinguished these two explicitly rather than lumping them together |
| Design | Correct separation: eligibility (`semtypes.FillerValue`) checked before construction (`values.FillerValue`), constant rebuild (`readonlyFillerValue`) is a pure function over the runtime filler value (no mutation of its input, so no risk of corrupting a shared factory-produced value), and the non-constant fresh-filler-per-slot path is untouched and has its own regression test (`fixed-independence-v.bal`) |
| Bug found (already fixed, not by me) | `values/values.go`'s XML filler used to build `&XMLText{}` directly, leaving `semType` at its zero value instead of `semtypes.XMLText` — affected every xml array/gap fill, not just constants. Fixed via `NewXMLText("")`; `xml-gap-v.bal` is a real regression test for it (confirmed `v is xml:Text` would have been false under the old code) |
| Patch coverage | 67% of 46 changed statement lines — below 80% target, but every uncovered range individually accounted for above (Must-fix #2, Flagged) rather than left as a bare number |
| jBallerina parity | 2/2 spot-checked programs match: nested list/tuple filler shapes + readonly flags (`[[0,0],[0,0]]`, `[[7,0],[0,0]]`, etc., all readonly `true`) matched exactly; mutation panic on a filled nested list matches in behavior and line number (message text differs — jBallerina raises a typed `InvalidUpdate` error, Nutcracker a generic message; per `AGENTS.md` panic tests validate only the first stack-trace line, not message text) |
| Verification pass | 1 initial hypothesis (PR description's XML-filler bullet contradicting the code's `unimplementedFailure` for XML) was checked against `values/values.go` and dropped once it became clear the description's XML bullet refers to the *general* runtime filler fix, not the constant-specific path, which are two different, both-correct things |
| Commands | `go build ./...`, `go vet ./...`, `gofmt -l -s`, `go run honnef.co/go/tools/cmd/staticcheck@latest ./semantics/... ./values/...`, `go test ./semantics/... ./values/... ./ast/... ./bir/... ./desugar/... ./nodebuilder/... ./common/...`, `go test ./corpus/...`, `patch_coverage.py`, `go run ./cli/cmd run` + `bal run` on 2 spot-check programs (nested filler shapes/readonly, mutation panic) |

|---|---|---|---|
| semantics/internal/types/constant_evaluator.go | 45 | 67% | 311-313, 332-333, 341-346, 354, 365, 377-378 |
| values/values.go | 1 | 100% | — |

<details>

```
issue: 907
reviewed_diff: sha256:72fa02b9893cef286ed6363668d2d4c39dfedbb77cb21b2d07510b70e4e16fbb
skill_version: sha256:dee208f9f7a47a193b4d0bd8c00d0bcfffad31b0de26a78f39e8b33e01464f57
date: 2026-09-24T05:50:13Z
patch_coverage: 67
important: 2
verify: 0
verdict: "2 Important findings — fix first"
```
</details>

PR #917, opened 2026-09-09
fix(types): avoid panic resolving recursive list defaults

### LINKED ISSUE #913: [Bug]: compiler panics on a record field default that constructs a list of the enclosing record type
### Description

Declaring a record whose field default is a list constructor, where the list's member type is the record being declared, crashes the compiler with a Go nil-pointer panic instead of compiling or reporting a diagnostic.

The record does not have to be used as a constant — an ordinary runtime record construction is enough to trigger it.

### Steps to Reproduce

```ballerina
import ballerina/io;

type Node record {|
    Node[] children = [];
|};

public function main() {
    Node n = {};
    io:println(n);
}
```

Expected output:

```
{"children":[]}
```

Actual:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x20]

goroutine 1 [running]:
semtypes.mappingFormulaIsEmpty(...)                        semtypes/mapping_ops.go:49
semtypes.IsEmpty(...)                                      semtypes/core.go:242
semantics/internal/types.selectListInherentType(...)
semantics/internal/types.resolveListConstructorWithExpectedType(...)
semantics/internal/types.resolveListConstructorExpr(...)
semantics/internal/types.resolveActionOrExpression(...)
semantics/internal/types.resolveBTypeInner(...)
semantics/internal/types.resolveTypeDefinition(...)
```

#### Scope

The crash is specific to a **list** constructor default whose member type is the enclosing record. These sibling declarations all compile and run correctly on the same commit:

| Declaration | Result |
| --- | --- |
| `type Node record {\| Node[] children = []; \|}` | **panic** |
| `type Node record {\| Node? next = (); \|}` | `{"next":null}` |
| `type Node record {\| map<Node> m = {}; \|}` | `{"m":{}}` |
| `type Node record {\| Node[] children; \|}` (no default) | `{"children":[]}` |

Adding further fields does not change the outcome — `type Node record {| Node[] children = []; int v = 1; |}` panics as well.

### Affected Version(s)

`main` @ `e5005d17d`

macOS (darwin/arm64), reproduced with `go run ./cli/cmd run <file>.bal`.

### Related area

Type checker

The mapping-constructor analogue of this problem — `type R record {| R inner = {}; |}`, which panics identically on `main` — is being given a proper `cyclic dependency in record field default` diagnostic as part of the constant record defaults work (#895). That change does not cover the list-constructor path, so this is filed separately.

Type/Bug

### Possible cause

While a record's field defaults are being resolved, the record's mapping definition has not been defined yet, so its semtype is a recursive atom whose atomic type is still `nil`. Resolving the `[]` default against `Node[]` reaches `selectListInherentType`, which calls `semtypes.IsEmpty` on a type containing that unset atom, and `mappingFormulaIsEmpty` dereferences it.

heshanpadmasiri:
## Summary
- safely resolve list constructor defaults containing incomplete recursive types
- add end-to-end corpus coverage for recursive record list defaults

## Testing
- `go test ./...`
- pre-commit build and lint hooks

Closes #913

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed type resolution for recursive list types with default empty values.
  - Recursive records now correctly initialize empty child lists and produce the expected output.

- **Tests**
  - Added coverage for recursive list defaults, including verification that empty child collections are represented correctly.


warunalakshitha (approved):
LGTM

PR #919, opened 2026-09-09
fix(types): resolve global variables in default expressions

### LINKED ISSUE #915: [Bug]: module-level variable reference in a default value expression is typed `never`
### Description

A reference to a **module-level variable** inside a **default value expression** does not get its type resolved. Wherever the expression needs that variable's type — as a call argument, or as the receiver of a langlib method — compilation fails, even though the exact same expression compiles and runs correctly outside a default.

This affects function default parameters, object/class method default parameters (including `init`), and record field defaults. No langlib, generics, or objects are required to hit it:

```ballerina
final int base = 5;

function twice(int x) returns int => x * 2;

function f(int n = twice(base)) returns int => n;   // module var as a call argument

public function main() {
    io:println(f());
}
```

```
error[SEMANTIC_ERROR]: incompatible type: expected int, got never
 --> t.bal:4:26
  |
4 | function f(int n = twice(base)) returns int => n;
  |                          ^^^^
```

`base` is typed `never`. The same call as a module-level initializer (`final int n = twice(base);`) prints `10`, and a bare `int n = base` default also works — the failure needs the reference's type to be *inferred* rather than taken from the surrounding context.

The bug has three user-visible faces, all from the same missing type.

**1. As a call argument → `expected <T>, got never`.** Fatal in function defaults, `init` defaults, and with langlib calls:

```ballerina
final readonly & int[] letters = [10, 20, 30, 20];

class C {
    int n;
    function init(int n = array:length(letters)) { self.n = n; }
}
```

```
error[SEMANTIC_ERROR]: incompatible type: expected [nil|boolean|int|float|decimal|string|error|typedesc|handle|function|regexp|future|stream|list|mapping|table|xml|object...], got never
 --> t.bal:6:40
  |
6 |     function init(int n = array:length(letters)) { self.n = n; }
  |                                        ^^^^^^^
```

**2. As a langlib method receiver → `method not found`.** With no type there is nothing to resolve the method against:

```ballerina
final readonly & int[] letters = [10, 20, 30, 20];

type R record {|
    int len = letters.length();
|};
```

```
error[SEMANTIC_ERROR]: method not found: length
 --> t.bal:3:28
  |
3 | type R record {| int len = letters.length(); |};
  |                            ^^^^^^^^^^^^^^^^
```

The same `letters.length()` in a function body prints `4`, and `function f(int[] arr, int n = arr.length())` works — a *parameter* receiver is fine, only a module-level variable is not.

**3. In a record field default → silent.** This one compiles and produces the right answer at runtime, but leaves the AST node untyped, which trips the type resolver's own invariant check:

```ballerina
final int base = 5;
isolated function twice(int x) returns int => x * 2;

type R record {|
    int n = twice(base);   // prints 10 correctly
|};
```

```
TestTypeResolver: expression *ast.BLangVarRef at (36:777,781) does not have determined type set
```

This matters for corpus tests: any new corpus case with a module-level variable reference in a record field default fails `TestTypeResolver` and `TestSemanticAnalysis` while running correctly end to end.

### Steps to Reproduce

Save any snippet above as `t.bal` and run `bal run t.bal`. For face 3, add the snippet as a corpus case and run:

```
go test ./semantics/internal/types -run 'TestTypeResolver/<path-to-case>'
```

Controls that **do** work, for contrast:

| Snippet | Result |
| --- | --- |
| `final int n = twice(base);` (module initializer) | `10` |
| `function f(int n = base)` (bare ref, type from context) | `5` |
| `function f(int[] arr, int n = arr.length())` (param receiver) | `3` |
| `io:println(letters.length())` (function body) | `4` |

### Affected Version(s)

Reproduces on `main` at `e5005d17d`. Not a regression — it predates the branches currently in flight.

### Possible cause

The default-value expression is lowered into a compiler-generated provider function whose parameters stand for the preceding parameters. A module-level variable reference in that body appears not to be bound to the module symbol's resolved type at the point the provider is type-resolved, so it falls back to `never` (or to no determined type at all in the record-field path). The ordering between module-level variable type resolution and provider body type resolution looks like the place to check.

### Related area

Type checker

### Notes

Found while adding corpus coverage for opaque-function default parameters. Worth recording that this is **not** opaque-function specific: `array:indexOf` (opaque) in an `init` default works when the container's type resolves, e.g. `function init(int? a = array:indexOf(letters(), 30))` returns `2`, and so does the method form `letters().indexOf(30)`. Only the module-level variable reference breaks it.

heshanpadmasiri:
## Summary

- resolve declared and inferred module-level variable types lazily when referenced from default expressions
- preserve cycle detection for lazy global variable resolution
- add corpus coverage for function, record, object, and langlib default expressions

## Testing

- `go test ./...`
- commit hooks (build and golangci-lint)

Closes #915

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resolution of module-level variables used in default function arguments.
  * Default values now work reliably in language-library calls, object initialization, and record fields, including variables declared after the consuming code.
  * Supports both explicitly typed and inferred module-level variables in these default-value scenarios.

* **Tests**
  * Added coverage and expected-output checks for function, object, record, and language-library default-value behavior.


corpus/bal/subset10/10-function/default-module-var-ref-arg-v.bal
warunalakshitha, line 1: Should we add these new test files to subset 10?

semantics/internal/types/type_resolver.go
warunalakshitha: Types are now resolved lazily via `ensureResolved`, but this still says module vars “don’t go through that path”. I think this should be updated

corpus/bal/subset7/07-function/default-module-var-ref-arg-v.bal
```diff
@@ -0,0 +1,43 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/io;
+
+// The module-level variables are declared after the functions that reference
+// them from parameter defaults, so their types have to be resolved on demand
+// while the function signatures are being resolved.
+function tag(string s, string p = prefix) returns string {
```
warunalakshitha: The new corpus files cover defaults that already worked (`p = prefix`, `n = offset`) or nearby cases (`offset + 1`), but not the #915 failure: a module var passed as a call argument.
Example:

```ballerina
import ballerina/io;
final int base = 5;
isolated function twice(int x) returns int => x * 2;
function viaCall(int n = twice(base)) returns int => n;
public function main() {
    io:println(viaCall()); // @output 10
    io:println(viaCall(3)); // @output 3
}

warunalakshitha (approved):
LGTM

PR #920, opened 2026-09-09
fix(semantics): prevent record default symbol collisions

### LINKED ISSUE #914: [Bug]: function-local record type picks up a module-level record's field default
### Description

When a module declares record types with field defaults and a function declares its own inline record types with field defaults, the local record's default is silently replaced by a module-level record's default.

Field names and types do not need to match. The wrong value is used and no diagnostic is reported, so this is a silent wrong-answer bug.

### Steps to Reproduce

```ballerina
import ballerina/io;

type Outer record {|
    string s = "outer";
|};

public function main() {
    Outer o = {};
    io:println(o.s); // "outer" - correct

    record {|string t = "local";|} inner = {};
    io:println(inner.t); // expected "local"
}
```

Expected output:

```
outer
local
```

Actual output:

```
outer
outer
```

#### The substitution is positional

The Nth function-local record default takes the Nth module-level record default:

```ballerina
import ballerina/io;

type M1 record {| int a = 11; |};
type M2 record {| int b = 22; |};

public function main() {
    M1 m1 = {};
    M2 m2 = {};
    io:println(m1.a); // 11
    io:println(m2.b); // 22

    record {|int p = 1;|} l1 = {};
    record {|int q = 2;|} l2 = {};
    io:println(l1.p); // expected 1
    io:println(l2.q); // expected 2
}
```

Expected `11 / 22 / 1 / 2`, actual `11 / 22 / 11 / 22`.

#### What does not break

Two module-level records together, or two function-local records together, both behave correctly:

```ballerina
type A record {| int p = 1; |};
type B record {| int q = 2; |};
// A.p -> 1, B.q -> 2  (correct)
```

```ballerina
record {|int a = 1;|} x = {};
record {|int b = 2;|} y = {};
// x.a -> 1, y.b -> 2  (correct)
```

The corruption only happens across the module-level/function-local boundary.

### Affected Version(s)

`main` @ `e5005d17d`

macOS (darwin/arm64), reproduced with `go run ./cli/cmd run <file>.bal`.

### Related area

Compilation

Type/Bug

### Possible cause

Generated default-function names come from a counter (`nextDefaultFnName`). `packageTypeResolver` and each `functionTypeResolver` keep separate counters, so a function-local record's first generated default function is given the same name as the module's first one, and the BIR function-lookup key then resolves to the module-level default function.

heshanpadmasiri:
## Summary
- allocate record field default-function symbols during symbol resolution
- share the module-level generated symbol namespace across package, function, lambda, and listener scopes
- populate generated function signatures once field types are resolved
- add corpus coverage for module/local records, parameters, lambdas, methods, and listeners

## Testing
- `go test ./ast/... ./semantics/... ./desugar/... ./bir/... ./corpus/...`
- commit hooks: build and golangci-lint checks

Closes #914

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed record field default values in nested scopes and listener-related scenarios.
  * Ensured defaults resolve correctly across local records, function parameters, nested functions, and sibling scopes.
  * Prevented naming collisions from causing incorrect default-value behavior.
  * Improved handling of module-level variable types and constant value types.
  * Preserved expected listener initialization and attachment behavior.

* **Tests**
  * Added regression coverage for default-value resolution and listener behavior across affected scenarios.


semantics/internal/symbols/symbol_resolver.go
```diff
@@ -1759,6 +1771,23 @@ func resolveObjectInclusions[T symbolResolver](resolver T, unresolvedInclusions
...
+// allocateRecordDefaultSymbols allocates a module unique function symbol for each record field with a
+// default expression. Since the field type is not known at symbol resolution the typed signature is left
+// empty and populated during type resolution.
+func allocateRecordDefaultSymbols(resolver symbolResolver, recordType *ast.BLangRecordType) {
+	scope := resolver.recordDefaultScope()
+	for _, field := range recordType.FieldPtrs() {
+		if field.DefaultExpr == nil {
+			continue
+		}
+		name := resolver.nextDefaultSymbolName()
```
warunalakshitha: Nit: skip if DefaultFnRef is already set so a second walk does not orphan the previous symbol.

warunalakshitha (commented):
LGTM. Please fix the build

warunalakshitha (approved):

PR #927, opened 2026-09-11
fix(types): resolve list emptiness on the per-function type context

### LINKED ISSUE #924: Nightly race detector build failing
The nightly race detector workflow failed on `c453fa0`.

Run: https://github.com/ballerina-nutcracker/ballerina/actions/runs/34447255712

heshanpadmasiri:
## Purpose

Fixes #924 — the nightly race detector build has been failing since 2026-09-10 (`c453fa0`). The 2026-09-09 nightly on `bcbe17fb` was the last green run.

## Cause

Introduced by **#917** (commit `52e766a`, "fix(types): avoid panic resolving recursive list defaults"), which changed one line in `selectListInherentType`:

```go
-	if semtypes.IsEmpty(tc, expectedListType) {          // the resolver's own context
+	if !t.ensureNotEmpty(expectedListType, func() { ...  // hops to the package resolver
```

`functionTypeResolver.ensureNotEmpty` forwarded straight to `parentResolver.ensureNotEmpty`, which runs the check against the package-wide `semtypes.Context`.

`ResolvePrivateNodes` resolves every function body in its own goroutine, each with its own `semtypes.ContextFrom(...)` — precisely so the semtypes memo tables stay goroutine-local. Routing this one check through the parent punched a hole in that isolation: concurrent goroutines then read and write the shared `cx.listMemo()` map (`semtypes/common.go:138`) and the shared `deferredEmptinessChecks` slice.

Confirmed by reverting just that line: 0 races, and the only failure is `recursive-list-default-v.bal`, the bug #917 fixed.

## Fix

`functionTypeResolver.ensureNotEmpty` now uses its own type context once the type env is ready, and delegates to the parent only in the pre-ready sequential phase, where deferral is genuinely needed. The #917 fix is preserved; the emptiness check just runs on the goroutine-local context.

## Testing

- `go test -race ./birgen/ -run TestBIRGeneration` — **ok**. Previously 16+ `WARNING: DATA RACE` and 3 test failures.
- `make test-race` (full suite) — **0 data races**. `semantics/internal/types`, `semantics/internal/analysis`, `cfg`, `symbols` and `birgen` all pass clean.
  - One unrelated failure locally: `corpus/TestBalBuildNativeDependency`, from a missing `../ballerina-driver-src-dev` checkout on this machine. Fails identically on unmodified `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type resolution for concurrently processed function bodies.
  * Prevented unnecessary deferred checks once the relevant type environment is ready.
  * Fixed handling of recursive record defaults that return empty lists, ensuring valid values are recognized correctly.

* **Tests**
  * Added regression coverage for recursive list and lambda defaults, including expected successful output.
  * Improved compiler test consistency for generated functions without affecting user-defined function ordering.


semantics/internal/types/type_resolver.go
```diff
@@ -486,8 +486,18 @@ func (f *functionTypeResolver) lookupClassMethodSymbol(receiverTy semtypes.SemTy
 	return f.parentResolver.lookupClassMethodSymbol(receiverTy, methodName)
 }
 
+// ensureNotEmpty checks emptiness with this resolver's own type context. Function
+// bodies are resolved concurrently, so the shared package resolver context (and its
+// deferred check list) must not be touched once the type env is ready.
 func (f *functionTypeResolver) ensureNotEmpty(ty semtypes.SemType, onEmpty func()) bool {
-	return f.parentResolver.ensureNotEmpty(ty, onEmpty)
+	if !f.typeEnv().IsReady() {
+		return f.parentResolver.ensureNotEmpty(ty, onEmpty)
```
warunalakshitha, line 495: This path isn’t covered — a lambda around the recursive [] default hits it (nil-deref without the hop). Add something like corpus/bal/subset10/10-bug/recursive-list-lambda-default-v.bal:

```
import ballerina/io;

type Node record {|
    function () returns Node[] make = function () returns Node[] { return []; };
|};

public function main() {
    Node node = {};
    io:println(node is Node); // @output true
}
```

heshanpadmasiri: fixed

heshanpadmasiri:
Superseded by a stacked pair of PRs from branches in this repository (cross-fork PRs cannot be stacked). Closing in favour of the new stack.

PR #928, opened 2026-09-11
test: cover trap on remote and resource actions

### LINKED ISSUE #829: nodebuilder: trap applied to an action panics the compiler
## Component

- [ ] Dependent stdlib: `ballerina/<name>`

## Summary

`trap` applied to an action — a remote method call (`trap c->rget()`) or a client resource access (`trap c->/albums`) — panics during AST construction instead of compiling. Found while writing corpus coverage for the `ballerina/http` client resource methods; it is not specific to `http` and reproduces with a plain user-defined `client class`.

`transformTrapExpression` (`nodebuilder/node_builder.go:3057`) calls `createExpression` on its operand, and `createExpression` (`node_builder.go:2000`) panics for any operand that transforms to a non-expression node. Actions are exactly that case: `*st.RemoteMethodCallActionNode` becomes `*ast.BLangRemoteMethodCallAction` and `*st.ClientResourceAccessActionNode` becomes `*ast.BLangClientResourceAccessAction`, neither of which is an expression node. The sibling `resolveTrapExpr` on the type-resolution side already goes through `resolveActionOrExpression`, so the trap machinery below the node builder appears to expect actions to be permitted.

## Reproduction

```ballerina
import ballerina/io;

client class MyClient {
    resource function get albums() returns int {
        return 1;
    }

    remote function rget() returns int {
        return 1;
    }
}

public function main() {
    MyClient c = new;
    int|error viaRemote = trap c->rget();
    io:println(viaRemote);
    int|error viaResource = trap c->/albums;
    io:println(viaResource);
}
```

Run with: `go run ./cli/cmd run trap1.bal`

Removing the `trap c->/albums` line still panics, on the remote call instead — so both action forms are affected, and the remote form is hit first.

## Expected vs actual

- **Expected:** compiles and prints `1` twice. jBallerina 2201.13.5 (Swan Lake Update 13) runs this file and does exactly that.
- **Actual:** a Go panic during AST construction, with no diagnostic:

```
panic: syntax node *st.RemoteMethodCallActionNode transformed to non-expression node *ast.BLangRemoteMethodCallAction

goroutine 1 [running]:
github.com/ballerina-nutcracker/ballerina/nodebuilder.(*nodeBuilder).createExpression(...)
	nodebuilder/node_builder.go:2000
github.com/ballerina-nutcracker/ballerina/nodebuilder.(*nodeBuilder).transformTrapExpression(...)
	nodebuilder/node_builder.go:3057
```

Verified against `origin/main` (`00172daec`) with a freshly built `./cli/cmd`; the panic is identical there, so it is pre-existing and not introduced by the client resource method work.

## Workaround applied in this port

None needed — the corpus test that surfaced it simply does not use `trap` on a resource method call. Recorded so the panic gets a real diagnostic (or support) rather than being rediscovered. A compiler panic with no diagnostic is worth fixing even if `trap` on an action stays unsupported: the node builder should report an unimplemented-feature error instead.

heshanpadmasiri:
## Purpose
Closes #829.

`trap` already accepts action operands following the start/wait work (`31d61107b`). The current AST, desugar expression thunks, and BIR trap regions handle remote and client resource calls without further production changes. Add dedicated regression coverage for the original report and operand-evaluation boundaries.

## Approach
Add three corpus cases and their AST, CFG, desugared, BIR, and integration goldens where applicable:
- `trap-action1-v.bal`: remote/resource success, method panics, returned-error identity, nested traps, and continued execution.
- `trap-action2-v.bal`: panics in receivers, computed resource paths, explicit/default arguments, and branching spread expressions requiring lowering setup. Call counters verify failed operand evaluation does not invoke the method and subsequent calls still execute.
- `trap-action3-e.bal`: assigning either trapped action to `int` rejects the `int|error` result with ordinary diagnostics.

## Automation tests
Passed:
```sh
go test ./nodebuilder ./semantics/... ./desugar ./birgen ./corpus \
  -run 'Test[^/]+/subset9/09-object/trap-action' -count=1
```
Both valid cases produce matching output under jBallerina (`bal run`); the error case reports incompatible types on the same two lines. Commit-hook builds and lint checks passed.

## Documentation
N/A — regression tests for existing behavior; no production API changes.

## Release note
Add regression coverage for trapping remote method and client resource access actions.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for trapping client remote and resource method calls.
  - Verified handling of successful results, panics, returned errors, nested traps, argument evaluation failures, and continued execution.
  - Added expected runtime output and semantic error validations for trapped calls assigned to incompatible types.

### DIFF
--- a/corpus/ast/subset9/09-object/trap-action1-v.txt
+++ b/corpus/ast/subset9/09-object/trap-action1-v.txt
@@ -1,0 +1,155 @@
+(compilation-unit
+(package-id $anon . 0.0.0)
+  (import-package ballerina io (as io))
+  (class-definition Client
+    (function $remote$broken () (
+      (value-type int))
+      (block-function-body
+        (panic
+          (error-constructor-expr (
+            (literal remote panic))))))
+    (function $remote$returnError (
+      (variable failure (type
+        (error-type)))) (
+      (union-type
+        (value-type int)
+        (error-type)))
+      (block-function-body
+        (return
+          (simple-var-ref failure))))
+    (function $remote$value () (
+      (value-type int))
+      (block-function-body
+        (return
+          (literal 1))))
+    (resource-function get name:albums
+      (value-type int)
+      (block-function-body
+        (return
+          (literal 1))))
+    (resource-function get name:broken
+      (value-type int)
+      (block-function-body
+        (panic
+          (error-constructor-expr (
+            (literal resource panic))))))
+    (resource-function get name:failure
+      (variable failure (type
+        (error-type)))
+      (union-type
+        (value-type int)
+        (error-type))
+      (block-function-body
+        (return
+          (simple-var-ref failure)))))
+  (function main () (
+    (value-type null))
+    (block-function-body
+      (var-def
+        (variable c (type
+          (user-defined-type Client)) (expr
+          (new ()))))
+      (var-def
+        (variable remoteValue (type
+          (union-type
+            (value-type int)
+            (error-type))) (expr
+          (trap-expr
+            (remote-method-call value expr:
+              (simple-var-ref c) ())))))
+      (expression-stmt
+        (invocation io println (
+          (simple-var-ref remoteValue))))
+      (var-def
+        (variable resourceValue (type
+          (union-type
+            (value-type int)
+            (error-type))) (expr
+          (trap-expr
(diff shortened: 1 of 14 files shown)


warunalakshitha (approved):
LGTM

PR #933, opened 2026-09-11
feat(projects): add support for includes field

azinneera:
Fixes partially https://github.com/ballerina-nutcracker/ballerina/issues/435

## Checklist

## Summary by CodeRabbit

* **New Features**
  * `bal pack` now supports the package-level `include` configuration in `Ballerina.toml`.
  * Include patterns support glob matching, directory expansion, exclusions with `!`, extensions, character classes, and brace alternatives.
  * Documentation and asset files can be selected for inclusion in generated packages.

* **Bug Fixes**
  * Prevented duplicate files when include patterns overlap.
  * Ensured exclusions inside matched directories are applied correctly.
  * Prevented symlinked files and `target` contents from being packaged.


projects/include_matcher.go
```diff
@@ -0,0 +1,245 @@
...
+func writeGlobBody(sb *strings.Builder, runes []rune) error {
+	for i := 0; i < len(runes); i++ {
+		switch c := runes[i]; c {
+		case '*':
+			if i+1 < len(runes) && runes[i+1] == '*' {
+				sb.WriteString(".*")
+				i++
+				if i+1 < len(runes) && runes[i+1] == '/' {
+					i++
+				}
```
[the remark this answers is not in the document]
azinneera: Confirmed real — traced it: `babble/**/bar` compiled to `^(?:.*/)?babble/.*bar$`, which matches `babble/notbar`. Applied the suggested fix (`(?:.*/)?` for `**/` instead of bare `.*`), verified it now rejects `babble/notbar` while still matching zero/one/multi-directory cases, and added `babble/notbar` as a fixture entry so the existing `mid doublestar` test case catches this as a regression.

projects/bala_writer.go
```diff
@@ -185,21 +195,75 @@ func writeModuleSources(zw *zip.Writer, pkg *Package) error {
...
+}
+
+// addIncludeFile writes a single include file into zw at relPath, skipping
+// paths already written — an include pattern may match the same path
+// twice, or match a path some earlier archive stage already wrote.
+func addIncludeFile(zw *zip.Writer, fsys fs.FS, fsPath, relPath string, written map[string]bool) error {
+	if written[relPath] {
+		return nil
+	}
+	content, err := fs.ReadFile(fsys, fsPath)
```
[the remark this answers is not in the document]
azinneera: Checked — this isn't specific to the includes feature. fs.ReadFile(fsys, ...) on the raw, unwrapped project fs.FS is used the same way throughout config_creator.go for Ballerina.toml, Dependencies.toml, README files, and module sources — the whole project loader reads through this same unsandboxed fs.FS, not just addIncludeFile. Also checked Java's own BalaWriter/ProjectUtils reference implementation — no symlink containment there either. Fixing just this one call site would be inconsistent and give a false sense of safety while leaving the identical exposure open everywhere else. A real fix needs a project-wide root-containment wrapper applied once where the project's fs.FS is established, which is well beyond this PR's scope (adding include-field support) — leaving this open rather than patching narrowly.

projects/include_matcher.go
```diff
@@ -0,0 +1,252 @@
...
+				// would silently bundle the file the negation was meant to
+				// exclude.
+				descendants, err := expandDirToFiles(fsys, root, rel)
+				if err != nil {
+					return err
+				}
+				matches = append(matches, descendants...)
+				return fs.SkipDir
+			}
+			matches = append(matches, rel)
```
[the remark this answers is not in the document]
azinneera: Fixed in  [95108e9](https://github.com/ballerina-nutcracker/ballerina/pull/933/commits/95108e9688535c3ba764c7c63d85fb4a621d7bc9)

projects/bala_writer.go
```diff
@@ -185,21 +195,75 @@ func writeModuleSources(zw *zip.Writer, pkg *Package) error {
...
+			if err := writeZipEntry(zw, zipPath, content, written); err != nil {
 				return err
 			}
 		}
 	}
 	return nil
 }
 
-func writeZipEntry(zw *zip.Writer, zipPath string, content []byte) error {
+// addIncludes copies files matching the package manifest's `include` glob
```
warunalakshitha, line 204: For me this comment is too long. can we simplify and provide only high level work of the function

projects/include_matcher.go
```diff
@@ -0,0 +1,265 @@
...
+	"io/fs"
+	"regexp"
+	"slices"
+	"strings"
+)
+
+// resolveIncludePaths resolves the `include` glob patterns declared in
+// Ballerina.toml against root (a directory within fsys, per the fs.FS
+// convention "." denotes fsys's own root), returning root-relative paths
+// (slash-separated) of every matching file or directory. A pattern prefixed
```
warunalakshitha: doc says returns “file or directory”, but dirs are always expanded to files

projects/include_matcher.go
```diff
@@ -0,0 +1,265 @@
...
+				// Expand to the directory's individual files rather than
+				// keeping one directory entry, so a later "!" pattern can
+				// negate a specific file inside it — resolveIncludePaths
+				// only ever removes exact-path matches, so a negation
+				// against e.g. "assets/secret.txt" would otherwise never
+				// match a bare "assets" entry, and the caller (which
+				// recursively archives whatever directories are matched)
+				// would silently bundle the file the negation was meant to
+				// exclude.
+				descendants, err := expandDirToFiles(fsys, root, rel)
```
warunalakshitha, line 95: expandDirToFiles already have a doc comment. so do we need this lengthy comment?

projects/bala_writer.go
```diff
@@ -76,24 +77,33 @@ func writeBala(pkg *Package, resolution *PackageResolution, outputDir string) (s
 }
 
 // populateBalaArchive writes every entry of the bala layout into zw. The
-// caller is responsible for closing zw.
+// caller is responsible for closing zw. written tracks every zip path
+// written so far across all stages, so a later stage (in practice, only
+// addIncludes) can silently skip a path some earlier stage already wrote —
+// matching Java's BalaWriter, which relies on ZipOutputStream throwing on
```
warunalakshitha: do we need this part
// matching Java's BalaWriter, which relies on ZipOutputStream throwing on
// any duplicate entry name and swallows that specific exception only
// around its own addIncludes call.

projects/include_matcher.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// relativeToRoot strips root's "root/" prefix from p (fs.FS-style
+// slash-separated paths), leaving p unchanged when root is "." — fs.FS's
+// own-root marker — or empty.
+func relativeToRoot(root, p string) string {
+	if root == "" || root == "." {
+		return p
+	}
+	return strings.TrimPrefix(p, root+"/")
```
warunalakshitha, line 148: can we write test for cover root path for both relativeToRoot and joinRoot?

projects/bala_writer.go
azinneera: Removed in 8b22099

projects/include_matcher.go
azinneera: Added in [8540154](https://github.com/ballerina-nutcracker/ballerina/pull/933/commits/8540154f0abb31a1d6506246e1cc8a2eac90ef57)

projects/include_matcher.go
azinneera: No, trimmed it in https://github.com/ballerina-nutcracker/ballerina/commit/8b22099c1bebdc123388f1b27df29af060757e62

projects/include_matcher.go
azinneera: Fixed in https://github.com/ballerina-nutcracker/ballerina/commit/8b22099c1bebdc123388f1b27df29af060757e62

projects/bala_writer.go
azinneera: Fixed in https://github.com/ballerina-nutcracker/ballerina/commit/8b22099c1bebdc123388f1b27df29af060757e62

warunalakshitha:
# Review report — #933 feat(projects): add support for includes field

> [!NOTE]

## Summary
**2 Important findings — fix first.** The headline one is a real, repro-confirmed bug: `bal pack` silently writes a **duplicate zip entry** into the `.bala` archive when an `include` pattern matches the same path a package/module readme also resolves to — confirmed by packing an actual project and inspecting the output zip. jBallerina's own `BalaWriter` explicitly guards against exactly this case (different stage order, plus an explicit duplicate-entry catch); this port has neither. `INTENT.md` is the second, process-only finding. Everything else is solid: the glob-to-regexp translation and its "root-only"/"directory-only" restrictions were checked line-by-line against jBallerina's actual `ProjectUtils` source (not just the doc comment) and match exactly, including a non-obvious quirk (a leading-`/` pattern with a nested body can never match anything, in both implementations); build/vet/tests/corpus all pass; patch coverage is 79%, one point under target, entirely from small error-propagation branches.

## Must fix (Important)
|---|---|---|---|---|
| 1 | Correctness (jBallerina parity) | `projects/bala_writer.go`: `populateBalaArchive` (stage order) + `addBalaDoc` (no dedup check) | Confirmed by repro: a project with `readme = "docs/README.md"` and `include = ["docs/**"]` produces a `.bala` containing **two** `docs/README.md` zip entries (Go's `archive/zip.Writer.Create` doesn't reject duplicate names, so this succeeds silently with a malformed archive). Root cause is two-fold: (a) `addIncludes` runs *before* `addBalaDocs` — jBallerina's `BalaWriter` runs `addIncludes` **after** `addPackageDoc`; (b) only `addIncludeFile` consults the `written` map before writing — `addBalaDoc` doesn't, so it can't be skipped when an earlier `addIncludes` already wrote its path. jBallerina's own `addIncludes` additionally catches `ZipException`/"duplicate entry" and swallows it as a defense-in-depth measure that this port has neither copy of. | Reorder `populateBalaArchive` so `addIncludes` runs after `addBalaDocs` (matching jBallerina), **and** have `addBalaDoc` (or `writeZipEntry` itself, as a single choke point) check `written` before writing — reordering alone still leaves the reverse collision (an include pattern matching a path a *later*, still-unprotected stage would also write) open to the same bug from the other direction. |
| 2 | Process | `docs/changes/v0.7.0/933/INTENT.md` | Missing on this branch | This run drafted it from the diff and PR description (see file). Per the pre-pr-review process, rewrite Problem/Approach in your own words before opening the PR — an agent-drafted INTENT.md is itself an Important finding. |

None.

| Kind | Detail |
|---|---|
| PR checklist | Both checklist boxes ("Read the Contributing Guide", "Added or updated corpus tests") are unchecked, but a real end-to-end corpus test (`TestBalPackCorpus_IncludeField`) and 15+ unit tests are present — looks like the boxes were just never ticked, not an actual gap. |
| Coverage detail | `bala_writer.go`'s uncovered lines (101-102, 225-230, 232-235, 237, 247-251) are the `addIncludes`/`resolveIncludePaths`/`fs.ReadFile` error-propagation branches — plausible to trigger (a malformed glob pattern in `Ballerina.toml`, or a matched path that vanishes between match and read) and not shown to be unreachable; a small addition would close this rather than leaving it unaddressed. |
| Coverage detail | `include_matcher.go`'s uncovered lines are similarly almost all error-return branches (invalid glob syntax, `fs.WalkDir` propagating a filesystem error) — `TestGlobToRegexp_InvalidPatterns` already covers some invalid-pattern cases; the remaining gaps are narrower variants of the same category, not a design gap. |
| Tooling limitation (not a PR933 issue) | `TestBalPackCorpus_IncludeField` runs the real `bal` binary as a subprocess (`integrationTestBalCLI`), so its coverage of `bala_writer.go`/`include_matcher.go` isn't visible to `-coverprofile` on the test process — the 79% number undercounts what this PR's own test suite actually exercises end-to-end. |
| New dependency | none |
| Corpus fixture changes | all new files/additions; no existing golden changed |

## Fixed during review
None — this is a review of an already-open external PR (#933), not a branch this session can push to, so findings are reported rather than fixed.

## Evidence
| Pass | Result |
|---|---|
| Branch / scope | PR933 head `23bc60fda` vs merge-base `451f950f9` (= `main`); 3 commits, 13 files (+843 / −30) |
| INTENT.md | missing — drafted by this run, flagged Important |
| Commits | Conventional ✔ on all 3 (`feat(projects): add support for includes field`, `fix(projects): address review suggestions`, `fix(projects): resolve joinRoot/writeZipEntry conflicts from rebase`) · no Co-authored-by ✔ · no AI author ✔ (single author/committer, human email) |
| Lint, fmt, license headers | `gofmt -l -s` clean; license headers present on new files; `go vet` clean |
| Dead code | none found |
| `go test` + corpus | `go build ./...`, `go vet ./...` clean; `go test ./projects/...` (all unit tests, including symlink-skipping and non-trivial-root cases) passes; `TestBalPackCorpus_IncludeField` (real `bal pack` + `.bala` zip inspection) passes |
| Reuse before reimplement (mechanical pass) | Listed every new function name and grepped the whole repo for each — no pre-existing duplicate. `joinRoot` is a relocation (`manifest_builder.go` → `include_matcher.go`), not a new duplicate. |
| Correctness — glob semantics | Checked `globToRegexp`'s `**`/`**/`-segment handling and `isCorrectIncludeMatch`'s leading-`/`/trailing-`/` restrictions line-by-line against jBallerina's actual `ProjectUtils.java` (`getPatternPrefix`, `removeTrailingSlashes`, `isCorrectPatternPathMatch`) — not just the port's own doc comment. Both implementations produce the *same* glob regardless of leading `/`; the root-only restriction is a pure post-filter on the matched path's own depth in both. Confirmed a non-obvious shared quirk: a leading-`/` pattern with a nested body (e.g. `/docs/*.md`) can never match anything in *either* implementation, since a nested file's parent is never the project root. |
| Correctness — the actual bug | Confirmed by direct repro, not inspection: packed a real project (`readme = "docs/README.md"`, `include = ["docs/**"]`) and inspected the resulting `.bala` with Python's `zipfile` — `docs/README.md` appears twice. Traced to jBallerina's `BalaWriter.java` (`populateBalaArchive` calls `addPackageDoc` before `addIncludes`, and `addIncludes` catches `ZipException`/"duplicate entry") to confirm this is a genuine divergence, not an intentional design difference. |
| Security | Symlink handling (never followed/archived, since `fs.ReadFile` would follow one) and `target/` exclusion are each implemented identically in `matchIncludePattern` and `expandDirToFiles` — flagged as duplicated logic worth a shared helper, but not Important (same behavior, matching comments, no divergence risk observed) |
| Patch coverage | 79% of 190 changed statement lines (target 80%) — one point under, entirely from error-propagation branches (see Flagged); real end-to-end coverage is higher than this number shows, since the CLI integration test's subprocess coverage isn't captured |
| jBallerina parity | Checked against `ProjectUtils.java`/`BalaWriter.java` source directly (not a running `bal`, since this is packaging/CLI behavior, not language semantics) — glob semantics match exactly; the archive-writing stage order and duplicate-handling do not, which is the Important finding above |
| Verification pass | Every unreachability/behavior claim above is backed by either a source-level trace against the actual jBallerina implementation or a runnable repro — none asserted from inspection alone, per the skill's own rule (added after two earlier reviews in this series caught exactly that kind of unverified claim) |
| Commands | `go build ./...`, `go vet ./...`, `gofmt -l -s`, `go test ./projects/... -v`, `go test ./corpus/... -run TestBalPackCorpus_IncludeField -v`, `patch_coverage.py`, `go run ./cli/cmd pack` on a constructed repro project + `python3 -m zipfile` inspection of the output |

|---|---|---|---|
| projects/bala_writer.go | 41 | 56% | 101-102, 225-230, 232-235, 237, 247-251 |
| projects/config_creator.go | 1 | 100% | — |
| projects/include_matcher.go | 143 | 85% | 42-43, 51-52, 63-64, 69-70, 80, 97-98, 107-108, 119-120, 123-126, 240-241 |
| projects/manifest_builder.go | 2 | 100% | — |
| projects/package_manifest.go | 3 | 100% | — |

<details>

```
issue: 933
reviewed_diff: sha256:5ff7d122d71cfb3ad2194edc20583ca398cf3ef85926c986a5c9f57b90d4a6a8
skill_version: sha256:c3008dd0e3e303a8e734999ca526e6ce3566ede56e5078812d922fde728ba007
date: 2026-09-24T06:32:27Z
patch_coverage: 79
important: 2
verify: 0
verdict: "2 Important findings — fix first"
```
</details>

PR #943, opened 2026-09-14
fix(semtypes): include function quals in function type rendering

### LINKED ISSUE #942: [Bug]: function type diagnostics omit the isolated qualifier
### Description

This is a diagnostics defect, not a conformance one — the compiler correctly rejects the program, as does jBallerina. Only the message is wrong.

Per §3.4.2 the qualifier is part of the type:

> All function values belongs to a function-type-descriptor `function`; a function value belongs to the function-type-descriptor `isolated function` if and only if it is an isolated function.

Since `function-type-descriptor := function-quals function function-signature`, rendering a function type without its qualifiers renders a different type than the one being reported. Both sides of the message then collapse to identical text and the diagnostic conveys nothing about what is wrong.

### Steps to Reproduce

`main.bal`:

```ballerina
function plain() {
}

public function main() {
    function () f = plain;
    isolated function () g = f;
    g();
}
```

nBallerina:

```
error[SEMANTIC_ERROR]: incompatible type: expected function() returns nil, got function() returns nil
 --> main.bal:6:30
  |
6 |     isolated function () g = f;
  |                              ^
```

jBallerina 2201.13.4:

```
ERROR [main.bal:(6:30,6:31)] incompatible types: expected 'isolated function () returns ()', found 'function () returns ()'
```

### Possible cause

`semtypes/to_string.go:346` prints the `isolated` qualifier for object types, but there is no equivalent path for function types, so the qualifier is dropped from the rendering.

Worth fixing ahead of isolation inference, which will compare many more functions against isolated function types and make this message class common.

### Affected Version(s)

`main` @ `5da2071fc` — Ballerina dev (Nutcracker), Language specification 2024R1

### Related area

Semantics

heshanpadmasiri:
Fixes #942

### Problem

`function-type-descriptor := function-quals function function-signature`, but `to_string.go` rendered function types without their qualifiers. A diagnostic comparing an `isolated function` against a plain `function` collapsed to identical text on both sides and conveyed nothing:

```
error[SEMANTIC_ERROR]: incompatible type: expected function() returns nil, got function() returns nil
```

### Fix

Render the `isolated`/`transactional` qualifiers from the function atom, for both standalone function types and object methods. The same message now reads:

```
error[SEMANTIC_ERROR]: incompatible type: expected isolated function() returns nil, got function() returns nil
```

### Tests

- New corpus test `corpus/bal/subset8/08-function/isolated-fn-14-e.bal` with the integration golden, reproducing the issue's example.
- Unit tests in `semtypes/to_string_test.go` covering isolated, transactional, isolated+transactional function types and isolated object methods.
- Existing BIR and integration goldens updated where the qualifier now appears.

`go test ./...` passes.

## Summary by CodeRabbit

- **New Features**
  - Function type and object method descriptions now display `isolated` and `transactional` qualifiers accurately.

- **Bug Fixes**
  - Stream type-mismatch diagnostics now correctly include isolation requirements for `next` and `close` methods.
  - Compiler diagnostics now clearly identify invalid assignments from non-isolated functions to isolated function variables.

- **Tests**
  - Added coverage for isolated, transactional, and combined function qualifiers, isolated object methods, and invalid function assignments.


warunalakshitha (approved):
LGTM

PR #947, opened 2026-09-15
test(desugar): canonicalise generated fn order in goldens

heshanpadmasiri:
## Purpose

Bottom of the stack for #948 (fix for #924).

Compiler-generated top level functions (`$default$N`, `$desugar$N` and friends) are appended to the package while ranging over maps keyed by symbol, so their relative order varies between runs even though each function body is identical. That makes the desugar goldens flaky once the type resolution work in #948 changes how many of them are produced.

## Approach

`normalizeDesugaredAST` now canonicalises the generated functions before comparing against the golden: only the slots already holding a `$`-prefixed function are rewritten, sorted by their printed form. User written functions stay where the frontend put them, so a genuine reordering of those still fails the comparison.

## Testing

- `go test ./desugar` — ok

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Compiler-generated functions now appear in a consistent, deterministic order in printed output.
  * Canonicalized output and initialization behavior are more stable when generated functions are present.


warunalakshitha (approved):
LGTM

desugar/corpus_desugar_test.go
```diff
@@ -318,10 +318,51 @@ func normalizeDesugaredAST(s string) string {
...
 }
 
+// canonicaliseGeneratedFns sorts the compiler-generated top level functions
+// ($default$N, $desugar$N and friends) among themselves. They are appended to
+// the package while ranging over maps keyed by symbol, so their relative order
+// varies between runs even though each function body is identical. Only the
+// slots already holding a generated function are rewritten, leaving the
+// user written functions where the frontend put them so a genuine reordering
+// of those still fails the comparison.
+func canonicaliseGeneratedFns(e *sexp) {
```
warunalakshitha, line 333: since we are handling compare part, will this work with -update as well. or -update will write different txt every time.

desugar/corpus_desugar_test.go
```diff
@@ -318,10 +318,51 @@ func normalizeDesugaredAST(s string) string {
 		return s
 	}
 	canonicalisePackageImports(root)
+	canonicaliseGeneratedFns(root)
 	canonicaliseInitFnBodies(root)
 	return printSExp(root)
 }
 
+// canonicaliseGeneratedFns sorts the compiler-generated top level functions
```
warunalakshitha, line 326: can we shortedn these comments

PR #953, opened 2026-09-17
fix(nodebuilder): support parenthesised actions

### LINKED ISSUE #951: [Bug]: parenthesised actions such as trap (wait f) fail with INTERNAL_ERROR
### Description

Wrapping an action in parentheses makes compilation fail with `fatal[INTERNAL_ERROR]` instead of compiling. The most visible case is `trap (wait f)`, but the problem is not specific to `trap` — any parenthesised action in a position where actions are allowed hits it, including `(wait f)` as a plain variable initializer and `(c->foo())`. jBallerina 2201.13.4 compiles and runs all of these.

Dropping the parentheses (`trap wait f`) works, so this is purely about the parenthesised form.

### Steps to Reproduce

`p6.bal`:

```ballerina
import ballerina/io;

function helper() returns int {
    return 5;
}

public function main() {
    future<int> f = start helper();
    int|error r = trap (wait f);
    io:println(r);
}
```

```bash
go build -o /tmp/nutbal ./cli/cmd && /tmp/nutbal run p6.bal
```

Actual:

```
fatal[INTERNAL_ERROR]: syntax node *st.WaitActionNode transformed to non-expression node *ast.BLangSingleWaitAction
 --> p6.bal:9:25
```

Expected (what `bal run p6.bal` prints): `5`

Other programs that fail the same way, and that jBallerina accepts:

```ballerina
int|error a = (wait f);        // fatal[INTERNAL_ERROR]
int b = (c->val());            // fatal[INTERNAL_ERROR]
var d = (start helper());      // fatal[INTERNAL_ERROR]
```

### Affected Version(s)

Reproduces on `main` at `dee511075`. Unrelated to the named-worker work.

### Related area

Compilation

### Possible cause

The parser gives `(wait f)` the kind `BRACED_ACTION` (`parser/parser.go:4982`), distinct from `BRACED_EXPRESSION`. In the node builder, `createActionOrExpressionInner` only special-cases `BRACED_EXPRESSION` (wrapping it in a `BLangGroupExpr`); a `BRACED_ACTION` falls through to the generic dispatch and reaches `transformBracedExpression`, which calls `createExpression` on its operand (`nodebuilder/node_builder.go:2184`). That requires a `BLangExpression`, but the operand is an action, so the assertion at `nodebuilder/node_builder.go:1991` fires.

jBallerina has the same structure, but there actions *are* expressions, so its cast succeeds. `BLangTrapExpr.Expr` in this port is already a `BLangActionOrExpression`, which is why the unparenthesised `trap wait f` works.

### Suggested assignee(s)

@heshanpadmasiri

A fix is in progress: `transformBracedExpression` builds an action-or-expression and unwraps the parentheses, matching jBallerina (parentheses are not meaningful around an action).

heshanpadmasiri:
Fixes #951

## Problem

Wrapping an action in parentheses failed compilation with `fatal[INTERNAL_ERROR]` instead of compiling:

```ballerina
int|error r = trap (wait f);   // fatal[INTERNAL_ERROR]
int|error a = (wait f);        // fatal[INTERNAL_ERROR]
var d = (start helper());      // fatal[INTERNAL_ERROR]
int b = (c->val());            // fatal[INTERNAL_ERROR]
```

## Root cause

The parser gives a parenthesised action the kind `BRACED_ACTION`, distinct from `BRACED_EXPRESSION`, but both are backed by the same `*st.BracedExpressionNode` and so dispatch to the same transform.

`createActionOrExpressionInner` special-cases only `BRACED_EXPRESSION`, wrapping it in a `BLangGroupExpr`. A `BRACED_ACTION` falls through to the generic dispatch and reaches `transformBracedExpression`, which called `createExpression` on its operand. That requires a `BLangExpression`, but the operand is an action, so the conversion failed and reported an `INTERNAL_ERROR`.

jBallerina has the same structure, but there actions *are* expressions, so its cast succeeds and the parentheses are simply dropped around an action. In this port `BLangTrapExpr.Expr` is already a `BLangActionOrExpression`, which is why the unparenthesised `trap wait f` already worked — only the braced wrapper was broken.

## Fix

`transformBracedExpression` now builds an action-or-expression instead of an expression, dropping the parentheses around an action as jBallerina does. `BRACED_EXPRESSION` is unaffected: it is still intercepted earlier and wrapped in a `BLangGroupExpr`, and that branch's assertion still holds because its operand is by construction an expression.

## Testing

New corpus test `corpus/bal/subset9/09-bug/braced-action-v.bal` covers `trap (wait f)`, a bare `(wait f)`, `(start f())`, `(c->m())`, nested `trap (trap (wait f))`, panic capture via `trap (c->broken())`, and `check (trap (c->val()))` — plus a plain `(1 + 2) * 3` to pin that expression grouping is unchanged.

The AST golden confirms the intended shapes: `trap (wait f)` becomes `(trap-expr (single-wait-action ...))` with no group wrapper, while `(1 + 2) * 3` still yields a `(group-expr ...)`.

`TestASTGeneration`, `TestIntegration` and `TestBIRSerializationRoundtrip` pass.

## Out of scope

Two orthogonal, pre-existing parser-level issues surfaced while investigating and are deliberately not addressed here:

- `int x = (wait f) + 1;` is rejected by the parser with `missing semicolon token` rather than a semantic "action not allowed here" diagnostic.
- `(c->val()).x` fails with `missing flush keyword`.

Both are diagnostic-quality issues in the parser, not internal errors.

## Summary by CodeRabbit

- **New Features**
  - Braced expressions now support action-based constructs, including asynchronous start and wait operations, nested actions, and trap expressions.
  - Improved handling of grouped actions and expressions enables more complex control flows and error recovery.

- **Documentation**
  - Updated the supported-expression documentation to include start actions, wait actions, trap expressions, and nested actions.

- **Bug Fixes**
  - Corrected evaluation of braced actions, including successful results and recovery from trapped remote-call failures.


nodebuilder/node_builder.go
```diff
@@ -2181,8 +2181,11 @@ func (n *nodeBuilder) transformBinaryExpression(binaryBLangExpression *st.Binary
 	return &bLBinaryExpr
 }
 
+// transformBracedExpression handles both BRACED_EXPRESSION and BRACED_ACTION. For the latter the operand is an action,
```
warunalakshitha, line 2184: I am not sure do we need this comment since code already implies handling both action and expression. we havent added any function comments similarly in other functions in node builder.

corpus/bal/subset9/09-bug/braced-action-v.bal
warunalakshitha, line 1: should we add these to subset 10?

warunalakshitha:
LGTM except minor NITs.

nodebuilder/node_builder.go
heshanpadmasiri: fixed

corpus/bal/subset10/10-bug/braced-action-v.bal
heshanpadmasiri: fixed

nodebuilder/node_builder.go
heshanpadmasiri: [AI RESPONSE] Agreed, dropped the function comment; the rationale now lives only in the commit message.

corpus/bal/subset10/10-bug/braced-action-v.bal
heshanpadmasiri: [AI RESPONSE] Good catch - the test uses `start`, `wait` and `trap`, which are subset 10 features. Moved the test and all goldens to `subset10/10-bug/`.

heshanpadmasiri:
@coderabbitai review

warunalakshitha (approved):
LGTM

PR #958, opened 2026-09-17
fix(semantics): run control-flow analyses inside anonymous functions

### LINKED ISSUE #952: [Bug]: control-flow analyses are skipped inside anonymous function bodies
## Description

Three compile-time checks that work inside a named function are silently skipped when the same code is written inside a block-bodied anonymous function: missing-return detection, uninitialized-variable detection, and unreachable-code detection.

Two of the three are not just missing diagnostics — they let a variable declared `int` hold nil, so the program compiles and then fails at runtime with a confusing `unsupported type combination: <nil> + int64` that points at the *caller* rather than at the defect.

### 1. Missing return — compiles, fails at runtime

```ballerina
import ballerina/io;

public function main() {
    var f = function(int n) returns int {
        if n > 1 {
            return 1;
        }
    };
    int v = f(0);
    io:println(v + 100);
}
```

```
error: unsupported type combination: <nil> + int64
        at main(a.bal:10)
```

The same body written as a named function is rejected at compile time:

```ballerina
function f(int n) returns int {
    if n > 1 {
        return 1;
    }
}
```

```
error[SEMANTIC_ERROR]: missing return statement
```

### 2. Uninitialized variable — compiles, fails at runtime

```ballerina
var g = function() returns int {
    int x;
    return x;
};
```

Same runtime failure as above. As a named function this is rejected with `error[SEMANTIC_ERROR]: variable may not be initialized`.

### 3. Unreachable code — compiles and runs, prints `1`

```ballerina
var h = function() returns int {
    return 1;
    io:println("never reported");
};
```

As a named function this is rejected with `error[SEMANTIC_ERROR]: Unreachable code`.

## Affected Version(s)

`Ballerina dev (Nutcracker)`, reproduced at commit `7a1e371d9`.

This is not a regression; it has been the behaviour since these analyses were introduced.

## OS, DB, other environment details and versions

macOS (darwin/arm64), Go 1.27. Not platform specific.

## Related area

Semantics

## Possible cause

`cfg.Build` (`semantics/internal/cfg/control_flow_analyzer.go:101`) only builds graphs for `pkg.Functions`, `pkg.InitFunction`, and class/service members. Nothing descends into `BLangLambdaFunction.Function`, so `PackageCFG` holds no entry for a lambda body.

All five passes in `cfg.Analyze` then iterate either `allFunctionCfgs` or `common.PackageFunctionDecls(pkg)`, neither of which includes lambdas, so every CFG-based analysis skips lambda bodies entirely.

Named workers had the same shape and were addressed on `feat/named-worker` by registering an independent CFG per worker, keyed by the worker's own symbol, and including it in reachability, explicit-return and uninitialized-variable analysis. The same approach should work for lambdas.

## Related issue(s)

Found while implementing named workers (#929 / #930).

heshanpadmasiri:
## Purpose
Resolves #952.

Missing-return, uninitialized-variable and unreachable-code detection were silently
skipped inside block-bodied anonymous functions. Two of these are not just missing
diagnostics: they let an `int`-typed variable hold nil, so the program compiles and
then fails at runtime with a confusing `unsupported type combination: <nil> + int64`
that points at the caller rather than at the defect.

## Approach
Anonymous function bodies are not reachable from any package level declaration list,
so the package CFG never contained a graph for them and every analysis that looks up
a graph by symbol bailed out.

- `Build` now walks the package with a `lambdaCollector` visitor and builds a CFG for
  every `BLangLambdaFunction`, including nested ones. Since these are not keyed by a
  declaration list, they are held as `fn`/`cfg` pairs in `PackageCFG.lambdaCfgs`.
- `allFunctionCfgs` yields the lambda graphs, so the analyses that iterate over all
  graphs pick them up automatically.
- The explicit-return and uninitialized-variable analyses gained an extra pass over
  `lambdaCfgs`. `analyzeInvokableExplicitReturn` was split so the lambda pass can
  reuse the body via `analyzeExplicitReturnWithCfg` with an already-resolved graph.
- The CFG pretty printer appends the lambda graphs, sorted by name since `Build`
  fills the slice concurrently.

## User stories
As a Ballerina developer, I get the same compile-time control-flow diagnostics
inside an anonymous function as I do inside a named function.

## Release note
Missing-return, uninitialized-variable and unreachable-code errors are now reported
inside anonymous function bodies instead of being silently skipped.

## Documentation
N/A - this restores the documented behaviour of existing compile-time checks inside
a context where they were not running; no documented behaviour changes.

## Training
N/A

## Certification
N/A - no impact on certification exams; this fixes missing diagnostics rather than
introducing new language behaviour.

## Automation tests
- Unit tests
  > CFG goldens under `corpus/cfg/` were regenerated; ~70 goldens now include the
  > lambda graphs that were previously absent.
- Integration tests
  > New negative corpus cases: `anonMissingReturn-e`, `anonNestedMissingReturn-e`,
  > `anonUninitVar-e`, `anonUnreachable-e` (subset7/07-closure) and
  > `anonInMethodMissingReturn-e` (subset9/09-object), each with the matching
  > `corpus/integration/` txtar.

## Security checks

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
macOS (darwin/arm64), Go toolchain per `go.mod`. Full `make test` and `make lint` run locally.

## Learning
The lint fixes in `compiler-tools` are a separate preparatory commit: the pre-commit
lint hook was already failing on pre-existing staticcheck/errcheck findings there,
unrelated to this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compiler analysis for anonymous and nested functions.
  - Missing return statements are now detected consistently, including within object methods.
  - Potentially uninitialized variables returned from anonymous functions are now reported.
  - Unreachable code in anonymous functions is now flagged.
  - Benchmark report exports now use the configured destination correctly.


heshanpadmasiri:
Closing in favour of two stacked PRs: the compiler-tools lint cleanup is unrelated to #952 and is being split out.

PR #968, opened 2026-09-21
fix(compiler-tools): stop benchmark tools leaking git worktrees

### LINKED ISSUE #950: [Bug]: benchmark tools leave stale git worktree entries and unbounded GOCACHE growth
### Description

The benchmark tooling under `compiler-tools/` creates git worktrees inside `$TMPDIR` and never removes them. On macOS the *directories* are eventually reclaimed (`dirhelper` deletes `$TMPDIR` content untouched for 3 days), so this looks harmless at first — but two effects are not covered by OS temp reaping:

1. **`git worktree list` accumulates stale entries forever.** When the OS reaps a checkout, git is never told, leaving a dangling registration in `.git/worktrees/`. These pile up across runs and show as `prunable`. Worse, an interrupted run can leave a worktree marked `locked`, which `git worktree prune` refuses to remove at all — it needs a manual `git worktree unlock` first.

2. **The Go build cache grows without bound.** Each checkout sits at a unique absolute temp path, and Go keys compile actions by source path, so every run seeds its own unshareable slice of `GOCACHE`. That lives in `~/Library/Caches/go-build` (not `$TMPDIR`), is never OS-managed, and Go's own GC only drops entries unused for 5 days.

Each `go test ./compiler-tools/benchmark-http` invocation costs roughly 420 MB of checkout plus its private cache slice, and nothing bounds how many accumulate within the OS retention window. Symptomatically this reads as "my Go cache keeps filling the disk", but the driver is the worktrees.

### Steps to Reproduce

```console
$ git worktree list | wc -l
21
$ ls -d "$TMPDIR"/httpbench-shared-bal-* 2>/dev/null | wc -l
0

$ (cd compiler-tools/benchmark-http && go test -run TestMeasureOnce -count=1 .)
ok  	httpbench	15.005s

$ git worktree list | wc -l
22                       # +1, never removed
$ du -sh "$TMPDIR"/httpbench-shared-bal-*
420M	.../httpbench-shared-bal-3939624730
```

Repeat and the counts keep climbing, one worktree per invocation. `git worktree prune` does not reclaim them while the directory still exists; once the OS reaps the directory, the registration is left behind as `prunable`.

### Possible cause

`ensureSharedBalBinary` (`compiler-tools/benchmark-http/runner_test.go:39`) creates a temp root and a detached worktree with no matching teardown:

```go
sharedBalOnce.Do(func() {
	root, err := os.MkdirTemp("", "httpbench-shared-bal-*")   // never removed
	...
	wt, err := checkoutWorktree(root, "shared", "HEAD")        // never removeWorktree'd
	...
	sharedBalPath, sharedBalErr = buildInterpreter(wt)
})
```

Because it's `sync.Once`-scoped, `t.Cleanup` on the calling test is the wrong hook; this likely wants a `TestMain` that tears the shared root down after `m.Run()`.

Related, same class of problem:

- `benchmark-http/main.go:93-118` (`httpbench-work-*`, `httpbench-hello-*`) and `compiler-tools/benchmark/runner.go:57` (`bal-bench-*`) clean up via `defer`, which does not run on SIGINT/SIGTERM — interrupted runs are what leave `locked` worktrees behind.

### Suggested fix

- Add `TestMain` teardown in `benchmark-http` that calls `removeWorktree` (not just `os.RemoveAll`) so git state stays consistent.
- Handle SIGINT/SIGTERM in both benchmark binaries so cleanup runs on interrupt.
- Run `git worktree prune` at startup in both tools so pre-existing stale entries self-heal, and avoid `--lock` (or always unlock in teardown) so prune can do its job.

### LINKED ISSUE #970: [Bug]: interrupting httpbench orphans the bal service and blocks the next run
Follow-up from #950 / #968. Not a regression — this predates that change — but it is the one interrupt-time leftover #968 does not close, and it is the most likely to actually bite someone.

### Description

`measureOnce` launches the service in **its own process group**, deliberately, so the whole tree can be killed together:

```go
cmd := exec.Command(balPath, "run", helloFile)
setProcAttr(cmd) // own process group (unix) so the whole tree can be killed
...
defer func() {
    killGroup(cmd)
    waitPortClose(servicePort, 15*time.Second)
}()
```
(`compiler-tools/benchmark-http/runner.go`)

That teardown is a `defer`. #968 added a SIGINT/SIGTERM handler that tears down worktrees and temp dirs and then calls `os.Exit(130)` — which **skips deferred functions**. And because the service is in a separate process group, a terminal Ctrl+C does not reach it either.

So on interrupt: the worktrees and temp dirs are cleaned up correctly, the tool exits 130, and the `bal` service survives indefinitely. On unix it keeps running off the deleted inode.

The next invocation then fails immediately at the fail-fast check:

```
port 9090 is already in use before launch; a prior service may not have shut down
```

which requires manually hunting down the orphan before benchmarking works again.

### Steps to reproduce

1. `go run ./compiler-tools/benchmark-http HEAD~1 HEAD`
2. Wait for `Repeat 1/2: ...` (the service is now up)
3. Ctrl+C
4. `lsof -i :9090` — the `bal` process is still there
5. Re-run step 1 — it dies at once with the port-in-use error

### Suggested fix

Record the live `*exec.Cmd` in a mutex-guarded package var and `killGroup` it from the cleanup stack that #968 introduced, so the interrupt path tears the service down the same way the deferred path does.

`compiler-tools/benchmark` has a milder version of this: hyperfine inherits the process group, so Ctrl+C does reach it; only a bare `kill <pid>` or a CI cancellation orphans it, and it exits on its own once its `bal` child fails.

heshanpadmasiri:
Closes #950.
Closes #970.

## The problem

The benchmark tooling under `compiler-tools/` registers git worktrees in the shared repository and never reliably unregisters them. On my checkout, **202 of 225** `git worktree list` entries were stale benchmark leftovers:

| prefix | count | source |
|---|---:|---|
| `worktree-shared-HEAD*` | 191 | `ensureSharedBalBinary` (test helper) |
| `worktree-*-bal-bench-*`, `worktree-{base,head}-HEAD*` | 11 | interrupted CLI runs |

Reproducer — one test invocation, before and after:

```console
$ git worktree list | wc -l          # 225  (195 prunable, 4 locked)
$ go test -run TestBuildInterpreterProducesRunnableBinary -count=1 ./compiler-tools/benchmark-http
ok  httpbench  57.152s
$ git worktree list | wc -l          # 227
$ du -sh $TMPDIR/httpbench-shared-bal-*   # 422M
```

Investigating from the symptoms turned up **three independent defects**, not one.

### 1. The test helper has no teardown

`ensureSharedBalBinary` creates an `os.MkdirTemp` root and a detached worktree inside a `sync.Once` and removes neither. `t.Cleanup` is the wrong hook — the worktree outlives the first caller — so this is torn down from `TestMain`.

### 2. GOCACHE growth, which removing the worktree does *not* reclaim

Without `-trimpath` the compiler embeds absolute source paths, so every checkout at a fresh temp path seeds its own unshareable cache slice. Measured on `./cli/cmd` with a scratch `GOCACHE`:

| | build at path A | then build at path B |
|---|---:|---:|
| `go build` | 329 MB | **485 MB** (+156 MB) |
| `go build -trimpath` | 329 MB | 333 MB (+4 MB) |

Deleting the worktree leaves those 156 MB behind — this needed a separate fix. The only runtime consumer of source paths is `common/errors/errors.go` (stack formatting), so the effect is cosmetic.

### 3. Cleanup is neither interrupt-safe nor able to remove locked worktrees

The issue guessed `--lock`; that is not it — there is no `--lock` anywhere in the code. The actual cause is git's own `locked: initializing` marker, written during `worktree add` and cleared on success. Interrupt mid-add and it persists forever, because `git worktree prune` skips locked entries and both tools passed a single `--force`:

```console
$ git worktree remove --force <locked>
fatal: cannot remove a locked working tree; use 'remove -f -f' to override
$ git worktree remove --force --force <locked>   # rc=0
```

## The fix

| File | Change |
|---|---|
| `benchmark-http/interrupt.go`, `benchmark/interrupt.go` *(new)* | `onInterrupt(cleanup) (stop func())` — SIGINT/SIGTERM handler, idempotent `stop`. `signal.Reset` runs before `cleanup` so a second signal is acted on immediately rather than queueing behind a ~420 MB `RemoveAll`. |
| both `runner.go` | `removeWorktree` → `--force --force`; new `pruneWorktrees()`; `buildInterpreter` → `-trimpath` |
| `benchmark-http/main.go`, `benchmark/runner.go` | `run` prunes at startup and replaces the per-resource `defer`s with one cleanup closure over pre-declared vars, registered via `defer` **and** `onInterrupt` |
| `benchmark-http/runner_test.go` | `TestMain` + `cleanupSharedBal`; shared state guarded by a mutex against the signal goroutine |

`compiler-tools/benchmark` and `compiler-tools/benchmark-http` are separate Go modules and cannot import the main module's PAL, so the direct `os/exec` use there is unchanged — and `onInterrupt` is duplicated because there is nothing to share it through.

## Tests

Corpus tests do not reach `compiler-tools` (separate modules, not an interpreter stage), so these are unit tests. Each was **mutation-verified in both modules** — reverted the production change, confirmed the test fails, restored:

- `TestRemoveWorktreeRemovesLockedWorktree` — fails with a single `--force`. Teardown uses raw git rather than the function under test, so a regression does not also leak the locked registration it exists to catch.
- `TestPruneWorktreesReclaimsReapedCheckouts` — deletes the checkout behind git's back; fails when `pruneWorktrees` is a no-op.
- `TestInterruptRunsCleanupBeforeExit` — re-execs the test binary, sends SIGTERM, asserts exit 130 *and* that cleanup ran. Fails when the handler omits `cleanup()`.

These use `git worktree add --no-checkout` (~0.1s rather than a 3.3s 17k-file checkout) since the removal and prune paths act on the registration, not the files — `go test -short ./compiler-tools/benchmark` stays cheap. Note `pruneWorktrees()` is repo-global: it also reclaims any unrelated registration whose directory happens to be absent at that moment.

**Known coverage gap:** the production *call sites* are not covered — deleting `defer onInterrupt(cleanup)()` or the `-trimpath` argument leaves the suite green. Covering them would mean restructuring `run()` for testability; flagging rather than doing that.

## Verification

Both modules: `go vet` clean, `golangci-lint` 0 issues, full suites pass under `-race` and `-short`, and `git worktree list` is unchanged before and after a full run.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup when benchmarks are interrupted, including handling of SIGINT and SIGTERM.
  - Ensured benchmark services are stopped and do not remain running after interruption.
  - Removed stale or locked Git worktree state more reliably after interrupted runs.
  - Prevented leftover temporary files and checkout directories from accumulating.
  - Improved build-cache reuse across temporary benchmark checkouts.
  - Ensured cleanup runs consistently after successful completion and during interruption.

---

### Review notes

Three review rounds; findings addressed in `6f7c74ec4`:

- **Data race** on the cleanup closure's captured paths — the interrupt handler read variables the main goroutine was still writing, so it could observe a stale empty string and skip removing the newest worktree. Replaced with a small guarded teardown stack where each resource registers its own step as it is created.
- `onInterrupt`'s doc asserted "at most one registration live", which the test binary itself breaks (`TestMain` plus `run()` during `TestRunEndToEndSelfComparison`). Reworded to describe the consequence instead of stating a rule the code violates.
- Two doc comments had been glued onto the wrong function by an earlier edit.

Also closes #970: interrupting the tool used to orphan the `bal` service, because it runs in its own process group and `os.Exit` skips the `killGroup` defer. `measureOnce` now registers its teardown on the cleanup stack, so the interrupt path reaps the service the same way the deferred path does (`TestInterruptStopsService`).

Fourth round:

- `onInterrupt` now registers against a single process-wide `interruptCoordinator` instead of installing a handler per call. Two registrations really can be live at once — `TestMain` plus `run()` during `TestRunEndToEndSelfComparison` — and each handler called `os.Exit` on its own, so the first exit truncated the other's cleanup. All registered steps now run before the exit. Reproducer: `compiler-tools/benchmark-http/interrupt_overlap_test.go`.
- Shortened the `onInterrupt` doc comment and the new test helpers' comments.


compiler-tools/benchmark-http/main.go
```diff
@@ -90,32 +90,50 @@ func parseConfig(args []string) (config, error) {
...
 	}
 
-	baseWT, err := checkoutWorktree(workRoot, "base", cfg.baseRef)
+	baseWT, err = checkoutWorktree(workRoot, "base", cfg.baseRef)
 	if err != nil {
 		return err
 	}
-	defer removeWorktree(baseWT)
-	headWT, err := checkoutWorktree(workRoot, "head", cfg.headRef)
+	headWT, err = checkoutWorktree(workRoot, "head", cfg.headRef)
```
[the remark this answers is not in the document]
heshanpadmasiri: [AI RESPONSE: If you are human reading this ignore this, I'll get back to this sorry about the noise] Failed to reproduce — can you present a test case for this?

The premise is accurate: the current code still registers the removal after `checkoutWorktree` returns (`baseWT, err := checkoutWorktree(...)` then `c.add(func() { removeWorktree(baseWT) })`), so the path is not published before `git worktree add` starts and there is no cancel/wait of the `add` child. But the predicted harm — "leaving a locked worktree registration" — did not occur in any scenario I tried.

What I ran, using the package's re-exec helper pattern: a helper mirroring `run()`'s ordering (register `workRoot` teardown, then `checkoutWorktree`), signalled while `git worktree add` was materializing files, then inspecting `git worktree list --porcelain` and `.git/worktrees`.

1. Signal to the whole process group (terminal Ctrl-C semantics), delays 0.3s–3s: no registration ever left behind. `git worktree add` installs its own `sigchain`/`atexit` junk removal, so on SIGINT/SIGTERM it deletes both the checkout and the `.git/worktrees/<name>` admin entry. Confirmed independently of Go: `git worktree add --detach … HEAD &` + `kill -INT` at 0.3s/0.8s/1.5s/2.5s left zero admin entries every time. No locked leftover to reclaim.
2. Signal to the tool process only, leaving `git worktree add` orphaned and racing `os.RemoveAll(workRoot)`, delays 1s/3s/5s: at one timing a registration survived the handler, but its `.git` link had already been removed by the racing `RemoveAll`, so it was prunable — and `pruneWorktrees()`, which `run()` calls at the start of every invocation, reclaimed it. Later delays left nothing at all.

Between git's own signal cleanup, `pruneWorktrees()` on every run, and the double `--force` in `removeWorktree`, I could not produce a persistent or locked registration. If you have a timing or platform where the locked marker actually survives, a concrete repro would help.

compiler-tools/benchmark-http/interrupt.go
```diff
@@ -0,0 +1,134 @@
...
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+// interruptExitCode is the conventional 128+SIGINT status for a signalled
+// process.
+const interruptExitCode = 130
+
+// onInterrupt runs cleanup on SIGINT/SIGTERM and exits, covering the gap that
```
warunalakshitha: This is longer than the usual style in this package. Can we shorten it to ~3 lines. e.g. that it runs cleanup on SIGINT/SIGTERM then exits (defers miss signals), stop is idempotent, and cleanup must be too (it can race a deferred cleanup; with two live handlers the first os.Exit wins)? 

Same for compiler-tools/benchmark/interrupt.go.

compiler-tools/benchmark/benchmark_test.go
```diff
@@ -628,3 +632,161 @@ func requireHyperfine(t *testing.T) {
...
+		t.Skip("git worktree integration is exercised on Linux/macOS only")
+	}
+	path := filepath.Join(t.TempDir(), name)
+	if err := runCmdSilent(".", "git", "worktree", "add", "--no-checkout", "--detach", path, "HEAD"); err != nil {
+		t.Fatalf("git worktree add: %v", err)
+	}
+	return path
+}
+
+// TestRemoveWorktreeRemovesLockedWorktree covers the interrupted-`worktree
```
warunalakshitha, line 649: new test helpers/tests have pretty long function comments (newBareWorktree, TestRemoveWorktree…, etc.) that we don’t usually do elsewhere here. Fine to drop those or leave a one-liner only when the name isn’t enough.

warunalakshitha:
is this fixing https://github.com/ballerina-nutcracker/ballerina/issues/970 too? then we can update PR

compiler-tools/benchmark/interrupt.go
```diff
@@ -0,0 +1,90 @@
...
+// deferred cleanup leaves when the process is signalled rather than returning
+// normally. The returned stop releases the handler and is safe to call more
+// than once.
+//
+// cleanup must be safe to call concurrently with the caller's own deferred
+// cleanup, since a signal arriving as the caller returns can run both. When
+// two registrations are live — as the test binary does, installing one in
+// TestMain and another inside run — both wake on a signal and whichever
+// reaches os.Exit first truncates the other's cleanup.
+func onInterrupt(cleanup func()) (stop func()) {
```
warunalakshitha, line 35: can we reuse code in benchmark-http interrupt.go or vice versa. look similar

warunalakshitha: Seems cant do since separate go modules

heshanpadmasiri:
> is this fixing #970 too? then we can update PR

~~No I'll revisit it later. Currently fixing the worktree/go cache leakage is the main issue for me. (constantly running out of disk space)~~
Added the fix to this PR

compiler-tools/benchmark-http/interrupt.go
```diff
@@ -0,0 +1,134 @@
...
+// deferred cleanup leaves when the process is signalled rather than returning
+// normally. The returned stop releases the handler and is safe to call more
+// than once.
+//
+// cleanup must be safe to call concurrently with the caller's own deferred
+// cleanup, since a signal arriving as the caller returns can run both. When
+// two registrations are live — as the test binary does, installing one in
+// TestMain and another inside run — both wake on a signal and whichever
+// reaches os.Exit first truncates the other's cleanup.
```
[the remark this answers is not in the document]
heshanpadmasiri: [AI RESPONSE: If you are human reading this ignore this, I'll get back to this sorry about the noise] Confirmed and fixed. `onInterrupt` now registers against a package-level `interruptCoordinator`: one `signal.Notify`, every registered step runs (LIFO) and `os.Exit(130)` happens only after the last one. `stop` deregisters and releases the handler once nothing is left, so a later signal terminates normally. Reproducer added at `compiler-tools/benchmark-http/interrupt_overlap_test.go` — it fails on the previous `interrupt.go` (`outer cleanup did not complete before the process exited`) and passes now. Applied to `compiler-tools/benchmark` too, to keep the two files in sync.

compiler-tools/benchmark-http/interrupt.go
heshanpadmasiri: fixed

compiler-tools/benchmark/benchmark_test.go
heshanpadmasiri: fixed

compiler-tools/benchmark/interrupt.go
heshanpadmasiri: fixed

heshanpadmasiri:
[AI RESPONSE] Yes — #970 is fixed by this PR: `measureOnce` registers `killGroup(cmd)` + `waitPortClose` on the cleanup stack (`compiler-tools/benchmark-http/runner.go`), so the interrupt path reaps the service in its own process group, covered by `TestInterruptStopsService`. PR description updated with `Closes #970`. (re: https://github.com/ballerina-nutcracker/ballerina/pull/968#issuecomment-5758130019)

warunalakshitha (approved):
LGTM

